PR #612, opened 2020-12-11
Python CLI

dmidnight:
Rewrite CLI in Python and move to this repo

- [x] Wrap start of ./main [options]
- [x] Generate Account keys
- [x] Generate VRF Keys
- [x] Create Node Config
- [x] Update Node Config from Template, preserving important values
- [x] Create Custom Chain Config using testnet as template

Note: This updates the ENTRYPOINT in the Dockerfile to point to the CLI. This adds new options, but the familiar startup functionality is wrapped by the `start` command. Any downstream projects will need to update their start command to include the word `start`.

Example:
* Old:
```bash
docker run taraxa/taraxa-node:latest --conf_taraxa ../path/to/file --conf_chain ../path/to/file
```

* New:
```bash
docker run taraxa/taraxa-node:latest start --conf_taraxa ../path/to/file --conf_chain ../path/to/file
```


sfrique:
superseded #632

PR #620, opened 2021-01-05
DAG DPOS Proposal

gaoqi8649:
## Purpose

_Provide any information reviewers might need to have context on your changes_

## For reviewers
_Describe anything you want to call out to the reviewers e.g. areas to focus on, decisions you made, specific feedback you are looking for, etc._

## Tests
_Describe what you did to test the changes you made. Include evidence such as screenshots or terminal commands and output._

## Version Notes ##
{ major | minor | patch }

_Decide whether this is a major, minor, or patch version change. Major is for backwards incompatible changes (This is only relevant once the service is live and in production and the major version is at least 1. While the major version is 0, it is ok to have backwards incompatible changes in a minor version.). Minor is for additional features. Patch is for bug fixes._

_Describe the changes in this version. This is essentially the details you would include in the squash commit message._

_NOTE: Include the version notes in the squash commit message_


justinsnapp (approved):
Thanks for this!

src/CMakeLists.txt
```diff
@@ -176,4 +178,5 @@ else ()
 endif ()
 
 # Main taraxad binary
-add_subdirectory(taraxad)
\ No newline at end of file
+add_subdirectory(taraxad)
+message("Boost_INCLUDE_DIR: ${Boost_INCLUDE_DIR}")
```
JakubFornadel, line 182: remove this line

src/dag/dag_block_manager.cpp
```diff
@@ -0,0 +1,268 @@
...
+      }
+      // Verify DPOS
+      auto period = getPeriod(blk.first.getLevel());
+      auto dag_block_sender = blk.first.getSender();
+      if (!final_chain_->dpos_is_eligible(period, dag_block_sender)) {
+        auto executed_period = pbft_chain_->getPbftExecutedChainSize();
+        auto dpos_period = executed_period;
+        if (dpos_config_) {
+          dpos_period += dpos_config_->deposit_delay;
+        }
```
JakubFornadel, line 231: would be acutally nice to use some new lines to separate some blocks of code, it is easier to orientate in code and read it...

src/dag/dag_block_manager.cpp
```diff
@@ -0,0 +1,268 @@
...
+      capacity_(capacity),
+      num_verifiers_(num_verifiers),
+      db_(db),
+      trx_mgr_(trx_mgr),
+      final_chain_(final_chain),
+      pbft_chain_(pbft_chain),
+      log_time_(log_time),
+      queue_limit_(queue_limit),
+      blk_status_(10000, 100),
+      seen_blocks_(10000, 100) {
```
JakubFornadel, line 21: maybe stupid question, but why are these constants se to 1000, 100 ? Some comment or even put it in constexpr variable with comments on them might be valuable. If it is obvious and I just don't get it because of my lack of knowledge let me know... otherwise no one will have any idea what is the reason behind these constant withou comments...

src/dag/dag_block_manager.cpp
```diff
@@ -0,0 +1,268 @@
...
+std::shared_ptr<DagBlock> DagBlockManager::getDagBlock(blk_hash_t const &hash) const {
+  std::shared_ptr<DagBlock> ret;
+  auto blk = seen_blocks_.get(hash);
+  if (blk.second) {
+    ret = std::make_shared<DagBlock>(blk.first);
+  }
+  if (!ret) {
+    return db_->getDagBlock(hash);
+  }
+  return ret;
```
JakubFornadel, line 64: why not just write: 
if (blk.second) {
    return std::make_shared<DagBlock>(blk.first);
  }
   
  return db_->getDagBlock(hash);

src/CMakeLists.txt
gaoqi8649: Fixed

src/dag/dag_block_manager.cpp
gaoqi8649: I agree with you, but I don't think should add new lines here

src/dag/dag_block_manager.cpp
gaoqi8649: ExpirationCacheMap class defined at util.hpp. ExpirationCacheMap(uint32_t max_size, uint32_t delete_step)
There is no specific reason to set max size 10000 and delete step to 100.
I remember long time ago, we set to max size 2000 and hit issue. So change to 10000
This will depends on TPS. And the value will change depend on performance/torture testing report.

src/dag/dag_block_manager.cpp
gaoqi8649: Right, fixed

JakubFornadel (approved):

PR #624, opened 2021-01-12
MacOS fixes for cmake build system

02p01r:


CMakeLists.txt
```diff
@@ -1,5 +1,13 @@
 # Defines Taraxa library target.
 
+if (DEFINED SYSTEM_HOME_OVERRIDE)
+    list(APPEND CMAKE_PREFIX_PATH ${SYSTEM_HOME_OVERRIDE})
+    include_directories(${SYSTEM_HOME_OVERRIDE}/include)
+    link_directories(${SYSTEM_HOME_OVERRIDE}/lib)
+endif ()
```
02p01r, line 7: This was the easiest way to support submodule builds (some of them need custom installation of openssl and boost on mac) + support FindXXX on the same custom install dir

CMakeLists.txt
```diff
@@ -1,5 +1,13 @@
...
+set(CMAKE_OSX_SYSROOT /)
+set(CMAKE_MACOSX_RPATH ON)
```
02p01r, line 9: I'm not completely sure about the effects, but It once helped me solve some problem long ago, and now I just include it everywhere :D

src/CMakeLists.txt
```diff
@@ -6,14 +6,7 @@ set(CMAKE_CXX_STANDARD 17)
...
 set(BOOST_COMPONENTS)
-list(APPEND BOOST_COMPONENTS program_options system filesystem)
-
-if (APPLE)
-    list(APPEND BOOST_COMPONENTS thread-mt log-mt log_setup-mt)
-else()
-    list(APPEND BOOST_COMPONENTS thread log log_setup)
-endif()
-
+list(APPEND BOOST_COMPONENTS program_options system filesystem thread log log_setup)
```
02p01r, line 9: I think we shouldn't respect `-mt` prefix anymore. It was an artifact of `brew` installation on Mac, and at the moment the brew boost version is not compatible with our code anyway, so we have to install boost manually on mac - in that case no `-mt` suffix will be present

justinsnapp: Agree

CMakeLists.txt
```diff
@@ -1,5 +1,13 @@
 # Defines Taraxa library target.
 
+if (DEFINED SYSTEM_HOME_OVERRIDE)
+    list(APPEND CMAKE_PREFIX_PATH ${SYSTEM_HOME_OVERRIDE})
+    include_directories(${SYSTEM_HOME_OVERRIDE}/include)
+    link_directories(${SYSTEM_HOME_OVERRIDE}/lib)
+endif ()
+set(CMAKE_OSX_SYSROOT /)
```
02p01r, line 8: This one has effects only on mac, and it will tell cmake to use the system root as the toolchain root. Without this it will use some "default toolchain" in mac os, which doesn't have compilers, etc. installed by default, and god knows how to set it up properly.

CMakeLists.txt
JakubFornadel: I guess I am ok with this. Just want to point out that there is a direct support in cmake for overriding boost and openssl custom instalation, not sure if that is the only challenge here, but if yes we can use cmake arguments: 
-DBOOST_ROOT=/path/to/custom/boost... -> [example](https://stackoverflow.com/questions/38905480/how-to-pass-boost-root-and-boost-library-to-cmake) [cmake doc](https://cmake.org/cmake/help/latest/module/FindBoost.html)
-DOPENSSL_ROOT_DIR=/path/to/custom/openssl...  -> example follows same logis as boost example, [cmake doc](https://cmake.org/cmake/help/v3.6/module/FindOpenSSL.html)

02p01r: We also need a way to provide custom clang-format and cppcheck. This is a generic solution, should cover many potential needs. And only needs setting single config variable.

JakubFornadel (approved):

PR #626, opened 2021-01-15
[WIP] Helping with Cmake

02p01r:
- New fixes for new MacOS problems (due to the new static linking)
- Improved the submodule builds. Now, we don't pollute the original submodule directories, and also the build system will now always pick up filesystem-level changes in the submodule directories. In other words, I believe now there will be zero attention to submodules needed from developers.

More detailed description TBD!


JakubFornadel (approved):

JakubFornadel:
I merget it and realized it does not work so probably you have not finished yet, sorry for that. I reverted this PR but can't reopen it, so please create a new one

PR #628, opened 2021-01-22
Cmake build system follow-up

02p01r:
- Improved submodule handling
- Complete static build in taraxa-evm
- MacOS build fixes + docs + improved installers
- General refactoring and small fixes


CMakeModules/cppcheck.cmake
```diff
@@ -6,8 +6,8 @@ if (NOT CPP_CHECK_EXE)
 else ()
     message(STATUS "cppcheck found: ${CPP_CHECK_EXE}")
 
-    add_custom_target(cpp-check ALL
-            COMMAND ${CPPCHECK}
```
JakubFornadel, line 10: why did you remove ALL ? it's there so when you run make all in jenkins it is automatically checked with cppcheck

02p01r: I think it's better to run it like `make cpp-check all` - this way cpp-check will be ordered first even in parallel mode. Otherwise you may run a bunch of stuff in parallel, and it will take a while until the whole process tree collapses on cpp-check error.

02p01r: It will also mess up your cppcheck error log

JakubFornadel (commented):
nope, try to run make all, cppcheck is always triggered as first thing before compiling starts... parallel or non-parallel

CMakeModules/cppcheck.cmake
JakubFornadel: nope, try to run make all, cppcheck is always triggered as first thing before compiling starts... parallel or non-parallel

JakubFornadel: on the other hand, same principle applies to clang-format-check and locally you have to run those 2 separate targets and then build all, which is...

submodules/CMakeLists.txt
```diff
@@ -1,18 +1,37 @@
-
+set(
+        __COMPILE_DEFINITIONS__
```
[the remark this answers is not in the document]
02p01r: The goal of this __COMPILE_DEFINITIONS__ is to have a set of compile definitions that is consistent across the submodule builds and the main build. In general, it is a safe approach to compile all dependencies as well as the main code with the union of all compile definitions needed for each particular module/dependency. This variable allows me to achieve DRY (Don't Repeat Yourself).

submodules/CMakeLists.txt
```diff
@@ -1,18 +1,37 @@
...
+        BOOST_SPIRIT_THREADSAFE
+        CRYPTOPP_DISABLE_ASM
+)
+if (TARAXA_STATIC_BUILD)
+    list(APPEND __COMPILE_DEFINITIONS__ BOOST_ALL_STATIC_LINK)
+else ()
+    list(APPEND __COMPILE_DEFINITIONS__ BOOST_ALL_DYN_LINK)
+endif ()
```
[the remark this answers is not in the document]
02p01r: See https://github.com/Taraxa-project/taraxa-node/pull/628#discussion_r563329862

In addition to that, I know that `BOOST_ALL_DYN_LINK` is covered by `Boost_USE_STATIC_LIBS` in the main code build, yet still I prefer a principled approach described above - this way it looks more robust I think.

CMakeModules/cppcheck.cmake
02p01r: Which is... what?

submodules/CMakeLists.txt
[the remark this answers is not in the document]
02p01r: I actually considered propagating our compile definitions to submodule builds, but the solution would be too complicated, you wouldn't have liked it for sure :D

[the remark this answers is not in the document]
JakubFornadel: well I don't share this opinion but ok, approved

CMakeModules/cppcheck.cmake
JakubFornadel: when I develop locally I have to think about running cppcheck and clang-format-check manually before pushing and creating PR, which I for example usually forget. Then of course PR fails, I realize that I forgot those checks, have to run it manually and push changes. On the other hand when it is run automatically no one will have to even think about it as it run atuomatically. Also I dont get your arguments why to switch it off - okey so maybe jenkins build fail will be delayed max by few seconds, that's totally irrelevant. Mess up cppcheck log ? Yes there might be few "compile" lines after printed error because of parallel processing but still it should be pretty easy and fast to scroll up(or just ctrl+f) and find the error so I don't find this very important compared to having to run those checks automatically...

02p01r: `ALL` means that a target will be a part of `all` target, which is generated by CMake by default. You say that having `ALL` is a remedy for forgetting to run clang-format and cppheck, because it will be done automatically. Yes, it will be done automatically, but only if all you run is `all` target. Do you always run `all` target before pushing to git? Me - not. Probably not so common across other people as well.
You found my arguments weak - yet they are arguments. I presented you with some logic why it's better not to have `ALL` for those targets. Your arguments for using `ALL`, on the other hand, are nonexistent.

02p01r: If you want to reduce the probability of forgetting to run cppcheck and clang-format, we can add them as dependency for `app_base` target. Or use git hooks (I'm not a big fan though)

02p01r: Moreover, there won't be "few" compile lines before the cppcheck error abort. There will be a lot of them.

02p01r: Also, if you look at the changes I made, you'll see that I actually grouped `clang-format-check` and `cpp-check` behind `static-check` target for convenience

submodules/CMakeLists.txt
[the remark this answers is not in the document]
02p01r: What opinion do you not share exactly? By the way, in the latest version of these changes, I moved `BOOST_ALL_DYN_LINK` and `BOOST_ALL_STATIC_LINK` inside the submodule makefile, so it should look better to you I suppose.

CMakeModules/cppcheck.cmake
02p01r: Let's do this way: If you're still not convinced - just say so, and I will do like you want. I'm ok to live with it, probably won't affect me anyway. And probably, if we don't pay much attention to running clang-format and cppcheck locally before pushing, some team members will feel the pain of finding out why their jenkins build failed early eventually, and probably we'll change the flow like I suggest - or probably not, not a big deal.

JakubFornadel: add those checks as taraxad dependency then

JakubFornadel (approved):

justinsnapp (approved):
LGTM!

PR #630, opened 2021-01-27
Disable problematic SYSTEM include directory inference by CMake

02p01r:


JakubFornadel (approved):

JakubFornadel:
If this works for Qi's problem, definitely cleaner solution...

gaoqi8649 (approved):
Tested and works

PR #631, opened 2021-01-27
`#pragma once` instead of include guards

02p01r:
I think we should use `#pragma once` instead of include guards. It is de-facto well supported, and much cleaner. And you don't have to think about include guard naming conventions/consistency. It is also believed to lead to slightly better compilation performance.


JakubFornadel (approved):

JakubFornadel:
> I think we should use `#pragma once` instead of include guards. It is de-facto well supported, and much cleaner. And you don't have to think about include guard naming conventions/consistency. It is also believed to lead to slightly better compilation performance.

justinsnapp (approved):
Definitely cleaner and its pretty widely supported now. :-)

PR #632, opened 2021-01-29
Add cli

sfrique:
## Purpose

Add simple CLI with the image, so we can auto-generate configs for named networks or other purposes that may come along.

This PR updates  #612 code was refactored and changed a bit.

Please verify if the template config makes sense. 

## For reviewers

<del> Do not merge, there is a change on `Jenkinsfile` to use another branch for testing. We need to verify this branch first and go back to the right branch after merge there. https://github.com/Taraxa-project/taraxa-testnet/pull/35 </del>
Everything should be good now.


docker-entrypoint.sh
```diff
@@ -0,0 +1,46 @@
...
+    exec taraxad \
+            --conf_taraxa $TARAXA_CONF_PATH
+    ;;
+  exec)
+    exec "${@:2}"
+    ;;
+
+  *)
+    echo "You should choose between:"
+    echo "cli, taraxad, sigle, join {NAMED_NETWOTK}"
```
justinsnapp, line 43: sigle should be single

sfrique: fixed in: cf38cc47689e7cd974603500c350611cb0e233fa

justinsnapp (approved):

PR #633, opened 2021-02-02
Graphql implementation

mfrankovi:


Dockerfile
```diff
@@ -66,8 +72,24 @@ RUN apt-get update \
...
+        pegtl-dev=$PEGTL_VERSION \
+        rapidjson-dev=$RAPIDJSON_VERSION \
+        libgtest-dev=$GTEST_VERSION \
     && rm -rf /var/lib/apt/lists/*
 
+# Install cppgraphqlgen
+RUN curl -SL https://github.com/microsoft/cppgraphqlgen/archive/v$CPPGRAPHQLGEN_VERSION.tar.gz \
+    | tar -xzC /tmp \
+    && curl -SL https://github.com/taocpp/PEGTL/archive/7d039707cf835cea63daa78a717e18fcc5bcf95b.tar.gz \
+    | tar -xzC /tmp/cppgraphqlgen-${CPPGRAPHQLGEN_VERSION} \
```
JakubFornadel, line 84: there is also official pegtl-dev ubuntu package, not sure what's the deal with versions if that would be suitable but if yes, might be easier and cleaner to use official package

mfrankovi: Actually I did add the official with this line:
pegtl-dev=$PEGTL_VERSION \

I should have actually remove that since I added it this way.
For some reason with that official one I am getting some errors. It only work with this specific version which is a submodule there

JakubFornadel: hmm ok, and did you also consider to download it through submodule init ?

mfrankovi: I guess I would need then to download cppgraphqlgen with git clone. I think if you download a zipped version there is no git submodule info in there or am I mistaken?

src/network/graphql/Log.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "Transaction.h"
+#include "graphqlservice/GraphQLSchema.h"
+#include "graphqlservice/GraphQLService.h"
+#include "graphqlservice/JSONResponse.h"
+#include "graphqlservice/introspection/Introspection.h"
+
+using namespace std::literals;
+
+namespace graphql {
+namespace taraxa {
```
JakubFornadel, line 27: you can even use nested namespaces since C++17, e.g. grapghql::taraxa

Dockerfile
mfrankovi: If you have a cleaner way to do it, please do. I know that this downloading and then copying is not really a nice way to do it :)

src/network/graphql/Log.cpp
```diff
@@ -0,0 +1,56 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
```
JakubFornadel, line 3: files in graphql/ except grapqhl/gen/ were created manually right ? Why is here microsoft disclaimer ?

src/network/graphql/Log.h
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License.
+
+#pragma once
+
+#ifndef SRCNETGRAPHQLLOG_H
+#define SRCNETGRAPHQLLOG_H
+
```
JakubFornadel, line 8: just a few days ago all guards were removed and #pragma once is used instead in all headers

src/network/rpc/RpcServer.h
```diff
@@ -16,14 +17,19 @@ class RpcHandler;
 
 class RpcServer : public std::enable_shared_from_this<RpcServer>, public jsonrpc::AbstractServerConnector {
  public:
-  RpcServer(boost::asio::io_context &io, boost::asio::ip::tcp::endpoint ep, addr_t node_addr);
+  enum ServerType { RpcType, GraphQlType };
```
JakubFornadel, line 20: I think enum class is pretty much preferred to basic enum

Dockerfile
JakubFornadel: Oh yeah you are right, you do not have git info when downloading .gz file... Did not realize that :)

src/node/full_node.cpp
```diff
@@ -142,6 +142,12 @@ void FullNode::init() {
           *jsonrpc_io_ctx_, boost::asio::ip::tcp::endpoint{conf_.rpc->address, *conf_.rpc->ws_port}, node_addr);
       jsonrpc_api_->addConnector(jsonrpc_ws_);
     }
+
+    if (conf_.rpc->gql_port) {
+      jsonrpc_http_ = make_shared<net::RpcServer>(
+          *jsonrpc_io_ctx_, boost::asio::ip::tcp::endpoint{conf_.rpc->address, *conf_.rpc->gql_port}, node_addr,
+          final_chain_, net::RpcServer::GraphQlType);
+    }
```
JakubFornadel, line 150: - Is this ok ? Because you are overriding the value of jsonrpc_http_, which was already initiliazed a few lines above in case http_port was specified...
- did you consider also graphql support over websockets ?  
- I guess we want to support both jsonrpc and graphql at the same time ? If yes then we would have to have separate servers and not a single rpc server, which is either jsonrpc or graphql..
- it would be nice to differenciate these ports also in config, we could for example separate jsonrpc and graphql settings or if we keep them both under rpc, then we could have something like  rpc: { jsonrpc: {}, graphql: {} }
- config values validation might be added to (I already created such function) to make sure that people won't accidentaly specify same port for graphql and jsonrpc, http and ws, etc...

JakubFornadel:
nice work overall :) I had a few comments, but most of them are just minor technicalities...

src/node/full_node.cpp
mfrankovi: No, I will fix that. I wanted to have two separate http server on separate ports. I do have already separate ports in the config

mfrankovi: This entire implementation of graphql is not very well tested. The API is maybe 90% implemented. Not 100%. There are some methods I was not sure how to implement. Also I am not sure if all the data formats I am returning are exactly correct. We need to start working with some clients. Either something existing that uses graphql or someone in Taraxa should start building something based on graphql and then we can fix stuff once they are being used.

JakubFornadel: yeah sure, ports naming is just little inconsistent thats what I wanted to sa, if I see config like rpc : { http_port:<port>, ws_port:<port>, gql_port:<port> } it does not say much if gql port is related to http or ws... Something like jsonrpc : { http_port:<port>, ws_port:<port>}, graphql : { http_port:<port>, ws_port:<port>} would be more intuitive I think as it clearly separates jsonrpc and graphql settings

mfrankovi: I agree and I think that we should have a redo of everything in the configuration file. There is a lot of things inside the config file that could be better named and grouped but maybe it would be better to do that in a one commit/PR because each time we make individual changes to existing config properties it creates problems when we want to redeploy nodes to some existing specific configurations

src/network/graphql/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+add_library(taraxa_graphqlservice STATIC
+        gen/TaraxaSchema.h
+        Account.h
+        Block.h
+        Log.h
+        Query.h
+        Transaction.h
+        Mutation.h
+        SyncState.h
+
```
JakubFornadel, line 41: one more minor technicality, whole project uses .hpp naming, let's keep that convention also here I guess

src/node/full_node.cpp
JakubFornadel: ok, @justinsnapp I guess we could put this config refactor task somewhere to baglog and look at it when we have more time...

src/network/rpc/RpcServer.h
```diff
@@ -16,14 +17,19 @@ class RpcHandler;
 
 class RpcServer : public std::enable_shared_from_this<RpcServer>, public jsonrpc::AbstractServerConnector {
  public:
-  RpcServer(boost::asio::io_context &io, boost::asio::ip::tcp::endpoint ep, addr_t node_addr);
+  enum class ServerType { RpcType, GraphQlType };
+
```
JakubFornadel, line 21: one more idea, which I leave to you how you decide :) RpcServer could even become base class and then JsonRpcServer and GraphQlServer might be derived from that and differenciate(override virtual func) "read() ".

JakubFornadel (approved):

src/network/rpc/RpcServer.h
justinsnapp: @JakubFornadel I would think graphql over websocket will be needed for the explorer to function efficiently at high throughput

mfrankovi: Ok, I will add the WS server for graphql as well

src/network/rpc/jsonrpc_ws_server.cpp
```diff
@@ -0,0 +1,77 @@
...
+    LOG(log_tr_) << "Executor missing - WS closed";
+    closed_ = true;
+    return "";
+  }
+
+  return response;
+}
+
+std::shared_ptr<WSSession> JsonRpcWsServer::createSession(tcp::socket &socket) {
+  return std::make_shared<JsonRpcWSSession>(std::move(socket), node_addr_, shared_from_this());
```
JakubFornadel, line 74: as socket is always moved, function parameter should be tcp::socket&& socket so no one accidentally calls this method without knowing that the socket will be moved...

JakubFornadel:
It would be also cool to have doc about our graphql endpoint... There are many tools that can generate it from schema

PR #635, opened 2021-02-11
Fixed and optimized the executor

02p01r:
TBD

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -188,7 +188,7 @@ void PbftManager::update_dpos_state_() {
     } catch (state_api::ErrFutureBlock &c) {
       LOG(log_nf_) << c.what() << ". PBFT period " << dpos_period_ << " is too far ahead of DPOS, need wait!"
                    << " PBFT chain size " << pbft_chain_->getPbftChainSize() << ", have executed chain size "
-                   << pbft_chain_->getPbftExecutedChainSize();
+                   << final_chain_->last_block_number();
       // Sleep one PBFT lambda time
       thisThreadSleepForMilliSeconds(LAMBDA_ms);
     }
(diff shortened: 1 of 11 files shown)


src/dag/dag.cpp
```diff
@@ -494,6 +494,9 @@ std::pair<uint64_t, std::shared_ptr<vec_blk_t>> DagManager::getDagBlockOrder(blk
 
 uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period, vec_blk_t const &dag_order,
                                   const taraxa::DbStorage::BatchPtr &write_batch) {
+  // TODO this function smells. It tries to manage in-memory and persistent state at the same time, which it
```
justinsnapp, line 497: Yes this has long been a stinky part of our code that we have never refactored.

justinsnapp (approved):

src/consensus/pbft_manager.cpp
```diff
@@ -1232,13 +1232,23 @@ bool PbftManager::pushPbftBlock_(PbftBlockCert const &pbft_block_cert_votes) {
...
+  }
+
   // Commit DB
   db_->commitWriteBatch(batch);
   LOG(log_nf_) << node_addr_ << " successful push unexecuted pbft block " << pbft_block_hash << " in period "
                << pbft_period << " into chain! In round " << getPbftRound();
 
-  // Notify executor
-  executor_->cv_executor.notify_one();
+  executor_->execute(pbft_block);
```
JakubFornadel, line 1252: What would happen if i call execute(block) with block that was already processed ?

JakubFornadel:
What is the problem/bug that this PR is supposed to fix ?

src/node/executor.cpp
```diff
@@ -17,171 +17,189 @@ Executor::Executor(addr_t node_addr, std::shared_ptr<DbStorage> db, std::shared_
       db_(db),
       dag_mgr_(dag_mgr),
       trx_mgr_(trx_mgr),
-      final_chain_(final_chain),
-      pbft_chain_(pbft_chain) {
+      final_chain_(final_chain) {
```
JakubFornadel, line 20: Why was pbft_chain_(pbft_chain) deleted ? Does not seem ok

PR #636, opened 2021-02-12
Block producer rewards distribution

JakubFornadel:
## Purpose

Distribute mining rewards based on dag blocks & transactions statistics

go part PR:
https://github.com/Taraxa-project/taraxa-evm/pull/21

@justinsnapp I left rewards distribution disabled. I think we need to decide how to enable it:

- either create HF on final chain as rewards distribution affects state 
- or we wipe out testnet data and just restart it from scratch
- Fixed amount of tokens (2) is created with each new final block. This number should be considered and decided in our tokenomics, which is more economical point of view rather than technical.


JakubFornadel:
It still needs little refactor:
- @justinsnapp I left rewards distribution disabled. I think we need to decide how to enable it:
     - either create HF on final chain as rewards distribution affects state or we 
     - wipe out testnet data and just restart it from scratch
- Fixed amount of tokens (2) is created with each new final block. This number should be considered and decided in our tokenomics, which is more economical point of view rather than technical.
- @02p01r I couldn't make encode_rlp functions work on my new types if I put them anywhere else than encoding_rlp.hpp, which is not the right place for it I guess... I will try to fix it but maybe you will see where is the problem very quickly !!!

src/chain/state/state_api.cpp
```diff
@@ -164,13 +165,16 @@ StateDescriptor StateAPI::get_last_committed_state_descriptor() const {
   return ret;
 }
 
-StateTransitionResult const& StateAPI::transition_state(EVMBlock const& block,  //
-                                                        RangeView<EVMTransaction> const& transactions,
-                                                        RangeView<UncleBlock> const& uncles) {
+StateTransitionResult const& StateAPI::transition_state(EVMBlock const& block,  // PBFT blocks
```
02p01r, line 168: why blockS when it's just one block?

src/dag/dag_block.hpp
```diff
@@ -9,15 +11,23 @@ using VdfSortition = vdf_sortition::VdfSortition;
...
-  uint64_t timestamp_ = 0;
-  vdf_sortition::VdfSortition vdf_;
-  mutable addr_t cached_sender_;  // block creater
+ public:
+  /**
+   * @brief list of Dag block RLP fields
+   *
+   * @note internal values cannot be changed and PIVOT must have value 0 as that is rlp fixed fields structure
+   */
+  enum RlpField : uint8_t {
```
02p01r, line 20: this enum is just an additional verbosity without added value. It was pretty obvious what field has what position in the encoding/decoding code

src/node/executor.cpp
```diff
@@ -175,7 +212,7 @@ void Executor::execute_(PbftBlock const &pbft_block) {
     ws_server_->newEthBlock(new_eth_header);
   }
 
-  transactions_tmp_buf_.clear();
+  clearTmpContainers();
```
02p01r, line 215: you created this function just for the sake of giving a name to a block of code? seems excessive to me

src/dag/dag_block.hpp
JakubFornadel: well it was not so obvious and writing things like rlp[3], etc... in code is just terrible. In case you need to work with that rlp you have to chcek the structure, think about it find the index and use that. Now you can just use enum, which is much easier to read and use.

src/node/executor.cpp
```diff
@@ -86,66 +93,96 @@ void Executor::execute_(PbftBlock const &pbft_block) {
...
-      // Non-executed trxs
-      auto const &trx = transactions_tmp_buf_.emplace_back(&trx_db_results[1 + i * 2], dev::eth::CheckTransaction::None,
-                                                           true, h256(db_query.get_key(1 + i * 2)));
-      if (replay_protection_service_->is_nonce_stale(trx.sender(), trx.nonce())) {
-        transactions_tmp_buf_.pop_back();
+
+      // Non-executed txs
+      auto tx = dev::eth::Transaction(&trx_db_results[tx_result_idx], dev::eth::CheckTransaction::None, true, tx_hash);
+
+      if (replay_protection_service_->is_nonce_stale(tx.sender(), tx.nonce())) {
```
02p01r, line 145: we conventionally call transactions "trx" - haven't you noticed?

src/node/executor.cpp
JakubFornadel: no i created it to clear all containers that need to be cleared and have it in one place. seems much better to me

src/node/executor.cpp
JakubFornadel: you are right, will fix that

src/node/executor.cpp
```diff
@@ -86,66 +93,96 @@ void Executor::execute_(PbftBlock const &pbft_block) {
...
-    unique_trxs.reserve(transactions_tmp_buf_.capacity());
+
     for (auto const &dag_blk_raw : dag_blks_raw) {
-      for (auto const &trx_h : DagBlock::extract_transactions_from_rlp(RLP(dag_blk_raw))) {
-        if (!unique_trxs.insert(trx_h).second) {
+      const auto &dag_blk_rlp = RLP(dag_blk_raw);
+      DagBlock dag_block(dag_blk_rlp);  // Full block has to be decoded from rlp to get block author from signature
+      // TODO: Check if dag block is validated to consist at least 1 tx. Is this checked before ???
+
+      addr_t dag_block_author = dag_block.getSender();
```
02p01r, line 104: const ref?

src/node/executor.cpp
JakubFornadel: but yeah if i just put clear method for all thoise containers here it would the the same effect, I just found it better to have it grouped in function... technicality...

02p01r (approved):
Overall it looks fine. I didn't like the verbose coding style that you maintain. Especially the amount of comments and explanations that seem to be too eager at this point, and just add clutter. This is not an open source library, and at the implementation level I think you can count on your teammates ability to read and understand the code. At the higher levels I agree that we should document. Also, I think we shouldn't much document things that are rather experimental - just wastes time and lines of code. 
Also, please don't use word "miner" :D

src/dag/dag_block.hpp
02p01r: I will replace it with a template-based encoding/decoding like in `enc_rlp` - this way it's clean and no need for enums

src/node/executor.cpp
02p01r: I'm fine with it, but I wouldn't have written it like this myself :)

02p01r:
Also @JakubFornadel don't worry about the RLP stuff, I will revisit all of it it soon.

JakubFornadel:
> Overall it looks fine. I didn't like the verbose coding style that you maintain (but I don't strongly object that). Especially the amount of comments and explanations that seem to be too eager at this point, and just add clutter. This is not an open source library, and at the implementation level I think you can count on your teammates ability to read and understand the code. At the higher levels I agree that we should document. Also, I think we shouldn't much document things that are rather experimental - just wastes time and lines of code.
> Also, please don't use word "miner" :D

Ok, I guess we have totally different approaches and opinions about how high-quality code should look like. Developers change over time, so when someone new comes and see 100 lines of code without any whitespace and no single comment in it, which seems like common habit here it is just pure pain to go through such code. Will all my respect, I will continue commenting stuff as it adds up let's say max 10% of time, but it saves so much time for other devs that overall properly documented code saves money for project. I would like to see also others do that but that is not in my competence to enforce that. 

When I first saw for example that hardcoded constants for rlp I spend decent amount of time looking at the struct it as was something new and realizing ok this is index I need. With enum and comments I would spend 30 seconds on it... Just one of many examples.

Now I want get little bit into the pbft, guess how "easy" that is without any comments ?

src/dag/dag_block.hpp
JakubFornadel: cool, looking forwards for that

JakubFornadel:
fixed according comments suggestions

libraries/core_libs/consensus/src/pbft/block_proposer.cpp
```diff
@@ -156,10 +157,14 @@ void BlockProposer::proposeBlock(DagFrontier&& frontier, level_t level, SharedTr
   std::transform(trxs.begin(), trxs.end(), std::back_inserter(trx_hashes),
                  [](std::shared_ptr<Transaction> const& t) { return t->getHash(); });
 
+  // Votes candidates for rewards from rewards_votes_, can include max k_max_rewards_votes_in_block_(100) unique votes
```
gaoqi8649, line 160: The current committee size is 1000, do you mean max 1000 unique votes?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1556,29 +1593,51 @@ bool PbftManager::pushPbftBlock_(SyncBlock &sync_block, vec_blk_t &dag_blocks_or
...
-  finalize_(*sync_block.pbft_blk, move(dag_blocks_order));
+  finalize_(std::move(sync_block));
 
   // Reset proposed PBFT block hash to False for next pbft block proposal
   proposed_block_hash_ = std::make_pair(NULL_BLOCK_HASH, false);
   db_->savePbftMgrStatus(PbftMgrStatus::ExecutedBlock, true);
   executed_pbft_block_ = true;
+
+  // Reset 2t+1 cert votes in rewards votes(votes that should be rewarded in the next pbft period)
+  rewards_votes_->resetVotes(std::move(cert_votes_hashes), pbft_block_hash);
```
gaoqi8649, line 1626: There are only 2t+1 cert votes because 2t+1 cert votes are enough for validate PBFT blocks syncing. But there are may more cert votes in verified queue. If you are going to rewards all cert votes, you may need to write a new function to return all cert votes in verified queue.

libraries/core_libs/network/src/tarcap/packets_handlers/votes_dag_sync_packet_handler.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+    if (vote_weighted_index >= voter_dpos_votes_count) {
+      LOG(log_dg_) << "Account " << voter << " was not eligible to vote. Vote: " << vote;
+      continue;
+    }
+
+    // TODO: pbft_mgr_->getPreviousPbftPeriodDposTotalVotesCount and
+    // pbft_mgr_->getPreviousPbftPeriodSortitionThreshold() mght be incosistent with current_pbft_period - 1
+    if (auto result = vote->validate(pbft_mgr_->getPreviousPbftPeriodDposTotalVotesCount(),
+                                     pbft_mgr_->getPreviousPbftPeriodSortitionThreshold());
```
gaoqi8649, line 92: You're right, getPreviousPbftPeriodDposTotalVotesCount and getPreviousPbftPeriodSortitionThreshold might be inconsistent.
I think one solution we could save all DPOS total votes count and sortition threshold for all periods in DB, not only save for previous period. And add period in vote structure. Then could use vote->getPeriod() to get DPOS total votes count and sortition threshold.

gaoqi8649:
Since our 6 consensus nodes have majority votes, do we will except votes for our consensus nodes and just count players votes for rewarding? Otherwise our nodes will take much more rewards.

PR #637, opened 2021-02-16
Consensus Fix 

gaoqi8649:
Change back with Algorand Paper


justinsnapp (approved):
Thanks for this!  I should never have commented this out (and definitely not without leaving a comment in the comment!)

src/chain/chain_config.cpp
```diff
@@ -57,8 +57,8 @@ decltype(ChainConfig::predefined_) const ChainConfig::predefined_([] {
     cfg.vdf.threshold_selection = 0x8000;
     cfg.vdf.threshold_vdf_omit = 0x7200;
     cfg.vdf.difficulty_min = 16;
-    cfg.vdf.difficulty_max = 21;
-    cfg.vdf.difficulty_stale = 22;
+    cfg.vdf.difficulty_max = 19;
+    cfg.vdf.difficulty_stale = 21;
```
gaoqi8649, line 61: These are basic on mathematic model setting the values. If the difficulty interval is small, that will propose blocks constantly. And lose the functionality of sortition.

JakubFornadel (approved):

PR #639, opened 2021-02-19
Made a block of code (in executor) work as was actually intended

02p01r:
See https://github.com/Taraxa-project/taraxa-node/pull/639/files#diff-17d5942e62aa35eb95395beed488fff05c082ec5981722b0dea4e62c6ee16b8aR77-R80
It was supposed to execute everything until (exclusive) the to_execute_ pbft block that was set by pbft manager, and then execute to_execute_ itself (without getting it from db). But I coded it backwards


src/node/executor.cpp
```diff
@@ -54,8 +54,11 @@ void Executor::stop() {
 }
 
 void Executor::execute(std::shared_ptr<PbftBlock> blk) {
-  std::unique_lock l(mu_);
-  to_execute_ = move(blk);
+  assert(final_chain_->last_block_number() < blk->getPeriod());
+  {
```
02p01r, line 58: Added brackets to exclude `notify_one` from the critical section, because it doesn't have to be done under the lock, and in fact it would be a pessimization

src/node/executor.cpp
```diff
@@ -54,8 +54,11 @@ void Executor::stop() {
 }
 
 void Executor::execute(std::shared_ptr<PbftBlock> blk) {
-  std::unique_lock l(mu_);
-  to_execute_ = move(blk);
+  assert(final_chain_->last_block_number() < blk->getPeriod());
```
02p01r, line 57: @JakubFornadel as requested before

justinsnapp (approved):
LGTM!

gaoqi8649 (approved):

PR #640, opened 2021-02-20
Cli changes to handle chain config, and deploy public testnet

justinsnapp:
## Purpose

Make the python based cli integrated with the node more functional.


sfrique:
Why did you add a `prop-override` option instead of just updating value if the option passed as an argument?

PR #641, opened 2021-02-23
Fix syncing issue

gaoqi8649:


justinsnapp (approved):
Great, I think its right to carry the force argument here into PBFT code.   Good!

PR #645, opened 2021-03-04
Fix PBFT next voting 2 values in the same step issue

gaoqi8649:
Fix PBFT next voting 2 values in the same step issue
* Implement next votes for previous PBFT round class and integrate with PBFT manager
* Implement next votes read/write in DB APIs and add unit tests
* Implement next votes syncing. Not only sync when node is behind to peer PBFT round,  but also will sync when node has a short next votes size in previous PBFT round
* Process synced next votes. If node is behind to peer, push synced next votes into unverified queue. If node has same PBFT round with peer but has a short previous next votes size, verify the votes and update own previous next votes structure
* Add next votes bundle syncing unit tests

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -626,8 +626,7 @@ void PbftManager::identifyBlock_() {
         placeVote_(leader_block.first, soft_vote_type, round, step_);
       }
     }
-  } else if (auto voted_value = previous_round_next_votes_->getVotedValue();
-             round >= 2 && voted_value != NULL_BLOCK_HASH) {
+  } else if (round >= 2 && voted_value != NULL_BLOCK_HASH) {
     if (shouldSpeak(soft_vote_type, round, step_)) {
       LOG(log_dg_) << "Soft voting " << voted_value << " from previous round";
       placeVote_(voted_value, soft_vote_type, round, step_);
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -253,7 +253,6 @@ bool PbftManager::resetRound_() {
     // Update in DB first
     auto batch = db_->createWriteBatch();
     db_->addPbftMgrStatusToBatch(PbftMgrStatus::executed_in_round, false, batch);
-    db_->addPbftMgrStatusToBatch(PbftMgrStatus::cert_voted_in_round, false, batch);
     db_->addPbftMgrVotedValueToBatch(PbftMgrVotedValue::own_starting_value_in_round, NULL_BLOCK_HASH, batch);
     db_->addPbftMgrStatusToBatch(PbftMgrStatus::next_voted_null_block_hash, false, batch);
     db_->addPbftMgrStatusToBatch(PbftMgrStatus::next_voted_soft_value, false, batch);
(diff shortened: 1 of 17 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -530,9 +490,7 @@ bool PbftManager::stateOperations_() {
   // Reset continue finish polling state
   continue_finish_polling_state_ = false;
 
-  // NOTE: PUSHING OF SYNCED BLOCKS CAN TAKE A LONG TIME
-  //       SHOULD DO BEFORE WE SET THE ELAPSED TIME IN ROUND
-  // push synced pbft blocks into chain
+  // NOTE: PUSHING OF SYNCED BLOCKS CAN TAKE A LONG TIME SHOULD DO BEFORE WE SET THE ELAPSED TIME IN ROUND
```
justinsnapp, line 493: This is no longer true right? Seems like this comment should be removed. :-)

src/consensus/vote.cpp
```diff
@@ -309,4 +315,162 @@ bool VoteManager::pbftBlockHasEnoughValidCertVotes(PbftBlockCert const& pbft_blo
...
+
+  // Protect for malicious players. If no malicious players, will include either/both NULL BLOCK HASH and a non NULL
+  // BLOCK HASH
+  LOG(log_nf_) << "PBFT 2t+1 is " << TWO_T_PLUS_ONE;
+  {
+    upgradableLock_ lock(access_);
+    auto it = next_votes_.begin();
+    while (it != next_votes_.end()) {
+      if (it->second.size() >= TWO_T_PLUS_ONE) {
+        LOG(log_nf_) << "Voted PBFT block hash " << it->first << " has " << it->second.size() << " next votes";
```
justinsnapp, line 409: Should we not add round and step to this log message to be more useful?

gaoqi8649: This is vote model, doesn't know the current round and step. The current round and step are in the pbft manager model.
And for the vote itself, has print the vote whole info(include voted round and step info) in the above.

src/network/taraxa_capability.cpp
```diff
@@ -378,23 +393,60 @@ bool TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
...
+          }
+
+          auto pbft_current_round = pbft_mgr_->getPbftRound();
+          auto pbft_previous_round_next_votes_size = next_votes_mgr_->getNextVotesSize();
+          auto peer_pbft_round = next_votes[0].getRound() + 1;
+
+          if (pbft_current_round < peer_pbft_round) {
+            // Add into votes unverified queue
+            vote_mgr_->addVotes(next_votes);
+          } else if (pbft_current_round == peer_pbft_round && pbft_previous_round_next_votes_size < next_votes_count) {
```
justinsnapp, line 441: What if our node has 3 votes and other node has 2 votes, but those two votes are different than our 3?

src/network/taraxa_capability.cpp
```diff
@@ -201,38 +206,48 @@ bool TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
...
+                                       << peer->pbft_previous_round_next_votes_size_;
+
           if (peer->syncing_) {
-            LOG(log_dg_pbft_sync_) << "Other node is behind, prevent gossiping " << _nodeID
-                                   << "Our pbft chain size: " << pbft_chain_size
-                                   << " Peer pbft chain size: " << peer_pbft_chain_size;
-            if (syncing_ && peer_syncing_pbft == _nodeID) {
-              // We are currently syncing to a node that just reported it is
-              // not synced, force a switch to a new node
+            LOG(log_dg_pbft_sync_) << "Peer node has a short PBFT chain, prevent gossiping to " << _nodeID
```
justinsnapp, line 234: This log should be removed I think, because I don't see how we are turning off gossiping to this node?

src/network/taraxa_capability.cpp
```diff
@@ -201,38 +206,48 @@ bool TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
...
+                                   << ". Own pbft chain size " << pbft_chain_size << " Peer pbft chain size "
+                                   << peer->pbft_chain_size_;
+            if (syncing_ && peer_syncing_pbft_ == _nodeID) {
+              // We are currently syncing to a node that just reported it is not synced, force a switch to a new node
```
justinsnapp, line 238: Could instead a log here saying we are restarting syncing because currently syncing from node that is itself syncing.

src/network/taraxa_capability.cpp
gaoqi8649: if our round < peer round, put into unverified queue, and will check the vote hash is exist
if our round == peer round && our previous next votes size < peer, in this case, at least they will have (t+1) same next votes. And our node adds the missing next votes

src/network/taraxa_capability.cpp
gaoqi8649: Yeah, I agree

src/consensus/pbft_manager.cpp
```diff
@@ -1115,12 +1115,13 @@ bool PbftManager::comparePbftBlockScheduleWithDAGblocks_(blk_hash_t const &pbft_
 
     pbft_block = db_->getPbftCertVotedBlock(pbft_block_hash);
     if (!pbft_block) {
+      auto round = getPbftRound();
```
justinsnapp, line 1118: Ah.  Excellent!

src/network/taraxa_capability.cpp
gaoqi8649: OK

justinsnapp (approved):

justinsnapp (approved):

PR #646, opened 2021-03-09
Added conan package manager

nikonok:
## Purpose

We need to use package manager instead of using 3party libraries as submodules or system libraries.

## For reviewers
How to install conan:
1) Install pip3 if you don't have one
2) `pip3 install conan`

Commands to test it:
1) `conan install -if build --build missing .` - install conan deps
2) `conan build -sf . -bf build .` - build project(or just use cmake for it)

## Submodule request link
https://github.com/Taraxa-project/taraxa-aleth/pull/9

## Changes
* Added conanfile.py
* Move submodules build to cmake targets
* Updated docs
* Removed unused submodules
* Fixed some bugs

## Some problems that PR doesn't solve
1) CryptoPP gives some warnings at linking executable stage. Because we build our project with `-fvisibility=hidden`, but CryptoPP don't have such option at conan. The solution is to use dynamic build for that library, but I just don't like it. It's only on Mac OS (clang).
2) I need to get `get_target_property(TARAXA_VDF_INCUDE taraxa-vdf INTERFACE_INCLUDE_DIRECTORIES)` from imported cmake target, it can't be included itself. It not a problem, but looks not good.


justinsnapp (commented):
Awaiting fixes for rocksDB version 6

.gitignore
```diff
@@ -9,4 +9,11 @@ cmake-*
...
-.DS_Store
\ No newline at end of file
+.DS_Store
+
+# build directory
+build
+
+# clang-format and clang-tidy binaries
+clang-format
+clang-tidy
```
JakubFornadel, line 19: - why would anyone have clang-* binaries in project root ? 
- for build directories we are using cmake-* naming convetion, so for example debug goes to cmake-build-debug etc...

CMakeModules/scripts/copydlls.cmake
```diff
@@ -0,0 +1,24 @@
+# this module expects
```
JakubFornadel, line 1: Isn't dll  windows only concept ? what do we use it for ?

conanfile.py
```diff
@@ -0,0 +1,132 @@
...
+        self.requires("cryptopp/8.4.0")
+        self.requires("gtest/1.10.0")
+        self.requires("rocksdb/6.8.1")
+        self.requires("jsoncpp/1.9.4")
+        self.requires("gmp/6.2.0")
+        self.requires("mpfr/4.0.2")
+        self.requires("snappy/1.1.8")
+        self.requires("zstd/1.4.4")
+        self.requires("lz4/1.9.2")
+
```
JakubFornadel, line 62: Just thinking out loud now. Maybe we could have a separate file with lib versions for this script. Might be cleaner but maybe unnecessary - don't know :)

.gitignore
nikonok: 1) Because build scripts on macos downloads that binaries to project dir, and I added such option to conan too.
2) Okay, I don't know that before, so I can change it

doc/building.md
```diff
@@ -61,16 +47,34 @@ will build out of the box without further effort:
...
-    cd cmake-build
+Add bincrafters repository to conan:
+
+    conan remote add bincrafters https://api.bintray.com/conan/bincrafters/public-conan
+
+In general to build you need to:
+
+    conan install -if build --build missing .
+    conan build -bf build -sf . .
+
```
JakubFornadel, line 58: so what is the difference if I build using cmake ../ vs conan build ?

src/taraxad/CMakeLists.txt
```diff
@@ -4,5 +4,5 @@ target_link_libraries(taraxad PRIVATE app_base)
 set(TARAXAD_INSTALL_DIR "/usr/local/bin" CACHE PATH "Aplication install directory")
 set(TARAXAD_CONF_INSTALL_DIR "/etc/taraxad" CACHE PATH "Aplication configuration install directory")
 
-install(PROGRAMS ${CMAKE_CURRENT_BINARY_DIR}/taraxad DESTINATION ${TARAXAD_INSTALL_DIR} COMPONENT taraxad)
+install(PROGRAMS ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/taraxad DESTINATION ${TARAXAD_INSTALL_DIR} COMPONENT taraxad)
```
JakubFornadel, line 7: is this change somehow related to conan stuff ?

doc/building.md
nikonok: Here is about 3 ways to build project with conan:
1. `conan install <...> && conan build <..>`. In such way will be called conan installed in the system. Project will be built in release mode with threads(like `make -j`).
2. `cmake .. && cmake --build .`. That way will download conan linke cmake script, you can find link in CMakeLists.txt . I found only one problem with such conan: it doesn't support Apple Clang 12.0
3. `conan install <...> && cmake --build .`. It will call conan like in 2, but libraries are already installed.

If it first build of project prefer `1`, but after it you can use all that you want.

src/taraxad/CMakeLists.txt
nikonok: Yeah, now all binaries will be at the <project_build_folder>/bin. So we need to change the way how we find bin files.

This is one of conan's requirements, as well as "specifying the build type" (Debug/Release/RelWithDebInfo)

conanfile.py
nikonok: This file is intended to list everything on the project, not in the style of conan to write multiple files =)

CMakeModules/scripts/copydlls.cmake
nikonok: Oh, I think this is superfluous, just accidentally captured from aleth along with secp256k1 =)

justinsnapp (approved):

PR #656, opened 2021-03-19
Aleth networking code upgrade + fixes and improvements in our networking code

02p01r:
TBD

Also see https://github.com/Taraxa-project/taraxa-aleth/pull/10

This PR introduces https://github.com/Taraxa-project/taraxa-node/issues/695


src/consensus/pbft_manager.hpp
```diff
@@ -145,8 +145,7 @@ class PbftManager {
   std::shared_ptr<VoteManager> vote_mgr_ = nullptr;
   std::shared_ptr<PbftChain> pbft_chain_ = nullptr;
   std::shared_ptr<DagManager> dag_mgr_ = nullptr;
-  std::shared_ptr<Network> network_ = nullptr;
-  std::shared_ptr<TaraxaCapability> capability_ = nullptr;
+  util::WeakRef<Network> network_;
   std::shared_ptr<DagBlockManager> dag_blk_mgr_;
```
justinsnapp, line 149: Why not initialize these as well to nullptr?

02p01r: 1) That `= nullptr` stuff is redundant and should be cleaned up 
2) WeakRef, which I wrote, is not like std library smart pointers. I didn't create conversion rules/constructors for pointers, including nullptr.  @justinsnapp

tests/network_test.cpp
```diff
@@ -44,10 +44,8 @@ struct NetworkTest : BaseTest {};
 // Test creates two Network setup and verifies sending block
 // between is successfull
 TEST_F(NetworkTest, transfer_block) {
-  std::shared_ptr<Network> nw1(
-      new taraxa::Network(g_conf1->network, g_conf1->chain.dag_genesis_block.getHash().toString(), addr_t()));
-  std::shared_ptr<Network> nw2(
-      new taraxa::Network(g_conf2->network, g_conf2->chain.dag_genesis_block.getHash().toString(), addr_t()));
+  std::unique_ptr<Network> nw1(new taraxa::Network(g_conf1->network));
+  std::unique_ptr<Network> nw2(new taraxa::Network(g_conf2->network));
```
JakubFornadel, line 49: please use make_unique. "New" is not really part of modern c++ anymore...

src/network/taraxa_capability.hpp
```diff
@@ -103,68 +103,31 @@ class TaraxaPeer : public boost::noncopyable {
...
   void syncPeerPbft(NodeID const &_nodeID, unsigned long height_to_sync);
   void restartSyncingPbft(bool force = false);
   void delayedPbftSync(NodeID _nodeID, int counter);
   std::pair<bool, blk_hash_t> checkDagBlockValidation(DagBlock const &block);
-  bool interpretCapabilityPacket(NodeID const &_nodeID, unsigned _id, RLP const &_r) override;
-  bool interpretCapabilityPacketImpl(NodeID const &_nodeID, unsigned _id, RLP const &_r);
+  void interpretCapabilityPacket(NodeID const &_nodeID, unsigned _id, RLP const &_r) override;
+  void interpretCapabilityPacket_(NodeID const &_nodeID, unsigned _id, RLP const &_r);
+  void interpretCapabilityPacketImpl(NodeID const &_nodeID, unsigned _id, RLP const &_r);
   void onDisconnect(NodeID const &_nodeID) override;
```
JakubFornadel, line 131: Can we refactor these names to something meaningful ? Functions/variables names should be as self explanatory as possible. I would have absolutely no clue what might be difference between these 3 methods when named interpretCapabilityPacket, interpretCapabilityPacket_  and interpretCapabilityPacketImpl...

src/network/taraxa_capability.cpp
```diff
@@ -70,29 +115,41 @@ std::pair<bool, blk_hash_t> TaraxaCapability::checkDagBlockValidation(DagBlock c
...
+    insertPeer(_nodeID, std::make_shared<TaraxaPeer>(_nodeID));
+    sendStatus(_nodeID, true);
+  });
+}
 
-  insertPeer(_nodeID, std::make_shared<TaraxaPeer>(_nodeID));
-  sendStatus(_nodeID, true);
+void TaraxaCapability::interpretCapabilityPacket(NodeID const &_nodeID, unsigned _id, RLP const &_r) {
+  ba::post(io_service_, [=, bb = _r.data().toBytes()] {
+    RLP r(bb);
```
JakubFornadel, line 130: bb ? what does that mean ? Isn't it possible to simply copy _r, e.g. [packet_rlp = _r] ?

src/util/thread_pool.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+  auto &unsafe_get_io_context() { return ioc_; }
+
+  void start();
+  void stop();
+
+  template <typename... T>
+  auto post(T &&... args) {
+    return boost::asio::post(ioc_, std::forward<T>(args)...);
+  }
```
JakubFornadel, line 30: Why put half of the class methods definitions into the cpp and half leave in header ?

src/util/thread_pool.hpp
```diff
@@ -0,0 +1,41 @@
...
+
+ public:
+  explicit ThreadPool(size_t num_threads, bool _start = true);
+  ~ThreadPool();
+
+  ThreadPool(ThreadPool const &) = delete;
+  ThreadPool &operator=(ThreadPool const &) = delete;
+
+  auto &unsafe_get_io_context() { return ioc_; }
+
```
JakubFornadel, line 23: I Dont think this is a good design. I went through code how you use this threadpool and it is widely used just by accessing io_service, which is retrieved by reference from this method... Then this io_context is sent to the generic boost::asio::post methods and it is very unclear that tasks are actually going to be processed by this threadpool. I had to track down where is that io_service from and found out that  you sent reference to io_context to the taraxa)capability constructor and then use post methos from util/boost_asio.hpp... It is just too complicated. Why not just sent this threadpool to the taraxa-capability and use post methods on it ? io_context should be totally encapsulated inside with no option to access it but class methods

src/network/network.cpp
```diff
@@ -7,201 +7,219 @@
...
+  if (!network_file_.empty()) {
+    networkData = contents(network_file_.string());
+  }
+  if (!networkData.empty()) {
+    host_ = std::make_shared<dev::p2p::Host>(net_version, net_conf, move(taraxa_net_conf), &networkData);
+  } else {
+    host_ = std::make_shared<dev::p2p::Host>(net_version, makeENR(key, net_conf), net_conf, move(taraxa_net_conf));
+  }
+  taraxa_capability_ = std::make_shared<TaraxaCapability>(*host_, tp_.unsafe_get_io_context(), conf_, db, pbft_mgr,
+                                                          pbft_chain, vote_mgr, next_votes_mgr, dag_mgr, dag_blk_mgr,
```
JakubFornadel, line 83: just use reference to tp_ inside TaraxaCapability instead f io_context. Much cleaner and easier to understand

src/network/network.cpp
```diff
@@ -7,201 +7,219 @@
...
+  }
+  assert(!host_->isStarted());
+  host_->start();
+  if (!boot_nodes_.empty()) {
+    for (auto const &[k, v] : boot_nodes_) {
+      host_->addNode(k, v);
+    }
+    // Every 30 seconds check if connected to another node and refresh boot nodes
+    util::post_periodic(tp_.unsafe_get_io_context(), {30000}, [this] {
+      // If node count drops to zero add boot nodes again and retry
```
JakubFornadel, line 114: post_periodic should be rather threadpool method and it would be obvious that it is processed by threadpool

src/network/network.cpp
```diff
@@ -7,201 +7,219 @@
...
+    LOG(log_dg_) << "On new block verified:" << blk->getHash().toString();
+  });
 }
 
-void Network::sendTransactions(NodeID const &id, std::vector<taraxa::bytes> const &transactions) {
-  taraxa_capability_->sendTransactions(id, transactions);
-  LOG(log_dg_) << "Sent transactions:" << transactions.size();
+void Network::onNewTransactions(std::vector<taraxa::bytes> transactions) {
+  tp_.post([=, transactions = std::move(transactions)] {
+    taraxa_capability_->onNewTransactions(transactions, true);
```
JakubFornadel, line 153: all  uitl::post_periodic and ba::post calls inside tarxa capability might be replaced by   tp_.pos and  tp_.pos_periodic

src/network/taraxa_capability.cpp
```diff
@@ -70,29 +115,41 @@ std::pair<bool, blk_hash_t> TaraxaCapability::checkDagBlockValidation(DagBlock c
 }
 
 void TaraxaCapability::onConnect(NodeID const &_nodeID, u256 const &) {
-  LOG(log_nf_) << "Node " << _nodeID << " connected";
-  cnt_received_messages_[_nodeID] = 0;
-  test_sums_[_nodeID] = 0;
+  ba::post(io_service_, [=] {
+    LOG(log_nf_) << "Node " << _nodeID << " connected";
```
JakubFornadel, line 119: tp_.post...

src/util/boost_asio.hpp
```diff
@@ -0,0 +1,42 @@
+#pragma once
+
```
JakubFornadel, line 2: This whole file is basically unnecessary. If you add post_periodic function to the the threadpool and send threadpool object into the taraxa capability by reference, this can be deleted

JakubFornadel:
I think Oleh did a great job especially with taraxa-aleth merging, but I find this design around the new thread pool pretty much over complicated. It could be much cleaner, easier to read and also easier to use - read my comments in code

src/network/taraxa_capability.cpp
```diff
@@ -130,499 +187,495 @@ bool TaraxaCapability::interpretCapabilityPacket(NodeID const &_nodeID, unsigned
...
+        LOG(log_dg_pbft_sync_) << "Syncing PBFT is completed";
+        // We are pbft synced with the node we are connected to but
+        // calling restartSyncingPbft will check if some nodes have
+        // greater pbft chain size and we should continue syncing with
+        // them
+        restartSyncingPbft(true);
+        // We are pbft synced, send message to other node to start
+        // gossiping new blocks
+        if (!syncing_) {
+          sendSyncedMessage();
```
gaoqi8649, line 632: The synced message will clean up peer clearAllKnownBlocksAndTransactions.
Why need to clear all DAG blocks and transactions?
This is inside PbftBlockPacket. Why don't clear PBFT blocks and votes?

02p01r:
@JakubFornadel All right, I agree with your comments about the cosmetics, but I think you are not fair to demand a refactoring of the old code (some of your comments indicate the demand).
The thread pool comments make sense, and I also considered this after I made this PR, but I thought it could be done separately.
I have a more general idea of what to do with the threadpool/post/post_periodic API; just putting it all inside threadpool is not good enough for me.

src/util/thread_pool.hpp
02p01r: This is a tip of an iceberg. I copied this code from the branch I'm working on in parallel, and there `unsafe_get_io_context` is justified, because I use the thread pool class as a unified scheduler, and unfortunately some our dependencies can only work with `io_context` interface. I called it `unsafe....` precisely to suggest that it is bad but we have to live with it.

src/util/thread_pool.hpp
02p01r: The templated methods should be in headers (read the basics to find why).
For the rest, I kind of agree, except I think leaving one-liners in headers is not too bad (depends on how much template expansions that might trigger).
The answer to your question - because it's not a huge problem, and because it's more convenient to work with logically same code in one file than in two files, I usually first write everything in header, but then sometimes forget to extract to .cpp file when I'm done with it.

tests/network_test.cpp
02p01r: Again, you demand that I refactor the old code. The moral aspect here is: if I do it - I'm a good person, but I'm not required to do it. And I actually do it quite often, but not *always*

src/network/taraxa_capability.hpp
02p01r: This is a demand to refactor the old code

src/network/taraxa_capability.cpp
02p01r: bb - ByteBuffer. Simply copying the RLP object will be a bug, because RLP is a reference type. I thought it is quite clear what that variable represents. What is your suggestion? `rlp_bytes`? `_r_bytes`?

src/network/taraxa_capability.cpp
02p01r: I think this change is obsolete, I never wrote that code, perhaps this was displayed as a diff because at some point this branch diverged from develop

02p01r:
So again on the threadpool comments - I see that the net result looks weird, but this is because that code underwent evolution, and because I'm going to follow up on the networking code soon again, I thought it's fine to postpone the improvements, similar to what you suggest

justinsnapp (approved):

justinsnapp (approved):
Not perfect, but a significant improvement over current develop branch, so lets work from this going forward!

PR #671, opened 2021-04-01
Log network thread queue size

02p01r:
- Log network threadpool queue size
- Moved `post` methods to the threadpool @JakubFornadel


src/network/network.cpp
```diff
@@ -40,6 +39,21 @@ Network::Network(NetworkConfig const &config, std::filesystem::path const &netwo
...
+    assert(log_interval.count() % sleep_duration.count() == 0 && sleep_duration < log_interval);
+    auto num_sleeps_left = 0;
+    while (!stopped_) {
+      if (num_sleeps_left != 0) {
+        std::this_thread::sleep_for(sleep_duration);
+        --num_sleeps_left;
+        continue;
+      }
+      num_sleeps_left = log_interval / sleep_duration;
+      LOG(log_nf_) << "NET_TP_NUM_PENDING_TASKS=" << tp_.num_pending_tasks();
```
02p01r, line 54: @justinsnapp logged it to this channel cause found this more convenient to code, I guess there's no bid difference

justinsnapp: OK thanks!

JakubFornadel (approved):

PR #674, opened 2021-04-02
Fix transaction queues and cache

justinsnapp:
## Purpose

_Provide any information reviewers might need to have context on your changes_

## For reviewers
_Describe anything you want to call out to the reviewers e.g. areas to focus on, decisions you made, specific feedback you are looking for, etc._

## Tests
_Describe what you did to test the changes you made. Include evidence such as screenshots or terminal commands and output._

## Version Notes ##
{ major | minor | patch }

_Decide whether this is a major, minor, or patch version change. Major is for backwards incompatible changes (This is only relevant once the service is live and in production and the major version is at least 1. While the major version is 0, it is ok to have backwards incompatible changes in a minor version.). Minor is for additional features. Patch is for bug fixes._

_Describe the changes in this version. This is essentially the details you would include in the squash commit message._

_NOTE: Include the version notes in the squash commit message_


config/networks/testnet.json
```diff
@@ -73,18 +73,18 @@
...
       "range": "0xa"
     },
     "vdf": {
-      "difficulty_max": "0x15",
+      "difficulty_max": "0x12",
       "difficulty_min": "0x10",
       "threshold_selection": "0xbffd",
       "threshold_vdf_omit": "0x6bf7",
-      "difficulty_stale": "0x16",
+      "difficulty_stale": "0x13",
```
gaoqi8649, line 87: If only for test purpose and not permanence change, should not change here. Let's change it in config file

src/util/util.hpp
```diff
@@ -307,6 +310,12 @@ class ExpirationCache {
   }
 
  private:
+  virtual void eraseOldest() {
+    for (auto i = 0; i < delete_step_; i++) {
+      cache_.erase(expiration_.front());
+      expiration_.pop_front();
+    }
+  }
```
gaoqi8649, line 318: Looks like the change will do the same thing

PR #675, opened 2021-04-02
Separate tests from build

sfrique:
Remove ctest from docker build
Add new image build for ctest
Add steps to run the ctest on Jenkins instead of docker build
Add artifacts to jenkins on failure

## Purpose

Run ctest separated from docker build so we can capture coredums.

## For reviewers

I am trying to get the temp folder, this also seems to contain the databases and some other files, for now, I don't know how big they will be, if they are too big, we may not able to get everything.

Does this folder help in figuring out the problem? If not I can get the coredump only.

ps: The new image build should be cached so not much time added

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -107,9 +107,7 @@ RUN mkdir $BUILD_OUTPUT_DIR \
              ../ \
     && make -j$(nproc) all \
     && make install \
-    && strip bin_install/taraxad \
-    && cd tests/ \
-    && ctest --output-on-failure
+    && strip bin_install/taraxad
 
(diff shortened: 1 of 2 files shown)


sfrique:
In the first build, I forced a segfault with `kill -SIGQUIT  $ctest_pid` to verify if the `coredump` would be generated.
It took a short while to appear at the blue ocean UI, but it was generated.

Please check if the separation and the tests seem ok.

justinsnapp (approved):
LGTM

02p01r (approved):

PR #682, opened 2021-04-07
Added debug to conan install

nikonok:
## Purpose

To cache libraries install step on Docker we need to add debug flag to it.
Also some fixes for clang tools location.

### DIFF
--- a/conanfile.py
+++ b/conanfile.py
@@ -19,7 +19,7 @@ class TaraxaConan(ConanFile):
     generators = "cmake"
 
     def _add_clang_utils_on_darwin(self):
-        current_path = os.getcwd()
+        current_path = self.recipe_folder # dir with conanfile.py
         clang_format = "clang-format"
         clang_tidy = "clang-tidy"
         path_to_format = current_path + "/" + clang_format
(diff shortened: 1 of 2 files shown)


conanfile.py
```diff
@@ -20,11 +20,10 @@ class TaraxaConan(ConanFile):
 
     def _add_clang_utils_on_darwin(self):
         current_path = os.getcwd()
-        source_path = os.path.dirname(current_path)
         clang_format = "clang-format"
         clang_tidy = "clang-tidy"
-        path_to_format = source_path + "/" + clang_format
-        path_to_tidy = source_path + "/" + clang_tidy
+        path_to_format = current_path + "/" + clang_format
+        path_to_tidy = current_path + "/" + clang_tidy
```
gaoqi8649, line 26: I changed this to source directory instead of build directory. Because sometimes need to clean up the build directory for re-building. In that case, don't need to download clang-format and clang-tidy again

nikonok: But that path is to conanfile, not to build dir...
Okay, I will double check that

nikonok: I think now it fixed=)

gaoqi8649 (approved):

justinsnapp (approved):

PR #693, opened 2021-04-15
Cirleci

agrebin:
## Purpose

This adds support for circleci builds triggered on **cirleci** and **develop** branches AND tagged with format **vX.Y.Z**
Also contains a builder/Dockerfile which is a strip of the original Dockerfile but just for the builder on which the linux binary depends to compile.
The Release will produce a taraxad-macos and a taraxad-amd64 (for linux) binaries on github repo.


builder/Dockerfile
```diff
@@ -0,0 +1,66 @@
+# Default output dir containing build artifacts
```
sfrique, line 1: Is this the same as the `builder` step on the main `Dockerfile`?

If so, we should use the main `Dockerfile` and build using `--target` option. That way we don't need to keep track of separated files and versions

Also, did you build this manually and upload it right? We may need to automate this since once someone changes the base versions, the builder for that release will need to be updated as well.

.circleci/config.yml
```diff
@@ -0,0 +1,174 @@
...
+       xcode: 12.0.0
+     environment:
+        CLANG_VERSION: "1:10.0-50~exp1"
+        GO_VERSION: "1.13.7"
+        CMAKE_VERSION: "3.16.3-1ubuntu1"
+        GCC_VERSION: "4:9.3.0-1ubuntu2"
+        GFLAGS_VERSION: "2.2.2-1build1"
+        JSONCPP_VERSION: "1.7.4-3.1ubuntu2"
+        JSONRPCCPP_VERSION: "0.7.0-1build3"
+        SCRYPT_VERSION: "1.21-3"
```
sfrique, line 22: Do we need this?

.circleci/config.yml
```diff
@@ -0,0 +1,174 @@
...
+        GOROOT: "/usr/local/go"
+        GOPATH: "/root/.go"
```
sfrique, line 75: Do we need this?

.circleci/config.yml
```diff
@@ -0,0 +1,174 @@
...
+        - run:       
+           name: Checkout Submodules
+           command: |
+              git submodule sync
+              git submodule update --init --recursive --jobs 8
+        - run: 
+           command: mkdir $BUILD_OUTPUT_DIR
+        - run: 
+           name: Copy /opt/taraxa/existing files to project
+           command: cp -rp /opt/taraxa/cmake-docker-build-debug/* /root/project/cmake-docker-build-debug/
```
sfrique, line 93: Can you add more info/docs on why we are copying the files from the image to the repository we clone? (I think this is it)

builder/Dockerfile
agrebin: i agree on both, i was focused on making it work, my next step will be automating the docker image builds.

.circleci/config.yml
agrebin: honestly testing if iti is actually needed would require time, perhaps it is not.

.circleci/config.yml
agrebin: it was difficult for me to make the working directory different to /root/project and the dockerfile used the working dir as /opt/taraxa caching all the conan dependencies there, this step is to take advantage of that cache already set on the docker image.

.circleci/config.yml
sfrique: You can just remove and see if everything will work as expected
I assume the vars are already at the docker when executed, and I don't even see where it's used

.circleci/config.yml
```diff
@@ -0,0 +1,174 @@
...
+#       - git/checkout-with-submodules
+#       - setup_remote_docker:
+#           docker_layer_caching: true
+        - run: 
+            name: Cmake
+            command: | 
+              cd $BUILD_OUTPUT_DIR &&
+              cmake -DCMAKE_BUILD_TYPE=Debug \
```
sfrique, line 101: We will probably want this to be `Release` instead of `Debug` and we will need to work with conan as well

.circleci/config.yml
agrebin: i just removed it, it will work there is not an actual dependency.

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+           name: Execute ctest 
+           command: cd cmake-build/tests && ctest --output-on-failure
+        - store_artifacts:
+           path: /Users/distiller/project/cmake-build/bin/taraxad-macos
+        - persist_to_workspace:
+           root: cmake-build
+           paths: [ bin ]
+   build-linux:
+     environment:
+        BUILD_OUTPUT_DIR: "cmake-docker-build-debug"
```
rattrap, line 104: `cmake-docker-build-release`

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+            command: cd $BUILD_OUTPUT_DIR && make clang-format && make clang-format-check
+        - run: 
+           name: make
+           command: cd $BUILD_OUTPUT_DIR && make -j8 all 
+        - run: 
+           name: make install 
+           command: cd $BUILD_OUTPUT_DIR && make install 
+        - run: 
+           name: strip binary
+           command: cd $BUILD_OUTPUT_DIR && strip bin_install/taraxad && cp bin_install/taraxad bin_install/taraxad-x86_64
```
rattrap, line 168: No need to strip if we don't build the debug version and let's change `bin_install/taraxad-x86_64` to `bin_install/taraxad-linux-amd64`.

I know I recommended this name.

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+        - run: 
+           name: Execute ctest 
+           command: cd $BUILD_OUTPUT_DIR/tests && ctest --output-on-failure
+        - store_artifacts:
+           path: /root/project/cmake-docker-build-debug/bin_install/taraxad
```
rattrap, line 173: You could use the `$BUILD_OUTPUT_DIR` env var here.

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+           name: Cmake
+           command: |
+               cd cmake-build
+               export GOPATH=$HOME/golang &&
+               export GOROOT=/usr/local/opt/go/libexec &&
+               export CC=gcc &&
+               export CPATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include &&
+               cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=gcc -DSYSTEM_HOME_OVERRIDE=/Users/root/project/cmake-build  ../ &&
+               make -j 6  && 
+               cp bin/taraxad bin/taraxad-macos
```
rattrap, line 93: Let's name this `bin/taraxad-macos-intel`

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+            branches: 
+              ignore: /.*/
+            tags:  
+              only: /^v\d+\.\d+\.\d+$/
+          context: taraxa-node
+      - github-release/create:
+          context: taraxa-node
+          attach-workspace: true
+          workspace-root: cmake-docker-build-debug
+          path: cmake-docker-build-debug/bin_install/taraxad-x86_64
```
rattrap, line 213: `cmake-docker-build-debug/bin_install/taraxad-linux-amd64`

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+            tags:  
+              only: /^v\d+\.\d+\.\d+$/
+          requires: 
+            - build-linux
+            - build-mac
+      - github-release/create:
+          context: taraxa-node
+          attach-workspace: true
+          workspace-root: cmake-build
+          path: cmake-build/bin/taraxad-macos
```
rattrap, line 226: `cmake-build/bin/taraxad-macos-intel`

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+          filters:
+            branches: 
+              ignore: /.*/
+            tags:  
+              only: /^v\d+\.\d+\.\d+$/
+          requires: 
+            - build-linux
```
rattrap, line 233: I think this only requires `build-mac`

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+          workspace-root: cmake-docker-build-debug
+          path: cmake-docker-build-debug/bin_install/taraxad-x86_64
+          filters:
+            branches: 
+              ignore: /.*/
+            tags:  
+              only: /^v\d+\.\d+\.\d+$/
+          requires: 
+            - build-linux
+            - build-mac
```
rattrap, line 221: I think this only requires `build-linux`

rattrap (changes requested):

.circleci/config.yml
sfrique: Did you remove it?

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+           command: |
+              git submodule sync
+              git submodule update --init --recursive --jobs 8
+        - run: 
+           command: brew update
+        - run: 
+           command: brew install coreutils 
+        - run:
+           command: |
+              brew install go  ccache gflags &&
```
sfrique, line 61: Extra space

sfrique, line 61: Why do we need `&&` ?

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+              brew install libscrypt libtool cmake libjson-rpc-cpp conan
+        - restore_cache:
+           keys:
+            - conan-cache-v2-mac--{{ checksum "conanfile.py" }}
```
sfrique, line 65: Is there a reason to have the double `--` instead of just one?
if not, I think single would be better. Also update all other entries.
if yes, just resolve this

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+        - run: 
+           name: Cmake
+           command: |
+               cd cmake-build
+               export GOPATH=$HOME/golang &&
+               export GOROOT=/usr/local/opt/go/libexec &&
+               export CC=gcc &&
+               export CPATH=/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.15.sdk/usr/include &&
+               cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_C_COMPILER=gcc -DSYSTEM_HOME_OVERRIDE=/Users/root/project/cmake-build  ../ &&
+               make -j 6  &&
```
sfrique, line 92: Do we need `&&` at end of each line?
I may be missing something.

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+            command: | 
+              cd $BUILD_OUTPUT_DIR &&
+              cmake -DCMAKE_BUILD_TYPE=Release \
+              -DTARAXA_STATIC_BUILD=ON \
+              -DTARAXAD_INSTALL_DIR=./bin_install \
+              -DTARAXAD_CONF_INSTALL_DIR=./bin_install \
+              ../ 
+        - run:
+            name: make clang-format
+            command: cd $BUILD_OUTPUT_DIR && make clang-format && make clang-format-check
```
sfrique, line 159: Wouldn't it be better to add the initial test/check to be the format check? 
So we can just tell the user he should commit the code with the format already in place instead of the ci "fixing" it?

.circleci/config.yml
sfrique: I think the goal is not to release anything if any build fail

.circleci/config.yml
```diff
@@ -0,0 +1,234 @@
...
+           keys:
+            - conan-cache-v1-linux--{{ checksum "conanfile.py" }}
+            - conan-cache-v1-linux
+        - run:
+           name: Conan
+           command: conan install -if $BUILD_OUTPUT_DIR --build missing -s build_type=Release .
+        - save_cache:
+           paths:
+             - ~/.conan
+           key: conan-cache-v1-linux--{{ checksum "conanfile.py" }}
```
sfrique, line 140: I think somehow conan should be already built at the `builder` target image.

PR #712, opened 2021-04-29
Cmd

gaoqi8649:
Implement embed config management command in Taraxa node


src/taraxad/main.cpp
```diff
@@ -10,60 +14,533 @@ using namespace std;
...
+    cerr << "Cannot open file " << vrf_file << endl;
+    exit(1);
+  }
+}
+
+Json::Value getNetworkBootNodes(int network_identifier) {
+  Json::Value network_boot_nodes(Json::arrayValue);
+
+  switch (network_identifier) {
+    case 1: {
```
JakubFornadel, line 107: what is case 1 ? Use enum class for this kind of code so it is obvious

src/taraxad/main.cpp
```diff
@@ -10,60 +14,533 @@ using namespace std;
...
+  // Account
+  auto account_file = getAccountFile();
+  if (!fileExist(account_file)) {
+    auto account = dev::KeyPair::create();
+    auto account_json = getAccountJson(account);
+
+    // Create account file
+    auto account_file = getAccountFile();
+    if (!writeFile(account_file, account_json)) {
+      exit(1);
```
JakubFornadel, line 274: first of all, EXIT_FAILURE is used instead of 1. And why would you use exit() in some internal function ? Use exception and log error please

JakubFornadel (changes requested):
None if tjis functionality should be in main, especially not functions like "getTaraxaDefaultDir()", etc... Please think about some structure/design... main.cpp should be as small as possible with only a few lines of code. Move all this somewhere else + clean up also main funcion - definitely not a good idea to add *150 lines of code to main, add something like handleCommand or something

src/taraxad/main.cpp
```diff
@@ -10,60 +14,533 @@ using namespace std;
...
+    case 2: {
+      Json::Value boot_node1(Json::objectValue);
+      boot_node1["id"] =
+          "1105725ff38294f42197a3851c87631610956404678a25f90da5f989bb590413a62665cadde1b820de3e9098367909110bba6d62d248"
+          "444ce45547a703bd8d10";
+      boot_node1["ip"] = "boot-node-0.testnet.taraxa.io";
+      boot_node1["tcp_port"] = 10002;
+      boot_node1["udp_port"] = 10002;
+      network_boot_nodes.append(boot_node1);
+      Json::Value boot_node2(Json::objectValue);
```
JakubFornadel, line 120: how about make this loop with configurable num of nodes ? it will be smaller and easier to change in case we want to try different set of nodes

src/taraxad/main.cpp
```diff
@@ -10,60 +14,533 @@ using namespace std;
...
+        cout << "Please save the secret key at a safe place. Account secret key: " << toHex(account.secret().ref())
+             << endl;
+        cout << "Account public key: " << account.pub() << endl;
+        cout << "Account address: " << account.address() << endl;
+
+        return 1;
+      }
+
+      if (cmd == "account-from-key") {
+        auto account_file = getAccountFile();
```
JakubFornadel, line 437: do we support multicommand ? if not why do you check all command options even if the previous one was specified. If you used some function like <enum Command> getCommand(std::string) you could handle there also invalid commands and use here switch...

src/taraxad/main.cpp
```diff
@@ -10,60 +14,533 @@ using namespace std;
...
+          return 1;
+        }
+
+        cout << "Generate account file at " << account_file << endl;
+        cout << "Please save the secret key at a safe place. Account secret key: " << toHex(account.secret().ref())
+             << endl;
+        cout << "Account public key: " << account.pub() << endl;
+        cout << "Account address: " << account.address() << endl;
+
+        return 1;
```
JakubFornadel, line 433: return 1 ?

PR #721, opened 2021-05-04
fix(init): wrong signal for sigkill

rattrap:
## Purpose

The docker container doesn't stop when sending sigkill via CRTL+C


justinsnapp (approved):
Yes this typo has been long making us SIGnificantly ILL.  Glad we KILLed this bug.

PR #726, opened 2021-05-06
Remove warnings

MatusKysel:
## Purpose

Removing compiler warnings and treat them as errors

### DIFF
--- a/src/aleth/filter_api.cpp
+++ b/src/aleth/filter_api.cpp
@@ -9,23 +9,23 @@ using namespace eth;
 using namespace util;
 
 struct FilterAPIImpl : virtual FilterAPI {
-  optional<LogFilter> getLogFilter(FilterID id) const override { return std::nullopt; }
+  optional<LogFilter> getLogFilter(FilterID /*id*/) const override { return std::nullopt; }
 
   FilterID newBlockFilter() override { return 0; }
 
   FilterID newPendingTransactionFilter() override { return 0; }
 
-  FilterID newLogFilter(LogFilter const& _filter) override { return 0; }
+  FilterID newLogFilter(LogFilter const& /*_filter*/) override { return 0; }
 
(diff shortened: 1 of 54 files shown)


src/aleth/filter_api.cpp
```diff
@@ -9,23 +9,23 @@ using namespace eth;
 using namespace util;
 
 struct FilterAPIImpl : virtual FilterAPI {
-  optional<LogFilter> getLogFilter(FilterID id) const override { return std::nullopt; }
+  optional<LogFilter> getLogFilter(FilterID /*id*/) const override { return std::nullopt; }
```
JakubFornadel, line 12: have you considered using __attribute__((unused)) ? Not sure what is better practice for this, just thinking out loud...

src/network/taraxa_capability.cpp
```diff
@@ -298,7 +298,7 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
       LOG(log_dg_dag_prp_) << "Received NewBlockPacket " << transactionsCount;
 
       std::vector<Transaction> newTransactions;
-      for (auto iTransaction = 1; iTransaction < transactionsCount + 1; iTransaction++) {
+      for (unsigned long iTransaction = 1; iTransaction < transactionsCount + 1; iTransaction++) {
```
JakubFornadel, line 301: why not size_t ? + refactored auto transactionsCount from a few lines above ?

MatusKysel: good point, I will fix it

src/aleth/filter_api.cpp
MatusKysel: attribute((unused)) works only for gcc/clang, but does not for MSVC.

JakubFornadel (approved):

02p01r (approved):

PR #733, opened 2021-05-07
Enable hack to copy coredumps on gke

sfrique:
## Purpose

Hack to get coredumps when running on GKE to persistent storage.

Right now I am letting it enabled as default, but I am ok with changing it and only enable in some deploys.

(not tested on actual gke)


rattrap (approved):
LGTM

JakubFornadel (approved):

PR #752, opened 2021-05-14
Fix core dump when creating snapshot with rebuild_db flag

MatusKysel:
## Purpose

Fix coredump on rebuild_db flag

### DIFF
--- a/src/storage/db_storage.cpp
+++ b/src/storage/db_storage.cpp
@@ -104,7 +104,8 @@ void DbStorage::loadSnapshots() {
 
 bool DbStorage::createSnapshot(uint64_t const& period) {
   // Only creates snapshot each db_snapshot_each_n_pbft_block_ periods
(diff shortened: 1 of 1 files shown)


JakubFornadel:
Note: according to the devnet node coredumps from the morning we had a bug when we tried to create db snapshot with some period and such snapshot already existed.... the flow is execute pbft block -> if condition met, create db snapshot
https://github.com/Taraxa-project/taraxa-node/blob/develop/src/node/executor.cpp#L189

Not really sure if this is the fix we want.... Technically yes this will prevent exception to be thrown and it should be handled... but the real question is how could this even happen ? Maybe there is more serious problem in the flow -> how could the node try to create db snapshot on some period if such snapshot already existed ? That means we already processed such pbft block, saved it into db and therefore on node bootup this block should have been loaded from db and never be processed again... 

It kind of smells that something weird happened here... So yes we should handle it as proposed in this PR, but we might also just hide the real problem by this fix...  @justinsnapp

JakubFornadel (approved):

PR #768, opened 2021-05-17
fix(tests): fix test configs, so test are not interfered by devnet or…

MatusKysel:
## Purpose
I have noticed that when I run a node connected to test/dev-net and then later I run tests. Tests receive packets that either crash tests or just cause error in tests. So, it looks like node discovery or something like this is causing it

### DIFF
--- a/tests/util_test/conf/conf_taraxa1.json
+++ b/tests/util_test/conf/conf_taraxa1.json
@@ -4,8 +4,8 @@
   "db_path": "/tmp/taraxa",
   "network_is_boot_node": true,
   "network_address": "0.0.0.0",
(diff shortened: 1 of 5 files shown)


justinsnapp (approved):
Simple enough.  Can't cause problems.

gaoqi8649 (commented):
Not sure if need to change IP/ID also?

MatusKysel:
> Not sure if need to change IP/ID also?

IP for tests is always localhost and far as I know ID is irrelevant to p2p, but when we change port it is sufficient to not be discovered by live nodes

JakubFornadel (approved):

PR #772, opened 2021-05-18
fix(vote): remove votes more than a round ahead from the unverified queue

### LINKED ISSUE #742: Votes from the "distant future" are not discarded by the node
### Description

If a node "maliciously" places PBFT votes in rounds way ahead of actual network PBFT round those votes will be forever kept in unverified queue. 

### Expected behaviour

A node that is confident it has been in sync should be able to place a bound on what are reasonable votes and discard unreasonably futuristic votes.

### Actual behaviour

These malicious votes remain in unverified queue and are validated over and over without success.

justinsnapp:
## Purpose

Remove votes more than a round head from the unverified queue, if they fail validation, to prevent excessive unvalidated vote buildup.

## Related Github issues

Fixes #742 

## Changes

In getVerifiedVotes we try to validate unverified votes, and if they fail we remove them now from unverified queue if they are more than one round ahead.

If we are boostrapping to the correct future round, and have synced to the latest full chain then votes will pass validation.   
If they are only a round ahead they may pass validation as soon as we catch up with our asynchronous execution.

### DIFF
--- a/src/consensus/vote.cpp
+++ b/src/consensus/vote.cpp
@@ -287,6 +287,7 @@ std::vector<Vote> VoteManager::getVerifiedVotes(uint64_t const pbft_round, size_
   // Cleanup votes for previous rounds
   cleanupVotes(pbft_round);
 
+  std::vector<Vote> future_unverifiable_votes;
   std::vector<Vote> verified_votes;
   auto votes_to_verify = getUnverifiedVotes();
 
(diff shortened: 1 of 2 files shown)


gaoqi8649 (commented):
One thing I can think is if remove votes in the future rounds, the node may not be able to receive the votes again since we have peer->markVoteAsKnown(vote_hash)
https://github.com/Taraxa-project/taraxa-node/blob/develop/src/network/taraxa_capability.cpp#L439

PR #776, opened 2021-05-19
Circleci

agrebin:
## Purpose

Adding support for build and push test and official Docker images

## For reviewers

The main changes are in just one job **build-and-push-docker-image** tested many ways, using docker executor found that the fastest method was to use machine executor and take advantage of docker_layer_caching, also using just one job speeds up the process since no other VMs needs to be spinned up, so everything on one job, multiple steps, which execution is filtered by comparing to ${CIRCLE_BRANCH}, probably code can be cleaned up and improved by creating an orb and commands.
Some testing still needs to be done, regarding the branch filtering for branches PR-* i did no tested that part., rest should be working i tested the code as thoroughly as i could without impacting existing process

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -10,6 +10,7 @@ jobs:
    build-docker-image:
     environment:
       IMAGE=taraxa-node
+      GCP_IMAGE=gcr.io/$GOOGLE_PROJECT_ID/${IMAGE}
     machine:
       image: ubuntu-1604:202007-01
       docker_layer_caching: true
@@ -36,16 +37,18 @@ jobs:
 
            mkdir -p  $PWD/tmp_docker 
       - gcp-gcr/build-image:
-         image: $IMAGE-$DOCKER_BRANCH_TAG-$CIRCLE_BUILD_NUM
+         image: $IMAGE
          tag: $DOCKER_BRANCH_TAG-$CIRCLE_BUILD_NUM
       - gcp-gcr/build-image:
-         image: $IMAGE-$DOCKER_BRANCH_TAG-$CIRCLE_BUILD_NUM-ctest
+         image: $IMAGE-ctest
          tag: $DOCKER_BRANCH_TAG-$CIRCLE_BUILD_NUM
          extra_build_args: --target build
       - run:
          name: Run Ctest
          command: |
-           docker run -d --rm -v $PWD/tmp_docker:/tmp --name taraxa-node-ctest-$DOCKER_BRANCH_TAG gcr.io/$GOOGLE_PROJECT_ID/${IMAGE}-${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}-ctest sh -c \
+           docker run -d --rm -v $PWD/tmp_docker:/tmp --name taraxa-node-ctest-$DOCKER_BRANCH_TAG \
+                       ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}-ctest\
+                        sh -c \
                        'cd cmake-docker-build-debug/tests \
                         && ctest --output-on-failure' 
       - run:
@@ -67,7 +70,7 @@ jobs:
           docker run --rm -d \
               --name taraxa-node-smoke-test-${DOCKER_BRANCH_TAG} \
               --net smoke-test-net-${DOCKER_BRANCH_TAG} \
-              gcr.io/$GOOGLE_PROJECT_ID/${IMAGE}-${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} \
+              ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} \
               single
           mkdir -p  $PWD/test_build-d/
           sleep 30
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -7,8 +7,172 @@ orbs:
 
 
 jobs: 
+   build-and-push-docker-image:
+    environment:
+      - IMAGE: taraxa-node
+    machine:
+      image: ubuntu-1604:202007-01
```
sfrique, line 14: We should be using a newer image
https://circleci.com/docs/2.0/configuration-reference/#available-machine-images

.circleci/config.yml
```diff
@@ -7,8 +7,172 @@ orbs:
...
+           echo "export START_TIME=$(date +%Y%m%d-%Hh%Mm%Ss)" >>$BASH_ENV
+           echo "export GCP_IMAGE=gcr.io/${GOOGLE_PROJECT_ID}/${IMAGE}" >> $BASH_ENV
+           sudo service apport stop
+           sudo bash -c 'echo /tmp/core.%t.%e.%p >/proc/sys/kernel/core_pattern'
+
+           mkdir -p  $PWD/tmp_docker 
+      - run:
+         name: Docker Build taraxa-node
+         command: |
+           docker build -t ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} .
```
sfrique, line 42: I don't recall exactly why, but we use to want `DOCKER_BUILDKIT=1` 
But we may don't need it anymore, would be nice to verify it.

.circleci/config.yml
```diff
@@ -7,8 +7,172 @@ orbs:
...
+              export CIRCLE_BUILD_NUM=300
+           fi
+           echo "export HELM_TEST_NAME=$(echo $CIRCLE_BRANCH | sed 's/[^A-Za-z0-9\\-]*//g' | tr '[:upper:]' '[:lower:]')" >>$BASH_ENV
+           echo "export DOCKER_BRANCH_TAG=$(./scripts/docker_tag_from_branch.sh $CIRCLE_BRANCH)" >>$BASH_ENV
+           echo "export START_TIME=$(date +%Y%m%d-%Hh%Mm%Ss)" >>$BASH_ENV
+           echo "export GCP_IMAGE=gcr.io/${GOOGLE_PROJECT_ID}/${IMAGE}" >> $BASH_ENV
+           sudo service apport stop
+           sudo bash -c 'echo /tmp/core.%t.%e.%p >/proc/sys/kernel/core_pattern'
+
+           mkdir -p  $PWD/tmp_docker
```
sfrique, line 38: I think this only works because we are using executer type machine
if we don't plan to move to execute docker, then this hack should be ok, but I would rather try to use docker.

.circleci/config.yml
```diff
@@ -7,8 +7,172 @@ orbs:
 
 
 jobs: 
+   build-and-push-docker-image:
+    environment:
+      - IMAGE: taraxa-node
+    machine:
```
sfrique, line 13: So we will use machine type instead of docker
is there a reason why?

.circleci/config.yml
```diff
@@ -7,8 +7,172 @@ orbs:
...
+                    }')
+          cat $PWD/test_build-d/http.out || true
+          if [[ $http_code -eq 200 ]] ; then
+              exit 0
+          else
+              exit $http_code
+          fi
+          docker kill taraxa-node-smoke-test-${DOCKER_BRANCH_TAG} || true
+          docker network rm smoke-test-net-${DOCKER_BRANCH_TAG} || true
+          docker images -a
```
sfrique, line 106: I think this will never run, as the `if/else` above `exit` on any case.

.circleci/config.yml
```diff
@@ -7,8 +7,172 @@ orbs:
...
+           echo ${GCLOUD_SERVICE_KEY} | docker login -u _json_key --password-stdin https://gcr.io
+           docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${GCP_IMAGE}:${CIRCLE_SHA1}
+           docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${GCP_IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}
+           docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${GCP_IMAGE}:${START_TIME}
+           docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${GCP_IMAGE}:${START_TIME}-${CIRCLE_SHA1}
+           docker push ${GCP_IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}
+           docker push ${GCP_IMAGE}:${START_TIME}
+           docker push ${GCP_IMAGE}:${CIRCLE_SHA1}
+           docker push ${GCP_IMAGE}:${START_TIME}-${CIRCLE_SHA1}
+           if [[ ! ${CIRCLE_BRANCH} =~ ^PR-* ]];then
```
sfrique, line 119: As discussed, I think circle ci don't show PRs as branches like this
So we need to fix this

.circleci/config.yml
```diff
@@ -7,8 +7,172 @@ orbs:
...
+              docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${GCP_IMAGE}:latest
+              #docker push ${GCP_IMAGE}:latest
+           fi
+           if [[ ${CIRCLE_BRANCH} == "master" ]];then
+             echo ${DOCKERHUB_PASS} | docker login -u taraxa --password-stdin
+             docker push ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}
+             docker push ${IMAGE}:${START_TIME}
+             docker push ${IMAGE}:${CIRCLE_SHA1}
+             docker push ${IMAGE}:${START_TIME}-${CIRCLE_SHA1}
+             if [[ ! ${CIRCLE_BRANCH} =~ ^PR-* ]];then
```
sfrique, line 129: As discussed, I think circle ci don't show PRs as branches like this
So we need to fix this

.circleci/config.yml
```diff
@@ -7,8 +7,172 @@ orbs:
...
+               docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${IMAGE}:latest
+               #docker push ${IMAGE}:latest
+             fi
+           fi
+      - run:
+         name: Install/configure helm and chart
+         command: |
+            if [[ ${CIRCLE_BRANCH} =~ ^PR-* ]];then
```
sfrique, line 137: As discussed, I think circle ci don't show PRs as branches like this
So we need to fix this

.circleci/config.yml
```diff
@@ -7,8 +7,172 @@ orbs:
...
+                --set replicaCount=5 \
+                --set test.pythontester.script=jenkins.py \
+                --set image.repository=${GCP_IMAGE} \
+                --set image.tag=${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} \
+                -f taraxa-node/values.yaml
+            fi
+      - run:
+         name: Run helm test
+         command: |
+            if [[ ${CIRCLE_BRANCH} =~ ^PR-* ]];then
```
sfrique, line 156: As discussed, I think circle ci don't show PRs as branches like this
So we need to fix this

.circleci/config.yml
```diff
@@ -7,8 +7,172 @@ orbs:
...
+              helm test ${HELM_TEST_NAME} \
+                --timeout 3600 \
+                --cleanup
+              ./scripts/kibana-url.sh || true
+            fi
+      - run:
+         name: Cleanup helm tests
+         command: |
+           if [[ ${CIRCLE_BRANCH} =~ ^PR-* ]];then
```
sfrique, line 165: As discussed, I think circle ci don't show PRs as branches like this
So we need to fix this

.circleci/config.yml
```diff
@@ -176,21 +340,28 @@ jobs:
...
+    jobs:
+      - build-and-push-docker-image:
+         context: taraxa-node
+         filters:
+          branches: 
+            only: 
+              - circleci
+              - develop
+              - master
+              - /^PR-*/
```
sfrique, line 353: Verify this PR part

sfrique (changes requested):

Issue #777, opened 2021-05-19
Ability to run a local network of nodes


rattrap:
## Is your feature request related to a problem? Please describe.

It's pretty hard right now to set up a local network of boot nodes and RPC nodes in order to test changes you make locally.
You need to create config files for each node, change the chain ID and genesis hash, find the boot node public addresses, add them in all config files in the DPOS section and so on.

## Describe the solution you'd like.

It would be nice to have a python script that you can run locally and it can set everything up and start the nodes. Similar to docker-compose but it should work with taraxad binaries.

You can, for example, call it like this:

```bash
./cli/local-net --boot-nodes 3 --rpc-nodes 3 <binary>
```

And it should configure and start 6 nodes locally and print the logs.

PR #778, opened 2021-05-19
feat(scripts): script that helps run a local network of nodes

rattrap:
## Purpose

Ability to run a local network of nodes

## For reviewers

Created a script that helps bootstrap a local network easily.

./cli/local-net [binary] # starts 3 boot nodes and 3 RPC nodes with the specified binary
./cli/local-net --boot-nodes 3 --rpc-nodes 1 [binary] # starts 3 boot nodes and 1 RPC node with the specified binary

## Related Github issues

https://github.com/Taraxa-project/taraxa-node/issues/777


cli/local-net
```diff
@@ -0,0 +1,198 @@
...
+    "purple": 95,
+    "cyan": 96,
+    "lightGray": 97,
+}
+worker_colors = {}
+
+
+def config(network, boot_nodes, rpc_nodes):
+    if not os.path.isdir('./local-net'):
+        os.mkdir('./local-net')
```
sfrique, line 26: This kind of conflicts with line 28
The script name is `local-net` and also the folder for the local data is called `local-net`   so if you run inside the `cli` folder, it will fail. 
Your "docs" in this pr tells to run `./local-net binary` but at the #777 it shows `cli/local-net`  
In any case, please add some doc to README explaining how to run it so the paths don't conflict.

cli/local-net
```diff
@@ -0,0 +1,198 @@
...
+    for i in range(boot_nodes):
+        worker_name = f'boot-node-{i}'
+        public_key = subprocess.check_output(
+            f'{taraxa_cli} account | grep public', shell=True).strip().decode('utf-8').split(' ')[2]
+        private_key = subprocess.check_output(
+            f'{taraxa_cli} account | grep private', shell=True).strip().decode('utf-8').split(' ')[2]
+        public_address = subprocess.check_output(
+            f'{taraxa_cli} account --secret {private_key} | grep address:', shell=True).strip().decode('utf-8').split(' ')[1]
+        private_keys[worker_name] = private_key[2:]
+        public_addresses[worker_name] = public_address
```
sfrique, line 43: The public/private keys and addresses have no relation to each other? 
Cause you are running the command multiple times.. I would assume they have some kind of relation.

cli/local-net
```diff
@@ -0,0 +1,198 @@
...
+    for i in range(rpc_nodes):
+        worker_name = f'rpc-node-{i}'
+        config_name = f'./local-net/conf-{worker_name}.json'
+
+        if not os.path.isdir(f'{local_net}/db-{worker_name}'):
+            os.mkdir(f'{local_net}/db-{worker_name}')
+
+        cmd = ''
+        cmd += f'{taraxa_cli} config node --network {network} '
+        cmd += f'{boot_nodes_flags} '
```
sfrique, line 109: I think this won't work since the nodes need to connect to the boot nodes, and the address here are for listening.
You may change everything to `127.0.0.1` on line 45 that it may work.

But I am not sure.

sfrique: A side note, when i tried to run with `0.0.0.0` it kept showing this:
` 2021-05-19 18:36:28.945035 SUMMARY SILENT STATUS: NOT CONNECTED TO ANY PEERS. POSSIBLE CONFIG ISSUE OR NETWORK CONNECTIVITY`
when I changed to `127.0.0.1`  that msg did not show up.

sfrique (changes requested):
Please add some docs maybe at README? 

Another thing, I am not sure if you want this script to be published inside the docker image, if so, please check if it's been added.

JakubFornadel:
just a minor note to commits names - dont need to change it if you dont want to, just for clarification :) This is not a feat but rather chore as it is just our internal tool for testing. Think about it from external user point of view - if we add rewards distribution, that is a feature for anybody who runs our node. They dont care about our internal testing stuff like jenkins, local testnets, etc...

cli/local-net
rattrap: Yes. You are right. I will add some docs but you can run this anywhere but in the cli directory. I don't know if I want to change the script name just for that.

cli/local-net
rattrap: Right. `public_key` should have also been derived from `private_key`. Fixed

cli/local-net
rattrap: Changed to 127.0.0.1 and it works. Fixed

cli/local-net
sfrique: Just having some docs should be fine.
Or change the folder to `local-net_data` or something would also "fix".

rattrap: Added docs and changed directory to `local-net-data`

PR #782, opened 2021-05-20
perf(vote): speed up vote processing

MatusKysel:
## Purpose

Small changes to significantly improve processing of votes

### DIFF
--- a/src/consensus/vote.cpp
+++ b/src/consensus/vote.cpp
@@ -81,13 +81,6 @@ bytes Vote::rlp(bool inc_sig) const {
   return s.out();
 }
 
-void Vote::process() const {
-  if (cached_voter_) return;
-  cached_voter_ = dev::recover(vote_signature_, sha3(false));
-  assert(cached_voter_);
-  cached_voter_addr_ = dev::toAddress(cached_voter_);
-}
-
 VoteManager::VoteManager(addr_t node_addr, std::shared_ptr<DbStorage> db, std::shared_ptr<FinalChain> final_chain,
                          std::shared_ptr<PbftChain> pbft_chain)
     : db_(db), pbft_chain_(pbft_chain), final_chain_(final_chain) {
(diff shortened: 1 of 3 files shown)


src/consensus/vote.cpp
```diff
@@ -85,6 +85,7 @@ void Vote::voter() const {
   if (cached_voter_) return;
   cached_voter_ = dev::recover(vote_signature_, sha3(false));
```
JakubFornadel, line 86: Let's just clean this up little bit. Why to use voter() function at all ? It would be better to do this caching in getVoter()

src/consensus/vote.cpp
```diff
@@ -131,6 +132,7 @@ VoteManager::VoteManager(addr_t node_addr, std::shared_ptr<DbStorage> db, std::s
 void VoteManager::addUnverifiedVote(taraxa::Vote const& vote) {
   uint64_t pbft_round = vote.getRound();
   auto hash = vote.getHash();
+  vote.process();  // speed up we do not need to calcualte again
```
JakubFornadel, line 135: not needed if we cache stuff as described in above comment

src/consensus/vote.hpp
```diff
@@ -103,6 +103,12 @@ class Vote {
     voter();
     return cached_voter_;
   }
+  addr_t getVoterAddr() const {
+    voter();
```
JakubFornadel, line 107: use getVoter() here instear of voter

src/consensus/vote.cpp
MatusKysel: it is need because later we work only with copies, so it has to be there

src/consensus/vote.cpp
```diff
@@ -85,6 +85,7 @@ void Vote::voter() const {
   if (cached_voter_) return;
   cached_voter_ = dev::recover(vote_signature_, sha3(false));
   assert(cached_voter_);
+  cached_voter_addr_ = dev::toAddress(cached_voter_);
```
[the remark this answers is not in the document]
MatusKysel: nope because we work we the copy of this

src/consensus/vote.cpp
JakubFornadel: so make that copy later when you need to ?

src/consensus/vote.cpp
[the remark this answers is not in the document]
JakubFornadel: I guess I dont get something here... copy or not what is the difference you can make that copy anytime ?

src/consensus/vote.hpp
MatusKysel: fixed

src/consensus/vote.cpp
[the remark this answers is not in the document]
MatusKysel: fixed

JakubFornadel (approved):

gaoqi8649 (approved):
Looks good to me

Issue #787, opened 2021-05-21
Node attempts to sync with a peer running wrong node version


justinsnapp:
### Description

Node will at times try to sync from node `19313afe` even though it can clearly see in processing initial status packet that its running the wrong node version...

```
2021-05-21 02:22:04.783228 SUMMARY INFO Currently syncing from node ##19313afe…
2021-05-21 02:22:04.783274 SUMMARY INFO Max peer PBFT chain size:      80705 (peer ##19313afe…)
2021-05-21 02:22:04.783323 SUMMARY INFO Max peer PBFT consensus round: 87923 (peer ##19313afe…)
2021-05-21 02:22:04.783363 SUMMARY INFO Max peer DAG level:            294866 (peer ##19313afe…)
2021-05-21 02:22:04.783768 TARCAP INFO Node ##19313afe… disconnected
2021-05-21 02:22:07.328666 TARCAP INFO Node ##19313afe… connected
2021-05-21 02:22:07.447971 TARCAP ERROR Incorrect node version: 0.01, our node major version1.06, host ##19313afe… will be disconnected
2021-05-21 02:22:07.465740 TARCAP INFO Node ##19313afe… disconnected
```

##### Node version

v1.6

### Expected behavior

Status packet would be first packet processed and would disconnect immediately if node doesn't pass initial status packet

### Actual behavior

Either we don't disconnect immediately, or more likely the initial status packet is NOT always the first packet processed for the node.

justinsnapp:
I think we should add a `peer->passed_initial_` boolean flag for the peer and ignore non-initial status packets if the peer hasn't passed initial status packet check.

MatusKysel:
Yesterday I tried to simulate this, because I have also thought this is happening, but actually it never happens. I think it is impossible to send anything before disconnecting

MatusKysel:
Never mind, just happened to me

PR #788, opened 2021-05-21
Bugfix/issue 787/Check initial status packet

gaoqi8649:
## Purpose

For status packets, need to check if receive the initial one

## Related Github issues

https://github.com/Taraxa-project/taraxa-node/issues/787

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -241,6 +241,8 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
           break;
         }
 
+        peer->passed_initial_ = true;
+
         peer->dag_level_ = peer_dag_level;
         peer->pbft_chain_size_ = peer_pbft_chain_size;
         peer->syncing_ = peer_syncing;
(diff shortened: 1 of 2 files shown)


src/network/taraxa_capability.cpp
```diff
@@ -241,6 +246,8 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
           break;
         }
 
+        peer->passed_initial_ = true;
```
JakubFornadel, line 249: Thinking about some edge case when the other node would restart very quickly in non-standard way (e.g. I would kill it with signal), would be such case detected and peer object deleted ? I guess even if not the other node sends initial packet  and it would not pass checks... Anyway just brainstorming if there is some sort of a situation when this flag would be still set to true even if it should not be as we do not set it to false anywhere...

MatusKysel: onDisconnect will erase peer object, so no worries there. Also keep in mind that your node needs to pass same check, so if the other node does not receive this packet from you, it will disconnect too

MatusKysel (approved):

justinsnapp (approved):

src/network/taraxa_capability.cpp
justinsnapp: I don't think such an edge case is really detectable.  A "split personality" node if you will could always maliciously connect in a valid way and then start sending garbage.   Handling such true maliciousness remains a TODO within taraxa capability, primely we don't stop syncing from the peer with most progress if they refuse to send us anything or they send us useless stuff.   Nor do we blacklist nodes for garbage. ...

PR #794, opened 2021-05-24
build/clang-12: fix build on macOS

kstdl:
## Purpose

Make project build on the macOS with latest apple-clang-12 compiler. 

## For reviewers

`set(CMAKE_OSX_SYSROOT /)` was removed from root CMakeLists.txt because it caused linking problems.


justinsnapp:
Works on my mac.

conanfile.py
```diff
@@ -45,15 +45,15 @@ def requirements(self):
         self.requires("boost/1.71.0")
         self.requires("cppcheck/2.3")
         self.requires("openssl/1.1.1f")
-        self.requires("cryptopp/8.4.0")
+        self.requires("cryptopp/8.5.0")
```
MatusKysel, line 48: why did you changed version number? Should not we check all of them?

CMakeLists.txt
```diff
@@ -179,13 +188,12 @@ endif()
 include(CMakeModules/clang_tidy.cmake)
 
 # clang-format
-include(CMakeModules/clang_format.cmake)
+# include(CMakeModules/clang_format.cmake)
```
JakubFornadel, line 191: why is clang-format commented out ?

src/consensus/pbft_chain.cpp
```diff
@@ -111,7 +111,7 @@ PbftBlockCert::PbftBlockCert(PbftBlock const& pbft_blk, std::vector<Vote> const&
 
 PbftBlockCert::PbftBlockCert(dev::RLP const& rlp) {
   pbft_blk.reset(new PbftBlock(rlp[0]));
-  for (auto const& el : rlp[1]) {
+  for (auto const el : rlp[1]) {
```
JakubFornadel, line 114: are we sure we are not making copies in these rlp loops when not using & ?

conanfile.py
kstdl: that was need for native arm m1 build without rosseta and i forget to change it back. there are some more unresolved issues with rocksdb, so we cant fix it easily. 
`gmp` version is changed to suppress conan override warning from `mpfr`
```WARN: mpfr/4.0.2: requirement gmp/6.2.1 overridden by taraxa-node/0.1 to gmp/6.2.0```

src/consensus/pbft_chain.cpp
kstdl: it was changed because of warnings that become errors
```error: loop variable 'n' is always a copy because the range of type 'dev::RLP' does not return a reference [-Werror,-Wrange-loop-analysis]```

justinsnapp (approved):

PR #795, opened 2021-05-24
chore(gitmodules): change taraxa submodule paths to relative

kstdl:
## Purpose

Make taraxa submodules paths relative. After that change on submodules initialization it will make a url from that one you used for the main repo saving a scheme. So if you clones a repo with `ssh` it will be also used for all submodules


MatusKysel (approved):

justinsnapp (approved):
LGTM

PR #804, opened 2021-05-25
Circleci

agrebin:
## Purpose

This is to migrate image build from Jenkins into CIRCLECI

## For reviewers

Will trigger Builds for PRs on any branch and specific actions into develop and master.
As mentioned before it is using machine executor since the docker executor was troublesome in reusing the docker cache layers to build quickly the images.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,14 @@
 version: 2.1
+
+parameters: 
+  pull_request:
+    type: string
+    default: ""
+
+  deploy_testnet:
+    default: true
+    type: boolean
+
 orbs:
   git: opuscapita/git@0.0.3
   github-release: izumin5210/github-release@0.1.1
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -7,8 +17,182 @@ orbs:
 
 
 jobs: 
+   build-and-push-docker-image:
+    environment:
+      - IMAGE: taraxa-node
+#      - CIRCLE_PULL_REQUEST: << pipeline.parameters.pull_request >>
+    machine:
+      image: ubuntu-1604:202007-01
```
sfrique, line 25: I would update the image to: `ubuntu-2004:202104-01`
since 1604 is not updated anymore

.circleci/config.yml
```diff
@@ -7,8 +17,182 @@ orbs:
...
+                        }]
+                    }')
+          cat $PWD/test_build-d/http.out || true
+          if [[ $http_code -eq 200 ]] ; then
+              exit 0
+          else
+              exit $http_code
+          fi
+      - run:
+         name: Cleanup smoke tests
```
sfrique, line 117: We should run this step also if the step above fails
I don't know how to do this or even if it's possible.

.circleci/config.yml
```diff
@@ -7,8 +17,182 @@ orbs:
...
+         command: |
+          docker kill taraxa-node-smoke-test-${DOCKER_BRANCH_TAG} || true
+          docker network rm smoke-test-net-${DOCKER_BRANCH_TAG} || true
+          docker images -a
```
sfrique, line 121: Do we need this line?

.circleci/config.yml
```diff
@@ -7,8 +17,182 @@ orbs:
...
+           docker push ${GCP_IMAGE}:latest
+           docker push ${GCP_IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}
+           docker push ${GCP_IMAGE}:${START_TIME}
+           docker push ${GCP_IMAGE}:${CIRCLE_SHA1}
+           docker push ${GCP_IMAGE}:${START_TIME}-${CIRCLE_SHA1}
+           if [[ ${PR} != "false" ]];then
+              docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${GCP_IMAGE}:pr-${PR}-${CIRCLE_BUILD_NUM} 
+              docker push ${GCP_IMAGE}:pr-${PR}-${CIRCLE_BUILD_NUM} 
+           fi
+           if [[ ${CIRCLE_BRANCH} == "master" ]];then
```
sfrique, line 140: We are also building the image `taraxa-node-develop` on GCR when merge to `develop` branch happens.
it's a different docker image at GCR, not only tags

.circleci/config.yml
```diff
@@ -7,8 +17,182 @@ orbs:
...
+              echo ${DOCKERHUB_PASS} | docker login -u taraxa --password-stdin
+              docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${IMAGE}:${CIRCLE_SHA1}
+              docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${IMAGE}:${START_TIME}
+              docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${IMAGE}:${START_TIME}-${CIRCLE_SHA1}
+              docker tag ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${IMAGE}:latest
+              docker push ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}
+              docker push ${IMAGE}:${START_TIME}
+              docker push ${IMAGE}:${CIRCLE_SHA1}
+              docker push ${IMAGE}:${START_TIME}-${CIRCLE_SHA1}
+              docker push ${IMAGE}:latest
```
sfrique, line 150: We should not push the images to docker hub before helm tests succeed.
Actually, we should ask the team if they really want this helm test to be run, currently, there is no useful test there, so right now this is not a problem, but if we do plan to have meaningful tests running over there, so the final images shouldn't be pushed before tests are good.

.circleci/config.yml
```diff
@@ -174,23 +358,25 @@ jobs:
...
+                           
+  create-builder-image:
     jobs:
       - create-and-push-builder:
           context: taraxa-node
           filters:
             branches: 
               only: 
-                #- develop
+#               - develop
```
sfrique, line 377: The `master` and `develop` branches will be added to another PR?
asking because of the PR description

.circleci/config.yml
agrebin: Due to the layer caching the whole workspace remains.

.circleci/config.yml
agrebin: i may have not understood that's a different image, can you please point me out the difference so i can make it happen?

.circleci/config.yml
agrebin: Good catch, will change that behavior did not see that.thanks.

.circleci/config.yml
sfrique: but why do we need to list the images?

.circleci/config.yml
agrebin: That's a different job, it is for building the binaries.

sfrique: ops hahah

.circleci/config.yml
sfrique: We have those images on GCR
`taraxa-node`
`taraxa-node-develop`

It was asked that we had an image with the builds from `develop` branch, and only tags wouldn't fit the request
So `develop` branch builds and push this image. So the team knows if getting the latest from `taraxa-node-develop` means latest build from that branch.
We can change that, but current this is what we are doing at jenkins

.circleci/config.yml
sfrique: `when: always`

.circleci/config.yml
agrebin: that was a debug message.

.circleci/config.yml
agrebin: @sfrique  So to confirm taraxa-node images are only pushed when built on master branch?

sfrique: `taraxa-node` on docker hub yes.
we do push it to GCR for the PR build, but we do not tag it as the latest.
As said, this tagging did not get much thought, so we could change it if it would improve anything.

.github/workflows/issue-link.yml
```diff
@@ -13,3 +13,31 @@ jobs:
           branch-prefix: 'issue-' # required
           resolve: 'true' # optional (default: "false")
           link-style: 'comment' # optional (default: "body")
+
```
rattrap, line 16: This doesn't work if added to the circleci.yaml file?

rattrap (approved):

sfrique (approved):

.circleci/config.yml
agrebin: Thank you , so now it is the current behaviour of the circleci pipeline,

.github/workflows/issue-link.yml
agrebin: As far as i understand it does not interfere with circleci.. if you know something else i should check let me know.

PR #810, opened 2021-05-26
fix(taraxa_capability): fix sending too big BlocksPacket -> if > 16MB…

JakubFornadel:
…, divide it into multiple smaller packets

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -1013,11 +1013,10 @@ void TaraxaCapability::onNewBlockVerified(DagBlock const &block) {
 void TaraxaCapability::sendBlocks(NodeID const &_id, std::vector<std::shared_ptr<DagBlock>> blocks) {
   taraxa::bytes packet_bytes;
   size_t packet_items_count = 0;
-  size_t dag_block_items_count = 0;
   size_t blocks_counter = 0;
 
   for (auto &block : blocks) {
-    dag_block_items_count = 0;
+    size_t dag_block_items_count = 0;
     size_t previous_block_packet_size = packet_bytes.size();
 
     // Add dag block rlp to the sent bytes array
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -103,9 +103,6 @@ void TaraxaCapability::sealAndSend(NodeID const &nodeID, unsigned packet_type, R
                  << ", receiver: " << nodeID.abridged();
     return;
   }
-  if (packet_type == BlocksPacket) {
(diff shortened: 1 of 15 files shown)


src/network/taraxa_capability.cpp
```diff
@@ -1416,58 +1416,129 @@ void TaraxaCapability::sendPbftBlocks(NodeID const &_id, size_t height_to_sync,
...
+    // MAX_PACKET_SIZE (15 MB) limit
+    if (act_packet_size + pbft_block_rlp.out().size() > MAX_PACKET_SIZE) {
+      if (pbft_blocks_rlps.empty()) {
+        // If this log is present in log, it means there was created such big pbft block, that we are not able to send
+        // it through network
+        // This situation should never ever happen !!!
+        LOG(log_er_) << "Unable to send pbft block: " << pbft_blocks[pbft_block_idx].pbft_blk->getBlockHash().abridged()
+                     << " due to MAX_PACKET_SIZE(" << MAX_PACKET_SIZE << ") limit. "
+                     << "Block size: " << pbft_block_rlp.out().size() << " [B]";
+        // Skip this block as it would go through network anyway
```
justinsnapp, line 1524: Think you mean to comment that it would NOT go through network anyway.   Perhaps we should kill the node if this occurs not just log an error?

JakubFornadel: That block is already in db so killing/restarting node would solve nothing... But yeah it should not be even possible to create such pbft block so if it happens, we have a huge problem... Definitely needs some additional changes...

src/network/taraxa_capability.cpp
```diff
@@ -1368,58 +1412,129 @@ void TaraxaCapability::sendPbftBlocks(NodeID const &_id, size_t height_to_sync,
...
+    // MAX_PACKET_SIZE (15 MB) limit
+    if (act_packet_size + pbft_block_rlp.out().size() > MAX_PACKET_SIZE) {
+      if (pbft_blocks_rlps.empty()) {
+        // If this log is present in log, it means there was created such big pbft block, that we are not able to send
+        // it through network
+        // This situation should never ever happen !!!
+        LOG(log_er_) << "Unable to send pbft block: " << pbft_blocks[pbft_block_idx].pbft_blk->getBlockHash().abridged()
+                     << " due to MAX_PACKET_SIZE(" << MAX_PACKET_SIZE << ") limit. "
+                     << "Block size: " << pbft_block_rlp.out().size() << " [B]";
+        // Skip this block as it would not go through network anyway
```
gaoqi8649, line 1517: I think we should assert here

PR #811, opened 2021-05-27
Test/send blocks

MatusKysel:


tests/network_test.cpp
```diff
@@ -76,6 +76,50 @@ TEST_F(NetworkTest, transfer_block) {
...
+    trx_bytes.emplace_back(*trx.rlp());
+    trx_hashes.push_back(trx.getHash());
+  }
+
+  // creating lot of non valid blocks just for size
+  for (int i = 0; i < 100; ++i) {
+    DagBlock blk(blk_hash_t(1111 + i), 0, {blk_hash_t(222 + i), blk_hash_t(333 + i), blk_hash_t(444 + i)}, trx_hashes,
+                 sig_t(7777 + i), blk_hash_t(888 + i), addr_t(999 + i));
+    dag_blocks.emplace_back(std::make_shared<DagBlock>(blk));
+  }
```
gaoqi8649, line 101: Why need those 100 invalid DAG blocks? Just for occupy size of sendBlocks. I think the unit test for testing sendBlocks could be able to split large packet. But how do we know if the total size is greater than 16MB or not?

tests/util_test/util.hpp
```diff
@@ -103,6 +103,15 @@ inline bool wait(wait_opts const& opts, function<void(wait_ctx&)> const& poller)
...
   assert(true)  // to justify trailing semicolon
 
+#define WAIT_EXPECT_NE(ctx, o1, o2)         \
+  if (o1 == o2) {                           \
+    if (ctx.fail(); !ctx.is_last_attempt) { \
+      return;                               \
+    }                                       \
+    EXPECT_NE(o1, o2);                      \
+  }                                         \
+  assert(true)  // to justify trailing semicolon
```
gaoqi8649, line 113: I don't get why need assert here. Looks like assert(true) will never fail

tests/network_test.cpp
MatusKysel: yes I need 100 blocks with 1.5k trxs so SendBlocks function has to split it. I know it because when you enable logging you can see msg, that it is splitting the packet

tests/util_test/util.hpp
MatusKysel: I will fix it, it is just for semicolon

JakubFornadel (approved):

PR #812, opened 2021-05-27
 Don't crash, log packet type and rlp-encoded length on invalid packet in Session::do_read

02p01r:


gaoqi8649 (approved):
Looks good to me

justinsnapp (approved):
Good fix!

PR #814, opened 2021-05-28
chore(unit test): Fix pbft manager unit tests timing issue

gaoqi8649:
Cannot reproduce at local, try fixing running time and waiting time

### DIFF
--- a/config/networks/testnet.json
+++ b/config/networks/testnet.json
@@ -25,14 +25,14 @@
       "level": "0x0",
       "pivot": "0x0000000000000000000000000000000000000000000000000000000000000000",
       "sig": "0xb7e22d46c1ba94d5e8347b01d137b5c428fcbbeaf0a77fb024cbbf1517656ff00d04f7f25be608c321b0d7483c402c294ff46c49b265305d046a52236c0a363701",
-      "timestamp": "0x60b03280",
+      "timestamp": "0x60b03281",
       "tips": [],
       "transactions": []
     },
     "final_chain": {
       "genesis_block_fields": {
         "author": "0x0000000000000000000000000000000000000000",
-        "timestamp": "0x60b03280"
+        "timestamp": "0x60b03281"
       },
       "state": {
         "disable_block_rewards": true,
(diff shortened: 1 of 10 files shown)


JakubFornadel:
so what are you fixing if you cannot reproduce it ?

JakubFornadel:
@justinsnapp please dont use "update-branch" button, which merges develop into the branch... We agreed we are going to use rebase for this purpose so the history(graph) is nice and clean... Currently there is no way of disabling the button button(or change its policy to rebase) and keeping the notification "your branch is not up to date". I left it enabled  so people can see their branch is not up to date, but they should manually rebase their branch instead of pushing 'update-branch' button

+ I would really like to start enforcing 2 reviews (approvals) before merging into develop. So instead of merging PR's that dont have approvals as we do now we could write to people "hey this is my PR, which is finished please either approve it or comment if you see some problems"... It might take few days/weeks for people to get used to it but it will force us to do proper reviews and that is for sure beneficial :)

justinsnapp:
@JakubFornadel yeah I agree.  I have been being bad for sure.

gaoqi8649:
> so what are you fixing if you cannot reproduce it ?
The unit tests could pass on Linux machine, but will fail on Mac machine on circleci. I checked the fail logs on circleci, find this is unit test waiting time issue.

JakubFornadel:
> > so what are you fixing if you cannot reproduce it ?
> > The unit tests could pass on Linux machine, but will fail on Mac machine on circleci. I checked the fail logs on circleci, find this is unit test waiting time issue.

hmm are you sure about 120s instead of 60s or its just raw guess ? WOuld be cool if keep this waiting time reasonably small as our tests already take a lot of time... We might optimize them 1 day when there is time for that...

gaoqi8649:
> > > so what are you fixing if you cannot reproduce it ?
> > > The unit tests could pass on Linux machine, but will fail on Mac machine on circleci. I checked the fail logs on circleci, find this is unit test waiting time issue.
> 
> hmm are you sure about 120s instead of 60s or its just raw guess ? WOuld be cool if keep this waiting time reasonably small as our tests already take a lot of time... We might optimize them 1 day when there is time for that...

From the CircleCI Mac machine running logs, knows 60s is not enough. So I just double the time. In the unit test,  will check every 1s. So it will exist if finish early. The timeout changes to 120s

MatusKysel (approved):

justinsnapp (approved):

PR #825, opened 2021-06-01
fix(networking): redefine where we mark peer as active & missing lock

MatusKysel:
## Purpose
- add missing lock
- change when we mark that peer is active

## For reviewers
What I have noticed, is that sometimes e.g. right now on testnet peer is trying to send my node a lot of missing DAG blocks (10k),
so peer splits those to multiple packets, but it takes while to process it. Meanwhile he is trying to send me status back, but is somewhere deep down in send/write queue and it will not be send in time. Then  we disconnect each other, because we did not get status packet.
One of the solutions is mark all incoming packet as some kind of activity between us, so we do not disconnect

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -213,8 +213,8 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
   if (dag_mgr_ && !peer->passed_initial_ && _id != StatusPacket) {
     return;
   }
-  //Any packet means that we are comunicating so let's not disconnect
-  //e.g. we could sned a lot of data and status packet can be in send queue
+  // Any packet means that we are comunicating so let's not disconnect
+  // e.g. we could sned a lot of data and status packet can be in send queue
   peer->setAlive();
 
   switch (_id) {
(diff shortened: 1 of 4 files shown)


JakubFornadel (approved):

src/network/taraxa_capability.cpp
```diff
@@ -1148,24 +1150,26 @@ void TaraxaCapability::sendTransactions() {
...
-      host->disconnect(peer.first, p2p::PingTimeout);
-    }
-    // Send status message
-    else {
-      sendStatus(peer.first, false);
+  {
+    boost::shared_lock<boost::shared_mutex> lock(peers_mutex_);
+    for (auto const &peer : peers_) {
+      // Disconnect any node that did not send any message for 3 status intervals
+      if (!peer.second->isAlive(5)) {
```
kstdl, line 1162: maybe this 5(max_check_count) should be config value? or at least class const?

MatusKysel: I agree, this part is just an old code that was there before

kstdl (dismissed):

JakubFornadel (approved):

kstdl (dismissed):

src/network/taraxa_capability.cpp
```diff
@@ -213,6 +213,9 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
   if (dag_mgr_ && !peer->passed_initial_ && _id != StatusPacket) {
     return;
   }
+  // Any packet means that we are comunicating so let's not disconnect
+  // e.g. we could sned a lot of data and status packet can be in send queue
```
justinsnapp, line 217: Typo in this comment... "sned" should be "send"

justinsnapp (dismissed):
Looks good.  Only comment I have is a minor spelling typo in a comment.  LUL

JakubFornadel (approved):

PR #830, opened 2021-06-02
fix(pbft): add flag for when reset own starting value to null blo…

justinsnapp:
…ck hash due to malicious block

## Version Notes ##
{ major | minor | patch }


gaoqi8649 (approved):
Looks good to me

PR #831, opened 2021-06-02
fix(pbft): Give up voted value

gaoqi8649:
## Purpose

Give up voted value when the voted value is not valid or have waited 5+ rounds

## Version Notes ##
{ major | minor | patch }


gaoqi8649:
Since the fix changes PBFT consensus, that need test carefully

justinsnapp (approved):
Very logical refactoring

PR #834, opened 2021-06-03
chore(unit test): Fix unit test NetworkTest.send_pbft_block in dynamic checking

gaoqi8649:


tests/network_test.cpp
```diff
@@ -125,15 +125,22 @@ TEST_F(NetworkTest, send_pbft_block) {
   auto nodes = launch_nodes(node_cfgs, 1);
   auto nw1 = nodes[0]->getNetwork();
   auto nw2 = nodes[1]->getNetwork();
+  EXPECT_EQ(1, nw1->getPeerCount());
+  EXPECT_EQ(1, nw2->getPeerCount());
```
mfrankovi, line 129: launch_nodes internally already checks if nodes are connected so no need to check here as well

tests/network_test.cpp
```diff
@@ -125,15 +125,22 @@ TEST_F(NetworkTest, send_pbft_block) {
...
   auto pbft_block = make_simple_pbft_block(blk_hash_t(1), 2);
   uint64_t chain_size = 111;
 
   nw2->sendPbftBlock(nw1->getNodeId(), pbft_block, chain_size);
-  taraxa::thisThreadSleepForMilliSeconds(200);
 
-  ASSERT_EQ(1, nw1->getPeerCount());
-  ASSERT_EQ(chain_size, nw1->getPeer(nw1->getAllPeers()[0])->pbft_chain_size_);
+  auto node2_id = nw2->getNodeId();
+  for (auto _(0); _ < 100; _++) {
```
mfrankovi, line 137: Could you use EXPECT_HAPPENS and WAIT_EXPECT_EQ macros that we use when we need to wait for something to become equal instead of manually using the loop

MatusKysel (approved):

tests/network_test.cpp
gaoqi8649: removed

tests/network_test.cpp
gaoqi8649: Fixed

justinsnapp (approved):

PR #836, opened 2021-06-03
perf(verification): store verified txs in to the DB with sender add, …

MatusKysel:
We were doing double verification on every trx. Once when we stored it and again where we were doing execution.

So my solution is to save sender to DB after verification (has sender ~ verified)

before:
old
after:
new

this should save 5% of cpu cycles

### DIFF
--- a/src/transaction_manager/transaction.cpp
+++ b/src/transaction_manager/transaction.cpp
@@ -34,11 +34,25 @@ Transaction::Transaction(uint64_t nonce, val_t const &value, val_t const &gas_pr
   getSender();
 }
 
-Transaction::Transaction(dev::RLP const &_rlp, bool verify_strict, h256 const &hash)
+Transaction::Transaction(dev::RLP const &_rlp, bool verify_strict, h256 const &hash, bool rlp_w_sender)
     : hash_(hash), hash_initialized_(!hash.isZero()) {
   u256 v, r, s;
-  util::rlp_tuple(util::RLPDecoderRef(_rlp, verify_strict), nonce_, gas_price_, gas_, receiver_, value_, data_, v, r,
-                  s);
+
+  if (_rlp.itemCount() < 10) {
+    util::rlp_tuple(util::RLPDecoderRef(_rlp, verify_strict), nonce_, gas_price_, gas_, receiver_, value_, data_, v, r,
+                    s);
+  } else {
+    addr_t sender;
+    util::rlp_tuple(util::RLPDecoderRef(_rlp, verify_strict), nonce_, gas_price_, gas_, receiver_, value_, data_, v, r,
+                    s, sender);
+
+    if (rlp_w_sender && !sender.isZero()) {
+      sender_initialized_ = true;
+      sender_valid_ = true;
+      sender_ = sender;
+    }
+  }
+
   // the following piece of code should be equivalent to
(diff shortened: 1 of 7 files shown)


tests/full_node_test.cpp
```diff
@@ -74,6 +74,25 @@ void send_dummy_trx() {
 
 struct FullNodeTest : BaseTest {};
 
+TEST_F(FullNodeTest, double_verification) {
+  auto db_ptr = s_ptr(new DbStorage(data_dir));
```
JakubFornadel, line 78: s_ptr(new DbStorage(data_dir)) ....

src/transaction_manager/transaction_manager.cpp
```diff
@@ -229,7 +229,7 @@ void TransactionManager::verifyQueuedTrxs() {
       auto status = db_->getTransactionStatus(hash);
       if (status == TransactionStatus::in_queue_unverified) {
         db_->saveTransactionStatus(hash, TransactionStatus::in_queue_verified);
-
+        db_->saveTransaction(*item.second, true);
```
JakubFornadel, line 232: so we have to save the same tx twice into db, for the first time when saving unverified tx and second time here ?

src/transaction_manager/transaction.cpp
```diff
@@ -104,9 +104,9 @@ addr_t const &Transaction::getSender() const {
                          "\nOriginal RLP: " + (cached_rlp_ ? dev::toJS(*cached_rlp_) : "wasn't created from rlp"));
 }
 
-template <bool for_signature>
+template <bool for_signature, bool w_sender>
```
JakubFornadel, line 107: Note: We might differentiate between standard parameters and template parameters by having template parameters names uppercase template <bool FOR_SIGNATURE, bool W_SENDER>, but not a requirement :)

src/transaction_manager/transaction_manager.cpp
MatusKysel: no it will be just this one, this will just update byte value of tx in db

src/transaction_manager/transaction.cpp
```diff
@@ -62,7 +62,7 @@ Transaction::Transaction(dev::RLP const &_rlp, bool verify_strict) {
       break;
     }
   }
-  if (fields_processed != 9) {
+  if (fields_processed > 10) {
```
gaoqi8649, line 65: From the above code, fields_processed only can be 9. Otherwise, that must include something go to else condition. What will cause that?

tests/full_node_test.cpp
```diff
@@ -74,6 +74,24 @@ void send_dummy_trx() {
...
+  auto trx_data = db.getTransactionRaw(g_trx_signed_samples[0].getHash());
+  trx_rlp = RLP(trx_data);
+  EXPECT_EQ(trx_rlp.itemCount(), 9);
+
+  db.saveTransaction(g_trx_signed_samples[0], true);  // save it verified
+  trx_data = db.getTransactionRaw(g_trx_signed_samples[0].getHash());
+  trx_rlp = RLP(trx_data);
+  EXPECT_EQ(trx_rlp.itemCount(), 10);
+}
+
```
gaoqi8649, line 94: Better move the unit test to transaction_test.cpp

src/transaction_manager/transaction.cpp
MatusKysel: Please check change in submodule https://github.com/Taraxa-project/taraxa-aleth/pull/22

gaoqi8649 (approved):

justinsnapp (approved):

PR #839, opened 2021-06-04
refactor(trx_mngr): refactor processing of block trxs

MatusKysel:
## Purpose

While working on double verification I have noticed that processing of block trxs is kind of mess. This is just small refactor of it

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -242,7 +242,7 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
         // We need logic when some different node versions might still be compatible
         if (node_major_version != FullNode::c_node_major_version ||
             node_minor_version != FullNode::c_node_minor_version) {
-          LOG(log_er_) << "Incorrect node version: " << getFormattedVersion(node_major_version, node_minor_version)
+          LOG(log_wr_) << "Incorrect node version: " << getFormattedVersion(node_major_version, node_minor_version)
                        << ", our node major version"
                        << getFormattedVersion(FullNode::c_node_major_version, FullNode::c_node_minor_version)
                        << ", host " << _nodeID << " will be disconnected";
@@ -251,13 +251,13 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
         }
 
         if (network_id != conf_.network_id) {
-          LOG(log_er_) << "Incorrect network id " << network_id << ", host " << _nodeID << " will be disconnected";
+          LOG(log_wr_) << "Incorrect network id " << network_id << ", host " << _nodeID << " will be disconnected";
           host->disconnect(_nodeID, p2p::UserReason);
           break;
         }
 
         if (genesis_hash != dag_mgr_->get_genesis()) {
-          LOG(log_er_) << "Incorrect genesis hash " << genesis_hash << ", host " << _nodeID << " will be disconnected";
+          LOG(log_wr_) << "Incorrect genesis hash " << genesis_hash << ", host " << _nodeID << " will be disconnected";
           host->disconnect(_nodeID, p2p::UserReason);
           break;
         }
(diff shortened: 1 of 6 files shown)


src/network/taraxa_capability.cpp
```diff
@@ -251,13 +251,13 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
         }
 
         if (network_id != conf_.network_id) {
-          LOG(log_er_) << "Incorrect network id " << network_id << ", host " << _nodeID << " will be disconnected";
+          LOG(log_wr_) << "Incorrect network id " << network_id << ", host " << _nodeID << " will be disconnected";
```
JakubFornadel, line 254: We could use _nodeID.abridged() in these logs

src/transaction_manager/transaction_queue.cpp
```diff
@@ -92,43 +92,56 @@ void TransactionQueue::removeTransactionFromBuffer(trx_hash_t const &hash) {
 }
 
 void TransactionQueue::addTransactionToVerifiedQueue(trx_hash_t const &hash, std::list<Transaction>::iterator iter) {
-  uLock lock(shared_mutex_for_verified_qu_);
+  uLock verify_lock(shared_mutex_for_verified_qu_);
+  {
+    // this can happend when tx is verified in backround thread
+    // and in same time we verified block, so it was already removed
```
JakubFornadel, line 98: hmm I would not really know what this is about if we did not have personal convo about it...

src/transaction_manager/transaction_manager.cpp
```diff
@@ -412,28 +350,71 @@ void TransactionManager::packTrxs(vec_trx_t &to_be_packed_trx, uint16_t max_trx_
...
+  std::set<trx_hash_t> known_trx_hashes(all_block_trx_hashes.begin(), all_block_trx_hashes.end());
+
+  auto trx_batch = db_->createWriteBatch();
   for (auto const &trx : trxs) {
-    auto valid = verifyTransaction(trx);
-    if (!valid.first) {
-      invalidTransaction = true;
-      LOG(log_er_) << "Invalid transaction " << trx.getHash().toString() << " " << valid.second;
+    known_trx_hashes.erase(trx.getHash());
+    auto status = db_->getTransactionStatus(trx.getHash());
```
JakubFornadel, line 367: Lets optimize this one. You already have vector of all txs hashes, there is a method to get all txs statuses from db at once and parameter is vector of hashes...

JakubFornadel: hmm okay you dont have the vector of hashes, but might be still worth to loop over txs first and get hashes + 1 db call instead off N db calls.... ???

JakubFornadel (approved):

src/transaction_manager/transaction_manager.cpp
MatusKysel: but status can be outdated as verification of all unverified trxs is done in background by other threads, so this way as it is you will get as much up to date status as possible. And one more thing reading from DB is cheap, re-validating is expensive

src/network/taraxa_capability.cpp
```diff
@@ -251,13 +251,13 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
...
           host->disconnect(_nodeID, p2p::UserReason);
           break;
         }
 
         if (genesis_hash != dag_mgr_->get_genesis()) {
-          LOG(log_er_) << "Incorrect genesis hash " << genesis_hash << ", host " << _nodeID << " will be disconnected";
+          LOG(log_wr_) << "Incorrect genesis hash " << genesis_hash << ", host " << _nodeID << " will be disconnected";
```
gaoqi8649, line 260: Same reason as above

src/network/taraxa_capability.cpp
```diff
@@ -242,7 +242,7 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
         // We need logic when some different node versions might still be compatible
         if (node_major_version != FullNode::c_node_major_version ||
             node_minor_version != FullNode::c_node_minor_version) {
-          LOG(log_er_) << "Incorrect node version: " << getFormattedVersion(node_major_version, node_minor_version)
+          LOG(log_wr_) << "Incorrect node version: " << getFormattedVersion(node_major_version, node_minor_version)
```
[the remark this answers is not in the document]
MatusKysel: OK

mfrankovi (approved):

PR #842, opened 2021-06-05
changed strategy for creating releases combining both workspaces (mac…… and linux) into one dir

agrebin:
## Purpose

Fixes issues with some releases when replacing existing tags
## For reviewers

Changed the strategy of having one job per binary release to have just one after all binaries were succesfully compiled and tested, making just one upload of the release

## Version Notes ##
{ major | minor | patch }


justinsnapp (approved):
Thanks for this fix!

sfrique (approved):

PR #843, opened 2021-06-07
Ethereum/FinalChain code refactoring, Revisited Ethereum json-rpc impl., and more

02p01r:
1) Ethereum/FinalChain code refactoring.
- Removed all taraxa-aleth high-level ethereum code, rewrote from scratch according to our needs, making it more compact and efficient.
- Redesigned FinalChain database schema, in particular we don't store full executed transactions twice now, and have better performance on querying executed transaction hashes. Also removed some redundant data that was stored previously (log block blooms) 
- Merged FinalChain and Executor, removed Executor.
- Put all the FinalChain-related code to one package, eliminated all it's dependencies except the DB 
2) Implemented eth json-rpc subscription API and revisited other json-rpc implementations according to the specs https://ethereum.org/en/developers/docs/apis/json-rpc/. Refactored json-rpc code.
3) Added python testing subproject, tested Ethrereum json-rpc: https://github.com/Taraxa-project/taraxa-node/pull/861.
4) Reverted synchronous db commit in taraxa-evm, now FinalChain recovery code considers taraxa-evm db being arbitrarily far behind FinalChain db.
5) Refactored DbStorage.
- Changed the batch type from `shared_ptr<Batch>` to just `Batch`
- Added some Slice converters
- Unified methods for inserting (with or without batch), reading, and deleting arbitrarily-typed keys and values
6) We now have only one Transaction class in the code base - taraxa::Transaction
7) Misc improvements all over the code.
- Added a publish/subscribe framework (`src/util/event.hpp`), used it in ethereum json-rpc subscription APIs 
- Extended the RLP framework (`src/util/encoding_rlp.hpp`): added more supported types and created a set of macros that allow to describe RLP serialization for arbitrary structures.
- In Transaction: added an ability to provide known precomputed hash to the constructor; added an indicator that a transaction was default-constructed. Used the improved RLP framework in the constructor. 

Also see https://github.com/Taraxa-project/taraxa-aleth/pull/24
Also see https://github.com/Taraxa-project/taraxa-evm/pull/28


MatusKysel:
It is very strange that you increased execution time of almost all tests (except `state_api_test`), that does not look good at all

MatusKysel:
>     1. Ethereum/FinalChain code refactoring.
> 
> 
>     * Removed all taraxa-aleth high-level ethereum code, rewrote from scratch according to our needs, making it more compact and efficient.
> 
>     * Redesigned FinalChain database schema
> 
>     * Merged FinalChain and Executor, removed Executor.
> 
>     * Put all the FinalChain-related code to one package, eliminated all it's dependencies except the DB
> 
> 
>     1. Implemented eth json-rpc subscription API and revisited other json-rpc implementations according to the specs https://ethereum.org/en/developers/docs/apis/json-rpc/. Refactored json-rpc code.
> 
>     2. Added python testing subproject, tested Ethrereum json-rpc [WIP].
> 
>     3. Reverted synchronous db commit in taraxa-evm, now FinalChain recovery code considers taraxa-evm db being arbitrarily far behind FinalChain db.
> 
>     4. Refactored DbStorage.
> 
> 
>     * Changed the batch type from `shared_ptr<Batch>` to just `Batch`
> 
>     * Added some Slice converters
> 
>     * Unified methods for inserting (with or without batch), reading, and deleting arbitrarily-typed keys and values
> 
> 
>     1. Misc improvements all over the code.
> 
> 
>     * Added a publish/subscribe framework (`src/util/event.hpp`), used it in ethereum json-rpc subscription APIs
> 
>     * Extended the RLP framework (`src/util/encoding_rlp.hpp`): added more supported types and created a set of macros that allow to describe RLP serialization for arbitrary structures.
> 
>     * In Transaction: added an ability to provide known precomputed hash to the constructor; added an indicator that a transaction was default-constructed. Used the improved RLP framework in the constructor.

I think you should split this PR exactly as you describe it here, to at least 6 separate PR, because right now you are mixing lot of unrelated staff together

JakubFornadel:
So just by looking at the "-" commits I guess you are going to squash all the commits together and then merge as you did before on your PR's. As a result current PR description will be commit message. So 1 single commit adds 5k lines, deletes 3k lines. You are changing stuff in multiple unrelated areas of the code. Now imagine you introduced some new bug, how on earth are we supposed to find out  which part of this gigantic commit did that ? This PR should have been multiple PR's + each PR might have multiple commits that could be then easily cherry-picked, reverted, tested, etc...

src/CMakeLists.txt
```diff
@@ -1,101 +1,47 @@
...
-        transaction_manager/transaction_order_manager.hpp
-        consensus/vote.hpp
-        transaction_manager/transaction.hpp
-        transaction_manager/transaction_queue.hpp
-        logger/logger_config.hpp
-        logger/log.hpp
-        chain/state_api.hpp
-        dag/dag_block_manager.hpp
-        dag/proposal_period_levels_map.hpp
-        consensus/pbft_manager.hpp
```
JakubFornadel, line 55: put list of headers back

src/common/types.hpp
```diff
@@ -1,16 +1,33 @@
...
+using dev::AddressSet;
+using dev::bytes;
+using dev::bytesConstRef;
+using dev::h256;
+using dev::h256Hash;
+using dev::h256s;
+using dev::h64;
+using dev::Secret;
+using dev::u256;
+
```
JakubFornadel, line 28: Having using in headers is very bad practice. You just pollute namespaces thanks to this as every file that includes "types.hpp" will contain these usings, which is pretty bad. Here are some explanation related to this: https://jmmv.dev/2013/12/header-files-avoid-c-using-directives.html
https://stackoverflow.com/questions/19940081/using-keyword-in-header-implementation-in-c

src/final_chain/final_chain.hpp
```diff
@@ -0,0 +1,96 @@
...
+#include "util/event.hpp"
+#include "util/exit_stack.hpp"
+#include "util/range_view.hpp"
+
+namespace taraxa::final_chain {
+using namespace ::std;
+using namespace ::dev;
+using namespace ::taraxa::final_chain;
+using namespace ::taraxa::util;
+
```
JakubFornadel, line 19: again, have these "usings" in cpp files.

src/network/rpc/EthClient.h
```diff
@@ -0,0 +1,384 @@
...
+ * This file is generated by jsonrpcstub, DO NOT CHANGE IT MANUALLY!
+ */
+
+#ifndef JSONRPC_CPP_STUB_TARAXA_NET_ETHCLIENT_H_
+#define JSONRPC_CPP_STUB_TARAXA_NET_ETHCLIENT_H_
+
+#include <jsonrpccpp/client.h>
+
+namespace taraxa {
+    namespace net {
```
JakubFornadel, line 11: we are using 2 space padding in whole codebase

src/network/rpc/NetClient.h
```diff
@@ -2,46 +2,50 @@
...
-      return result.asBool();
-    else
-      throw jsonrpc::JsonRpcException(jsonrpc::Errors::ERROR_CLIENT_INVALID_RESPONSE, result.toStyledString());
-  }
-};
-
-}  // namespace net
-}  // namespace taraxa
+                std::string net_version() throw (jsonrpc::JsonRpcException)
+                {
```
JakubFornadel, line 18: we are using 2 space padding in whole codebase

src/network/rpc/TestClient.h
```diff
@@ -2,235 +2,260 @@
  * This file is generated by jsonrpcstub, DO NOT CHANGE IT MANUALLY!
  */
 
-#pragma once
+#ifndef JSONRPC_CPP_STUB_TARAXA_NET_TESTCLIENT_H_
+#define JSONRPC_CPP_STUB_TARAXA_NET_TESTCLIENT_H_
```
JakubFornadel, line 7: I guess these guards were generated automatically ? As we are using pragmas but this seems like generation-related issue...

tests/pbft_manager_test.cpp
```diff
@@ -64,7 +64,7 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
     trxs_count++;
     EXPECT_HAPPENS({120s, 1s}, [&](auto &ctx) {
       for (auto &node : nodes) {
-        if (ctx.fail_if(!node->getFinalChain()->isKnownTransaction(trx.getHash()))) {
+        if (ctx.fail_if(!node->getFinalChain()->transaction_location(trx.getHash()))) {
```
JakubFornadel, line 67: I believer we are using camel case for function names

tests/CMakeLists.txt
```diff
@@ -1,6 +1,12 @@
 # Enable testing for current directory and below.
 enable_testing(true)
 
+# TODO
+#add_executable(py_test py_test.cpp)
+#add_dependencies(py_test taraxad)
+#target_compile_definitions(py_test PRIVATE NODE_EXECUTABLE_PATH=\"${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/taraxad\")
+#add_test(py_test ${CMAKE_RUNTIME_OUTPUT_DIRECTORY}/py_test)
+
```
JakubFornadel, line 9: is this supposed to be commented out ? How is the new test run then ?

src/node/full_node.hpp
```diff
@@ -94,9 +94,8 @@ class FullNode : public std::enable_shared_from_this<FullNode> {
   void init();
   void close();
 
-  template <typename T, typename... ConstructorParams>
-  auto &emplace(std::shared_ptr<T> &ptr, ConstructorParams &&... ctor_params) {
-    ptr = std::make_shared<T>(std::forward<ConstructorParams>(ctor_params)...);
+  template <typename T>
+  auto const &register_s_ptr(std::shared_ptr<T> const &ptr) {
```
JakubFornadel, line 98: I was looking at this before. Can you please explain me what is this even good for ?

src/node/full_node.cpp
```diff
@@ -204,53 +233,22 @@ void FullNode::start() {
...
+}
 
 void FullNode::close() {
-  util::ExitStack finally;
-  // because `this` shared ptr is given to it
-  finally += [this] { jsonrpc_api_.reset(); };
   if (bool b = false; !stopped_.compare_exchange_strong(b, !b)) {
     return;
   }
+  jsonrpc_api_ = nullptr;  // TODO Because it indirectly refers to FullNode
```
JakubFornadel, line 244: TODO what ?

src/network/rpc/eth/impl/Watches.hpp
```diff
@@ -0,0 +1,193 @@
...
+// TODO taraxa simple exception
+DEV_SIMPLE_EXCEPTION(WatchLimitExceeded);
+
+using WatchID = uint64_t;
+
+inline auto const watch_id_type_mask_bits = WatchType(ceil(log2(float(COUNT))));
+
+struct placeholder_t {};
+template <WatchType type_, typename InputType_, typename OutputType_ = placeholder_t, typename Params = placeholder_t>
+struct WatchGroup {
```
JakubFornadel, line 21: why do you use struct instead of class for such relatively complex data stucture ?

src/network/rpc/CMakeLists.txt
```diff
@@ -1,35 +1,20 @@
...
 
-foreach (api_name Net Taraxa Test)
+set(CPP_NAMESPACE taraxa::net)
+file(GLOB API_DEF_FILES "${CMAKE_CURRENT_SOURCE_DIR}/*.jsonrpc.json")
+add_custom_target(rpc_stubs)
+foreach (api_def_file ${API_DEF_FILES})
+    get_filename_component(api_name ${api_def_file} NAME_WE)
     jsonrpcstub_create(
-            gen_rpc_stubs ${api_name}.jsonrpc.json
+            rpc_stubs ${api_name}.jsonrpc.json
```
JakubFornadel, line 14: well gen_rpc_stubs is definitely more clear in case this custom target is still supposed to generate rpc stubs :)

src/final_chain/final_chain.cpp
```diff
@@ -0,0 +1,440 @@
...
+#include "replay_protection_service.hpp"
+#include "util/thread_pool.hpp"
+
+namespace taraxa::final_chain {
+using namespace std;
+using namespace dev;
+
+enum DBMetaKeys { LAST_NUMBER = 1 };
+
+struct FinalChainImpl final : FinalChain {
```
JakubFornadel, line 14: class...

src/final_chain/final_chain.cpp
```diff
@@ -0,0 +1,440 @@
...
+  shared_ptr<DB> db_;
+  StateAPI state_api;
+  unique_ptr<ReplayProtectionService> replay_protection_service_{new ReplayProtectionServiceDummy};
```
JakubFornadel, line 17: please fix this. Using "new" is just huge NO

src/final_chain/final_chain.cpp
```diff
@@ -0,0 +1,440 @@
...
+    // Update number of executed DAG blocks and transactions
+    auto num_executed_dag_blk = num_executed_dag_blk_ + new_blk.dag_blk_hashes.size();
+    auto num_executed_trx = num_executed_trx_ + to_execute.size();
+    if (!new_blk.dag_blk_hashes.empty()) {
+      db_->insert(batch, DB::Columns::status, StatusDbField::ExecutedBlkCount, num_executed_dag_blk);
+      db_->insert(batch, DB::Columns::status, StatusDbField::ExecutedTrxCount, num_executed_trx);
+      LOG(log_nf_) << "Executed dag blocks #" << num_executed_dag_blk_ - new_blk.dag_blk_hashes.size() << "-"
+                   << num_executed_dag_blk_ - 1 << " , Transactions count: " << to_execute.size();
+    }
+    auto result = s_ptr(new FinalizationResult{
```
JakubFornadel, line 155: do not use "new", use make_shared/unique instead

src/final_chain/final_chain.cpp
```diff
@@ -0,0 +1,440 @@
...
+      return h256((uint8_t*)(serialized_.data() + i * h256::size), h256::ConstructFromPointer);
+    }
+
+    size_t count() const override { return count_; }
+  };
+};
+
+unique_ptr<FinalChain> NewFinalChain(shared_ptr<DB> const& db, Config const& config, Opts const& opts,
+                                     addr_t const& node_addr) {
+  return u_ptr(new FinalChainImpl(db, config, opts, node_addr));
```
JakubFornadel, line 437: do not use "new", use make_shared/unique instead

JakubFornadel:
Seems there might be actually a lot of good stuff here, just the way this PR is done degrades the overall quality a lot... I dont even know where and how to do some logical review here as there is so many lines and so much stuff mixed here... I tend to agree with Matus that this should have been split into multiple different PR's with proper commits descriptions etc...

src/common/types.hpp
02p01r: It is equivalent to a bunch of typedefs that we already have in that header. Therese particular types are so ubiquitous in our code that I thought it is okay to have them globally defined in our namespace. Would you be more accepting to `using h64 = dev::h64` for example (exactly same thing)? Or you think those type names should not belong to `taraxa` namespace at all?

src/network/rpc/EthClient.h
02p01r: this is a generated code

src/network/rpc/NetClient.h
02p01r: this is a generated code

src/network/rpc/TestClient.h
02p01r: this is a generated code

tests/pbft_manager_test.cpp
02p01r: agree that the naming convention should be consistent. we have established naming for some types of members, but not for functions. I hope this is not critical to you right now, and anyway it should be fixed at once with a sweeping change.

tests/CMakeLists.txt
02p01r: The python tests are in progress, I believe there's a TODO above those commented lines that suggests that it should be uncommented eventually.
How this test will work: we compile a launcher executable that will itself just run the script `tests/py/run.sh` which does all the job. It seems ugly that we have to create an intermediate executable for this - I made it to stick to the pattern we have here. But I can also implement it via cmake custom target if you want.

src/node/full_node.hpp
02p01r: Long ago we had been having a devastating problem in the tests: nodes would randomly not fully terminate when a test finishes, and the next test would then fail on acquiring some unreleased OS resource like network port or rocksdb db lock.
I fixed it by refactoring the node lifecycle, and put this safeguard to make sure we don't have the problem again.
The idea of it is to make sure that all the shared pointers inside full node are released when the node is destroyed.

src/node/full_node.cpp
02p01r: I will make it more clear in the comment.
TODO - remove that line. We should not care about explicitly controlling destruction of objects. The lifecycle of objects should be as declarative as possible (kind of RAII).

src/network/rpc/eth/impl/Watches.hpp
02p01r: The only difference between class and struct in C++ is that class has private visibility by default and struct - public.

I use class or struct keyword only with this in mind. For instance, if I want the first definitions in a class/struct be private - I will use `class`. Otherwise - `struct`

src/network/rpc/CMakeLists.txt
02p01r: I will rename it to gen_rpc_stubs

src/final_chain/final_chain.cpp
02p01r: https://github.com/Taraxa-project/taraxa-node/pull/843#discussion_r648265919

src/final_chain/final_chain.cpp
02p01r: ok

src/final_chain/final_chain.cpp
02p01r: Here I used `new` just because I wanted to use aggregate initialization. make_shared doesn't allow to do it unfortunately. That means I need to define a normal constructor in FinalizationResult, which I wanted to avoid. You may have noticed that I prefer aggregate structs for simple data types. 

I will find a way to avoid `new` there

src/final_chain/final_chain.cpp
02p01r: ok about make_unique.

>What is this whole NewFinalChain function even good for ?

It is the only publicly exposed thing in this .cpp file. The idea of this pattern is to completely hide the implementation. The header defines an interface + a factory (this function). And the only job .cpp file does - provide an interface implementation by defining the factory.

02p01r:
> Seems there might be actually a lot of good stuff here, just the way this PR is done degrades the overall quality a lot... I dont even know where and how to do some logical review here as there is so many lines and so much stuff mixed here... I tend to agree with Matus that this should have been split into multiple different PR's with proper commits descriptions etc...

As discussed offline, I will simulate a history of well-defined and build-able commits based on these changes

src/CMakeLists.txt
02p01r: Ok, but is there any particular reason for this? As far as I know there's no effect from listing headers as compilation units

02p01r:
> So just by looking at the "-" commits I guess you are going to squash all the commits together and then merge as you did before on your PR's. As a result current PR description will be commit message. So 1 single commit adds 5k lines, deletes 3k lines. You are changing stuff in multiple unrelated areas of the code. Now imagine you introduced some new bug, how are we supposed to find out which part of this gigantic commit did that ? This PR should have been multiple PR's + each PR might have multiple commits that could be then easily cherry-picked, reverted, tested, etc...

Right, I wanted to turn those "-" in one big description using squash.
The rest of my answer to this overlaps with this thread: https://github.com/Taraxa-project/taraxa-node/pull/843#issuecomment-857667530

src/storage/db_storage.hpp
```diff
@@ -58,13 +58,23 @@ class DbException : public exception {
...
+  using Batch = rocksdb::WriteBatch;
   using OnEntry = function<bool(Slice const&, Slice const&)>;
 
-  struct Column {
-    string const name;
+  class Column {
+    string const name_;
+
+   public:
     size_t const ordinal;
```
MatusKysel, line 73: why no suffix *_ here?

src/util/encoding_rlp.hpp
```diff
@@ -13,129 +13,215 @@ namespace taraxa::util::encoding_rlp {
 using namespace std;
 using namespace dev;
 
+using RLPEncoderRef = RLPStream&;
+struct RLPDecoderRef {
+  RLP const& target;
+  RLP::Strictness strictness;
```
MatusKysel, line 19: also missing *_ suffix

src/util/encoding_rlp.hpp
```diff
@@ -13,129 +13,215 @@ namespace taraxa::util::encoding_rlp {
...
+    rlp(encoding, target.emplace());
   }
 }
 
 template <typename Sequence>
-void dec_rlp_sequence(RLP const& rlp, Sequence& target) {
-  for (auto const i : rlp) {
-    dec_rlp(i, target.emplace_back());
+auto rlp(RLPDecoderRef encoding, Sequence& target) -> decltype(target.emplace_back(), void()) {
+  for (auto const& i : encoding.target) {
```
MatusKysel, line 121: remove & from here also up in few more loop like this. This will trigger warning under clang  (https://github.com/Taraxa-project/taraxa-node/pull/794)

src/storage/db_storage.hpp
02p01r: I believe the "_" suffix is only for private fields. Perhaps I got this convention wrong though. To me, a public field with underscore seems quite weird when you access it from outside.

src/util/encoding_rlp.hpp
02p01r: https://github.com/Taraxa-project/taraxa-node/pull/843#discussion_r648598852

src/util/encoding_rlp.hpp
02p01r: ok

src/CMakeLists.txt
JakubFornadel: yes there is. To see headers for IDE-generated projects you have to provide list of headers. Just put it back
https://stackoverflow.com/questions/36174499/why-add-header-files-into-add-library-add-executable-command-in-cmake

src/network/rpc/EthClient.h
JakubFornadel: so there is no possibility to set 2 space padding in tool that generated it ?

src/common/types.hpp
JakubFornadel: I think it should not be in taraxa namespace at all but I see that this practice is broadly used almost everywhere so it might be more of a task for bigger refactor...

tests/pbft_manager_test.cpp
JakubFornadel: Really ? Because when I joined the project I was told that variables are underscore, class names and functions are camel-case

src/network/rpc/eth/impl/Watches.hpp
JakubFornadel: This is more about conventions. I will quote Stroustrup: "structs are primarily used for data structures where the members can take any value; that is, we can't define any meaningful invariant "  - you can find this in book "Principles and Practice".

In other words struct in modern C++ are used mostly for short data structures where all members are public, someone uses it purely for POD's, etc....

Software engineering is also about making life easier to other devs so if there is sort of 'rule' in cpp community and people are used to it, lets stick to that.

src/final_chain/final_chain.cpp
JakubFornadel: ok for factory function

src/storage/db_storage.hpp
JakubFornadel: I believe you got it wrong. Underscore is used to easily distinguish between class and non-class variables. All class members should have _ suffix. 

https://google.github.io/styleguide/cppguide.html#Variable_Names - "Data members of classes (but not structs) additionally have trailing underscores. For instance: a_local_variable, a_struct_data_member, a_class_data_member_."

src/node/full_node.hpp
JakubFornadel: Ok thanks for explanation. I get the reasoning although I do think that this mechanism lost meaning once you fixed problems with shared pointers... But that will be separate task to refactor this probably...

tests/CMakeLists.txt
JakubFornadel: well plain TODO does not say much to me of what should be eventually done. But if I understand you correctly you are going to finish these python tests and uncomment it before its merged so ok :) Will keep this convo opened then...

src/node/full_node.cpp
JakubFornadel: yeah this whole close() function might be in destructor I guess without the need to manually reset shared ptrs, although the author maybe implemented it this way for some reason...

src/util/encoding_rlp.hpp
02p01r: Is this a misuse of `struct`? If not, then the naming is correct as per the Google naming convention. @JakubFornadel @MatusKysel please confirm

MatusKysel: I do no think that this is misuse, so this should be alright as it is

src/CMakeLists.txt
02p01r: done

src/network/rpc/EthClient.h
02p01r: done - included `.h` files for clang-format

src/network/rpc/NetClient.h
02p01r: https://github.com/Taraxa-project/taraxa-node/pull/843#discussion_r649886637

src/network/rpc/TestClient.h
02p01r: @JakubFornadel please resolve - can't influence the generator

tests/pbft_manager_test.cpp
02p01r: Agreed offline that we don't aim at function naming consistency at this point, please resolve @JakubFornadel

src/node/full_node.cpp
02p01r: Improved the comment, please resolve

src/network/rpc/eth/impl/Watches.hpp
02p01r: done

src/network/rpc/CMakeLists.txt
02p01r: done

src/final_chain/final_chain.cpp
02p01r: done

src/final_chain/final_chain.cpp
02p01r: done

src/final_chain/final_chain.cpp
02p01r: done

src/final_chain/final_chain.cpp
02p01r: done with the `new`

src/util/encoding_rlp.hpp
02p01r: done

02p01r:
@Taraxa-project/ledger Also went over all known struct/class field naming problems that were not mentioned and fixed them

src/storage/db_storage.hpp
02p01r: done

02p01r:
@Taraxa-project/ledger Fixed the test performance degradation.

02p01r:
> @Taraxa-project/ledger Fixed the test performance degradation.

The explanation of the perf issue:
When I first worked on this PR ~4 months ago, I added an uninterruptible sleep in one place not realizing what I did, and forgot about it.
It was in Watches.hpp (then renamed to watches.hpp/cpp). The watches can be installed via json-rpc API, and they have to have an idle timeout. When the idle timeout is hit, we have to remove those stale watches so to be nice about the resources. So I made it in a daemon thread that periodically does the cleanup. Obviously it cannot busy-spin all the time, so I added a sleeping interval, but that sleep was uninterruptible (`std::this_thread::sleep_for` - OS-level sleep).
I changed it to a sleep through `std::condition_variable` (wait with timeout). When the destructor of Watches is called, I notify that condition variable, so the destructor exists ASAP (it joins the watch cleaner thread). Now it's okay

02p01r:
@Taraxa-project/ledger Moved the python tests to another PR https://github.com/Taraxa-project/taraxa-node/pull/861

02p01r:
> > Seems there might be actually a lot of good stuff here, just the way this PR is done degrades the overall quality a lot... I dont even know where and how to do some logical review here as there is so many lines and so much stuff mixed here... I tend to agree with Matus that this should have been split into multiple different PR's with proper commits descriptions etc...
> 
> As discussed offline, I will simulate a history of well-defined and build-able commits based on these changes

I attempted to split it into several independent build-able commits, but I cannot achieve this at all:
All the changes in taraxa_node have to stem from the taraxa_aleth changes, where I removed all the high-level ethereum code, including ethereum json-rpc impl.
That removed stuff in taraxa_aleth has to have an immediate answer in taraxa_node. The replacement for it is the new src/final_chain package and the new ethereum json-rpc implementation code that depends on final_chain. src/final_chain  also depends on the src/util package changes (encoding_rlp.hpp, event.hpp). It also depends on taraxa Transaction and DbStorage changes.
PbftManager changes immediately depend on final chain, because FinalChain interface has changed.
The rest of code changes just stem from the above (renamings, removing some includes, etc.)

MatusKysel (approved):

JakubFornadel (approved):

JakubFornadel (changes requested):
@02p01r please squash before merging. Squash & merge button is disabled

02p01r:
> @02p01r please squash before merging. Squash & merge button is disabled

Done, please check.

JakubFornadel (approved):

PR #846, opened 2021-06-08
feat: add additional cli options, auto generate config, refactor config

mfrankovi:
Adding some CLI tools for generating keys in main binary
Refactoring processing command line options
Generating default config files for devnet and testnet
Separating wallet from other config


src/config/cli_config.cpp
```diff
@@ -0,0 +1,145 @@
...
+    return;
+  }
+
+  // Create dir if missing
+  if (!fs::exists(data_dir)) {
+    fs::create_directory(data_dir);
+  }
+  if (!fs::exists(default_dir)) {
+    fs::create_directory(default_dir);
+  }
```
sfrique, line 94: With default options this should fail as line 90 does not create the parent directory `.taraxa`  and the default `data_dir` is `.taraxa/db` 
Changing the order may solve the issue with default values.

I am not sure if we want to create parent dirs automatically when passed as parameters, if we do, this won't work as well, it would need to check all parents and create them as needed.

sfrique (changes requested):
The tests are failing because the binary isn't able to start with the default parameters, (see review comments)

Another thing to note is should we keep the `cli` option?
https://github.com/Taraxa-project/taraxa-node/blob/0b8da9c51a5ceb445a723ee03c2f5a8e93dfcc90/docker-entrypoint.sh#L10
 @justinsnapp I ask because I recall you wanted more than just the config, you also wanted some sort of rpc calls and such.

If we don't need the cli I think this PR would remove the need for the cli folder, entry point commands, and also the docker image as well.
As most of them are now behind `--command` flag.

Or we can add the clean-up in another PR as we can give more thought to the `entry point` script and maybe even leave only `exec` and `taraxa` as pretty much everything will be only a few parameters to the new `taraxad` binary.


a few logs to add information...

Failed because there is no parent dir for db to be created.
```
root@86b865345d7b:/opt/taraxa# taraxad
Dynamic exception type: std::filesystem::__cxx11::filesystem_error
std::exception::what: filesystem error: cannot create directory: No such file or directory [/root/.taraxa/db]

root@86b865345d7b:/opt/taraxa# mkdir /root/.taraxa
root@86b865345d7b:/opt/taraxa# taraxad
Configuration file does not exist at: /root/.taraxa/config.json. New config file will be generated
Wallet file does not exist at: /root/.taraxa/wallet.json. New wallet file will be generated
Account secret key: f2cab09d5e03185adba6ea4da8144f5fee54f0605cdcbf20b145aaf3df25ea0c
VRF secret key: 737824a355dcfdc2d478166a3f523dff7a3b126cd3bdf7727b8fbe0bce33d03b31173cf6413cc92fc40a051dbbf697605405ac0dc1197074b9eb621371b03532
2021-06-08 22:15:25.034975 FULLND SILENT Node public key: a01e859ac616d40366f0a663473c22d57af90ddae1acf5ed27468a59f50aedca67c7434259bb56118f5c851c70ea682d0d5ef5fa6cd5ef7c1db66ccaf625e55c
Node address: b8f532f12104550208f9d354ed08ca67b5ccf5f2
Node VRF public key: 31173cf6413cc92fc40a051dbbf697605405ac0dc1197074b9eb621371b03532
...
```
Working only adding the `data-dir` option on an existing parent folder
```
root@86b865345d7b:/opt/taraxa# ls -ls /root/.taraxa
ls: cannot access '/root/.taraxa': No such file or directory
root@86b865345d7b:/opt/taraxa# mkdir /data-dir
root@86b865345d7b:/opt/taraxa# taraxad --data-dir /data-dir/test
Configuration file does not exist at: /root/.taraxa/config.json. New config file will be generated
Wallet file does not exist at: /root/.taraxa/wallet.json. New wallet file will be generated
Account secret key: 4e351bb598f68af49ba336feafd4d2ab7ecdd106e5f7ad17b4d63ad02a76c232
VRF secret key: 13f19f1751373de19678f8434ffa2769499fd34ef7b51a9949de89c9d5a1a6c5e73687f4b9627f1f34c5920921bb6f5269eafbb1ac1e49aeca701d9d1f8c08fc
2021-06-08 22:19:24.382205 FULLND SILENT Node public key: cb43e9a6b84b6749a0a5c9d559b9ef0287eceb631059ac14bd3b8a4b4ef2b59b2b9bd6819b092ab63d9292b8e6bbf0bd6376798036f0b1a3479b8218b7237fda
Node address: c1a7ffffb4acfbd41162c12d212b2ac9936a7804
Node VRF public key: e73687f4b9627f1f34c5920921bb6f5269eafbb1ac1e49aeca701d9d1f8c08fc
2021-06-08 22:19:25.326312 RPC SILENT Taraxa WS started at port: 0.0.0.0:8777

...
```

Failed on a `data-dir` without parent dir.

```
root@86b865345d7b:/opt/taraxa# ls -ls /data-dir/no-parent/
ls: cannot access '/data-dir/no-parent/': No such file or directory
root@86b865345d7b:/opt/taraxa# taraxad --data-dir /data-dir/no-parent/test
Dynamic exception type: std::filesystem::__cxx11::filesystem_error
std::exception::what: filesystem error: cannot create directory: No such file or directory [/data-dir/no-parent/test]
```

docker-entrypoint.sh
```diff
@@ -17,25 +17,19 @@ case $1 in
...
   join)
-	cli/taraxa config \
-        node \
-        --network $2 \
-        --file $TARAXA_CONF_PATH
-
     echo "Starting taraxad..."
     taraxad \
-            --conf_taraxa $TARAXA_CONF_PATH
+            --config $TARAXA_CONF_PATH
```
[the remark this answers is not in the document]
mfrankovi: With the new changes, it should work as is. So I am not sure if there is a need to remove this?

src/config/cli_tools.cpp
```diff
@@ -0,0 +1,143 @@
...
+  auto [pk, sk] = taraxa::vrf_wrapper::getVrfKeyPair();
+  auto account_json = createWalletJson(account, sk, pk);
+
+  // Create account file
+  writeJsonToFile(wallet, account_json);
+}
+
+void CliTools::generateAccount(dev::KeyPair account) {
+  cout << "secret : " << toHex(account.secret().ref()) << endl;
+  ;
```
JakubFornadel, line 47: what is this extra semicolon for ?

src/config/cli_tools.hpp
```diff
@@ -0,0 +1,46 @@
...
+
+  // Account generation
+  static void generateAccount();
+  static void generateAccount(dev::KeyPair account);
+  static void generateAccountFromKey(std::string key);
+
+  // VRF generation
+  static void generateVrf();
+  static void generateVrf(taraxa::vrf_wrapper::vrf_sk_t sk, taraxa::vrf_wrapper::vrf_pk_t pk);
+  static void generateVrfFromKey(std::string key);
```
JakubFornadel, line 26: is there any reason why "const &" is not used for functions in lines 19 - 26 ?

src/config/cli_tools.cpp
```diff
@@ -0,0 +1,143 @@
...
+  json["vrf_secret"] = sk.toString();
+  json["vrf_public"] = pk.toString();
+  return json;
+}
+
+Json::Value CliTools::getAccountSecretKey(const string &wallet) { return readJsonFromFile(wallet)["node_secret"]; }
+
+Json::Value CliTools::getVrfSecretKey(const string &wallet) { return readJsonFromFile(wallet)["vrf_secret"]; }
+
+string CliTools::getHomeDir() { return string(getpwuid(getuid())->pw_dir); }
```
JakubFornadel, line 93: I guess for this use case it is ok as there is not many calls to these functions but in general might be a better idea to create all these variables in constructor and then just return const std::string &

src/config/cli_config.hpp
```diff
@@ -0,0 +1,70 @@
...
+  std::string data_dir;
+  std::vector<std::string> command;
+  bool boot_node = false;
+  bool destroy_db = 0;
+  bool rebuild_network = 0;
+  bool rebuild_db = 0;
+  bool version = false;
+  uint64_t rebuild_db_period = 0;
+  uint64_t revert_to_period = 0;
+  std::shared_ptr<FullNodeConfig> node_config;
```
JakubFornadel, line 44: what is the benefit/need for node_config to be a shared_ptr instead of just FullNodeConfig ?

src/config/cli_config.hpp
```diff
@@ -0,0 +1,70 @@
...
+  enum NetworkIdType { Mainnet = 1, Testnet, Devnet };
+
+ protected:
+  static constexpr int DEFAULT_NETWORK_ID = Testnet;
+
+  boost::program_options::options_description main_options;
+  boost::program_options::options_description node_command_options;
+  boost::program_options::options_description allowed_options;
+
+  std::string config;
```
JakubFornadel, line 31: "_" suffix is used for class members...

src/config/cli_config.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+  ~CliConfig() {}
+
+  // Returns true if node configuration is loaded successfully and command is node
+  bool nodeConfigured();
+
+  // Retrieves loaded node configuration
+  std::shared_ptr<FullNodeConfig> getNodeConfiguration();
+
+  enum NetworkIdType { Mainnet = 1, Testnet, Devnet };
```
JakubFornadel, line 22: not sure if there are some specific limitations why enum is used here but in general enum class is preferred...

src/config/cli_config.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+#include <iostream>
+
+#include "cli_tools.hpp"
+
+using namespace std;
+
+namespace bpo = boost::program_options;
+
+namespace taraxa {
```
JakubFornadel, line 11: This comment is not related just to this specific line + its more like opened question rather than requirement. For me this seems like separate "cli" stuff so I would personally put all files into the src/cli folder and also created new nemaspace taraxa::cli. Classes could be then called without Cli prefix + I would also put  devnet/tetsnet_config.hpp into the taraxa::cli namespace

src/config/cli_config.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+CliConfig::CliConfig(int argc, const char* argv[], const std::string& taraxa_version)
+    : main_options("OPTIONS"),
+      node_command_options("NODE COMMAND OPTIONS"),
+      allowed_options("Allowed options"),
+      network_id(DEFAULT_NETWORK_ID) {
+  // Set node as default command
+  command.push_back(NODE_COMMAND);
```
JakubFornadel, line 19: might be cleaner to put all this processing from constructor into some separate method

docker-entrypoint.sh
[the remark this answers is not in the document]
sfrique: well, if it's working you can keep it, but I think it adds some confusion as the docker "default" db and config location will be different.

src/cli/devnet_config.hpp
```diff
@@ -0,0 +1,149 @@
+const char *devnet_json = R"foo({
```
JakubFornadel, line 1: one last minor thing:

- these configs should be also in taraxa::cli namespace
- #pragma once might be probably used to avoid possible duplicate declaration ?
- maybe it could be even constexpr - not sure about this one if it is possible due to usage ?

MatusKysel (approved):

rattrap (changes requested):
I think we need a few more flags for the `node` command so we can have all the functionality we have now:
```
--boot-node ip_address:port_number/node_id
--log channel:level
--node-secret secret
--vrf-secret secret
```

rattrap:
Another thing is that we should remove the `cli` option from the `docker-entrypoint.sh` script

```
  *)
    echo "You should choose between:"
    echo "cli, taraxad, single, join {NAMED_NETWOTK}"
    ;;
```

And fix the typo in `NAMED_NETWOTK`

rattrap:
> Another thing is that we should remove the `cli` option from the `docker-entrypoint.sh` script
> 
> ```
>   *)
>     echo "You should choose between:"
>     echo "cli, taraxad, single, join {NAMED_NETWOTK}"
>     ;;
> ```
> 
> And fix the typo in `NAMED_NETWOTK`

Fixed this

docker-entrypoint.sh
```diff
@@ -17,25 +17,19 @@ case $1 in
...
-	cli/taraxa config \
-        node \
-        --as-boot-node \
-        --file $TARAXA_CONF_PATH
-
-    echo "Starting taraxad..."
+	  echo "Starting taraxad..."
     taraxad \
-            --conf_taraxa $TARAXA_CONF_PATH
+            --config $TARAXA_CONF_PATH
```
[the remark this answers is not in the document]
rattrap: Done

docker-entrypoint.sh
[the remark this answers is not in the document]
rattrap: Done

rattrap:
```--data-dir arg          Data directory for the databases (default: "~/.taraxa/db")```

We should change this default data directory to `~/.taraxa/data`

There's already a `db` directory in that `data` dir

**Comment by rattrap** [comment]
```
  --command arg         Command arg:
                        node                            Runs the actual node (default)
                        account                       Generate new account private and public key
                        **account-from-key      Restore account from private key**
                        vrf                                Generate new VRF private and public key
                        **vrf-from-key arg         Restore VRF from private key**
```

I think we should simplify the cli by removing the `account-from-key` and `vrf-from-key` commands and having `--node-secret` and `--vrf-secret` flags on the `account` and `vrf` commands

So if you want to generate a new account you can do:
```
taraxad --command account
```
If you want to restore an account from a secret you can do:
```
taraxad --command account --node-secret THISISTHESECRET
```

Same for the `vrf` command.

**Comment by rattrap** [comment]
We have two use cases regarding how this new cli will be used:
1. Users starting a node locally
2. Automated scripts/deployments

For the first case, the users won't care about having options to overwrite the config file. They can just stop the node, modify the config file and start the node.

Automated scripts on the other hand, like our helm deployment, need to be able to generate the config file based on some parameters like logging channels and boot nodes.

**Comment by rattrap** [comment]
When we deploy "known" networks like the testnet or devnet we store the secret node secret and vrf secret somewhere safe and start the nodes with those.
For unknown or random nets we launch automatically we can just use the secrets generated by this cli.

The problem is we only have one script that does both things so it would be helpful if we had a new command here that's basically the same as the node command (same flags) but doesn't start the node after generating the config files.

```
--command config                  Generates the config files and exits
```

PR #847, opened 2021-06-08
fix(syncing): reduce sendBlocks traffic by skipping known blocks

justinsnapp:
## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -1049,7 +1049,7 @@ void TaraxaCapability::sendBlocks(NodeID const &_id, std::vector<std::shared_ptr
 
     // Split packet into multiple smaller ones if total size is > MAX_PACKET_SIZE
     if (packet_bytes.size() > MAX_PACKET_SIZE) {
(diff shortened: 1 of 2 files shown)


src/network/taraxa_capability.cpp
```diff
@@ -92,6 +92,12 @@ void TaraxaCapability::sealAndSend(NodeID const &nodeID, unsigned packet_type, R
     return;
   }
 
+  auto peer = getPeer(nodeID);
+
+  if (!peer->passed_initial_ && packet_type != StatusPacket) {
```
MatusKysel, line 97: I think this condition will break tests

MatusKysel: Maybe something like this https://github.com/Taraxa-project/taraxa-node/blob/develop/src/network/taraxa_capability.cpp#L213 could.  We should start to use something like `#ifdef TESTS` to distinguish between normal build and test build, but that's for the whole new task

MatusKysel (approved):

gaoqi8649 (approved):

PR #849, opened 2021-06-09
Issue 823/syncing optimization

JakubFornadel:
## Purpose

Syncing process refactor. I tried to keep the logic the same as before as we agreed to have a discussion before some major syncing refactor... 

Fixes:

- does not constantly restart syncing as before - now it restarts only if there is no ongoing syncing process or it is inactive for more than 60 seconds
- syncing_peer race condition
- set dag_syncing to false after the final dag blocks sync packet is received
- does not re-send already sent dag block
- not verifying dag blocks - only as tmp solution before major syncing refactor is done


mfrankovi (changes requested):
There is a network protocol version defined in full_node.hpp:
// Any time a change in the network protocol is introduced this version should be increased
  static constexpr uint16_t c_network_protocol_version = 4;

Since your changes are changing some of the packets format, you should increase this value

src/node/executor.cpp
```diff
@@ -69,6 +69,12 @@ void Executor::execute(std::shared_ptr<PbftBlock> blk) {
   cv_.notify_one();
 }
 
+void Executor::executeSynced(const std::shared_ptr<PbftBlock> &blk) {
+  assert(final_chain_->last_block_number() < blk->getPeriod());
```
gaoqi8649, line 73: The PBFT block is going to execute should be exactly in order.
I think should be assert(final_chain_->last_block_number() + 1 == blk->getPeriod())

justinsnapp: agree with @gaoqi8649

src/network/taraxa_capability.cpp
```diff
@@ -310,6 +312,11 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
                                    << ", peer PBFT previous round next votes size "
                                    << peer->pbft_previous_round_next_votes_size_;
 
+      // If we are still syncing and actual syncing process is active - do not trigger new syncing
```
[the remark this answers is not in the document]
justinsnapp: @gaoqi8649  perhaps don't need to trigger round syncing if we have pending PBFT blocks in our syncing queue, or believe we are behind by more than 1 PBFT block or the like?  What do you think?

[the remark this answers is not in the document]
gaoqi8649: @justinsnapp Yes, I agree

src/node/executor.cpp
JakubFornadel: so this one is invalid too right ? 
https://github.com/Taraxa-project/taraxa-node/blob/develop/src/node/executor.cpp#L61

gaoqi8649: No, that one is correct. 
Your new executeSynced is synchronous, that will execute PBFT block one by one in exact ordering
The old execute function is asynchronous, that will execute a bunch of PBFT blocks

gaoqi8649: I think after your changes, you are right. Since PBFT only will get multiple blocks from syncing. Each time only be able to vote one block.
So the PR will change back PBFT blocks execution in synchronous.
@justinsnapp I think we need to discuss on that if it's OK to change back on synchronous execution

gaoqi8649: @JakubFornadel @justinsnapp How about move synced PBFT blocks execution to taraxa capability thread?

justinsnapp:
@JakubFornadel , @MatusKysel  this PR can be closed now?  If so, go ahead. :-)

PR #853, opened 2021-06-10
chore(ci) fixes trigger for testnet branch with valid branch name

rattrap:
Fixes trigger for testnet branch with valid branch name.

Right now it's calling the circleci action with `refs/heads/testnet`
This change calls the action just with the branch name

### DIFF
--- a/.github/workflows/trigger-ci-on-branch.yml
+++ b/.github/workflows/trigger-ci-on-branch.yml
@@ -11,9 +11,12 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Checkout repository
+      uses: actions/checkout@v2
+    - name: Set env
+      run: echo "CIRCLE_BRANCH=$(git symbolic-ref -q --short HEAD)" >> $GITHUB_ENV
     - name: Trigger CircleCI build
       env:
-        CIRCLE_BRANCH: ${{ github.ref }}
         TOKEN: ${{ secrets.CIRCLE_TOKEN }}
       run: |
         curl -X POST \


.github/workflows/trigger-ci-on-branch.yml
```diff
@@ -11,9 +11,10 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
+    - name: Set env
+      run: echo "CIRCLE_BRANCH=$(git symbolic-ref -q --short HEAD)" >> $GITHUB_ENV
```
sfrique, line 15: I don't know if it makes any difference, but the official repo make the use a bit different https://docs.github.com/en/actions/reference/workflow-commands-for-github-actions#setting-an-environment-variable

both for how to set and how to use it.
like this:
```
steps:
  - name: Set the value
    id: step_one
    run: |
      echo "action_state=yellow" >> $GITHUB_ENV
  - name: Use the value
    id: step_two
    run: |
      echo "${{ env.action_state }}" # This will output 'yellow'
```

sfrique (changes requested):

.github/workflows/trigger-ci-on-branch.yml
rattrap: It works either way
`${CIRCLE_BRANCH}` === `${{ env.CIRCLE_BRANCH }}`

I prefer the current version

sfrique: The create env is also different, it as both name and value
`echo "action_state=yellow" >> $GITHUB_ENV`

And it uses `|` at run, but that I don't know if makes a difference.

If you are sure it works, I can approved it, I am just not familiarized with gh actions.

sfrique: I just read it again, and it should work, I missed one thing.

sfrique (approved):

JakubFornadel (approved):

PR #882, opened 2021-06-16
fix(build): match type consistent for Mac build

gaoqi8649:
## Version Notes ##
{ major | minor | patch }


src/final_chain/final_chain.cpp
```diff
@@ -117,7 +117,8 @@ class FinalChainImpl final : public FinalChain {
         }
       }
     }
-    transaction_count_hint_ = max(transaction_count_hint_, to_execute.size());
+    uint32_t to_execute_trxs_size = to_execute.size();
```
JakubFornadel, line 120: why to use this type conversion from size_t to uint32_t when you can just declare transaction_count_hint_ as size_t and keep this part of code as it was ?

PR #883, opened 2021-06-16
chore(unit test): refactor and fix network unit test transfer_transaction

gaoqi8649:
## Purpose

Need wait nodes connect to each other. Otherwise will fail sending packet:
https://github.com/Taraxa-project/taraxa-node/blob/develop/src/network/taraxa_capability.cpp#L97 

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/tests/network_test.cpp
+++ b/tests/network_test.cpp
@@ -136,34 +136,30 @@ TEST_F(NetworkTest, send_pbft_block) {
                  [&](auto& ctx) { WAIT_EXPECT_EQ(ctx, nw1->getPeer(node2_id)->pbft_chain_size_, chain_size) });
 }
 
-// Test creates two Network setup and verifies sending transaction
-// between is successfull
+// Test creates two Network setup and verifies sending transaction between is successfull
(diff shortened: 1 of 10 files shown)


tests/network_test.cpp
```diff
@@ -136,34 +136,30 @@ TEST_F(NetworkTest, send_pbft_block) {
...
 
-// Test creates two Network setup and verifies sending transaction
-// between is successfull
+// Test creates two Network setup and verifies sending transaction between is successfull
 TEST_F(NetworkTest, transfer_transaction) {
-  std::unique_ptr<Network> nw1(new taraxa::Network(g_conf1->network));
-  std::unique_ptr<Network> nw2(new taraxa::Network(g_conf2->network));
-
+  auto nw1 = u_ptr(new Network(g_conf1->network));
+  auto nw2 = u_ptr(new Network(g_conf2->network));
```
JakubFornadel, line 142: when you refactor old code, please lets do not copy paste things like, use make_unique or make_shared instead of new... and get rid of u_ptr too...

PR #887, opened 2021-06-17
Create PR-NETs from received triggers from Github

agrebin:
## Purpose

This pr is to atually add the helm commands required to react to the gh actions received by /deploy, /redeploy and /cleanup 

## For reviewers

please use this pull request to test the actual actions to be processed at circleci 

## Version Notes ##
{ major | minor | patch }


agrebin:
Deployed prnet successfully

agrebin:
ReDeployed prnet successfully

agrebin:
prnet cleaned-up successfully

.circleci/config.yml
```diff
@@ -5,18 +5,363 @@ parameters:
...
+            do
+               GETALL=$(kubectl get all -n blockchain-prnet | grep pr-"<< parameters.pr-name >>") || true
+               if [ -z "${GETALL}" ] ; then
+                 exit 0
+               else
+                echo "waiting.. for all resources to be deleted"
+                echo ${GETALL}
+                sleep 10
+              fi             
+            done
```
sfrique, line 267: Shouldn't we add a log msg saying not everything was clean up?
Not sure what difference does it make...

sfrique (approved):
LGTM

.circleci/config.yml
agrebin: what do you mean...
change the message? instead of 'waiting for all resources to be deleted' , to say 'not everything was cleaned up', or when you say 'log' you mean to log a message to the pr?
in the second case i think it will flood, i dont like that
and regarding the message to echo in the pipeline, i think it is fine.. so i am lost

sfrique: it just an echo so we can know if we exit because everything was clean up or if we timeout.

agrebin: added it, check if it works fine.

rattrap (approved):

PR #891, opened 2021-06-18
perf(pbft): Remove duplicate PBFT synced blocks verification and getDagBlockOrder

gaoqi8649:
## Version Notes ##
{ major | minor | patch }


justinsnapp (approved):

JakubFornadel:
Not sure if this PR is supposed to optimize the stuff we were talking about on Thursday call, so the rest of the comment might not be relevant.

All this PR does is that now we dont compare parent hash of current synced pbft block twice, but only once. I mean yeah it does not need to be checked twice but it will have zero impact on performance. The real optimization we were talking about on Thursday call is that "getDagBlockOrder" is called twice, not this basic check...

src/consensus/pbft_manager.cpp
```diff
@@ -1332,31 +1333,22 @@ void PbftManager::pushSyncedPbftBlocksIntoChain_() {
...
-      pbft_last_observed_synced_queue_size_ = pbft_synced_queue_size;
-      break;
-    }
+
     if (!comparePbftBlockScheduleWithDAGblocks_(*pbft_block_and_votes.pbft_blk)) {
+      LOG(log_nf_) << "DAG has not build up for anchor " << pbft_block_and_votes.pbft_blk->getPivotDagBlockHash()
+                   << " in PBFT block " << pbft_block_hash;
       break;
     }
     if (pushPbftBlock_(pbft_block_and_votes)) {
```
JakubFornadel, line 1348: Here is the real perf problem. We call getDagBlockOrder inside "pushPbftBlock_" as well as in  "comparePbftBlockScheduleWithDAGblocks_", which is called few lines above...

JakubFornadel (changes requested):
This commit does not fix anything, does it ? It could be perf as you optimized it so it does not compare that 1 pbft parent hash twice but still this wont affect performance in any manner so chore might be more appropriate... But I could live with perf too :) And dont use capital letters if possible - these commits will be automatically processed and changelog will be generated out of them. Thats why it is important to think about the proper type and write it in a form that was defined

gaoqi8649:
> Not sure if this PR is supposed to optimize the stuff we were talking about on Thursday call, so the rest of the comment might not be relevant.
> 
> All this PR does is that now we dont compare parent hash of current synced pbft block twice, but only once. I mean yeah it does not need to be checked twice but it will have zero impact on performance. The real optimization we were talking about on Thursday call is that "getDagBlockOrder" is called twice, not this basic check...

Could you please show me where call the "getDagBlockOrder" twice? I didn't find that

src/consensus/pbft_manager.cpp
gaoqi8649: Thanks, fixed

gaoqi8649:
> This commit does not fix anything, does it ? It could be perf as you optimized it so it does not compare that 1 pbft parent hash twice but still this wont affect performance in any manner so chore might be more appropriate... But I could live with perf too :) And dont use capital letters if possible - these commits will be automatically processed and changelog will be generated out of them. Thats why it is important to think about the proper type and write it in a form that was defined

Thanks! Fixed

JakubFornadel (approved):

PR #896, opened 2021-06-18
removed rocket reaction to fix problem with credentials

agrebin:
## Purpose

Github actions reading issue comments were failing on this organization this is to fix it
## For reviewers

i tested on my fork and this was not happening. after reading the problem looks to be with the GH token, and it is only required to have the 'rocket' reaction, i think it is non important so i am removing it, please approve and merge to develop so we can continue the work, thanks

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -13,9 +13,6 @@ jobs:
         id: check
         with:
           trigger: '/deploy'
-          reaction: rocket
-        env:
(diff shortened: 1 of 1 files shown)


rattrap (approved):

JakubFornadel:
@agrebin How is it possible that this was merged with 1 approval ? From technival point of view it should not be possible unless you are admin ?

PR #905, opened 2021-06-19
build(conan): change bincrafters url; update building doc

kstdl:
## Purpose

Set new working url for bincrafters conan repo

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -273,7 +273,7 @@ jobs:
             - conan-cache-v2-mac
         - run:
            name: add conan bincrafters remote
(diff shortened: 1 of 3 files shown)


agrebin:
So far looks good to me, but i think we need to test it by creating a release out of the commit and see if the binary releases work

justinsnapp (approved):

JakubFornadel (approved):

JakubFornadel:
> So far looks good to me, but i think we need to test it by creating a release out of the commit and see if the binary releases work

hmm why do you think the binary should not work If it passed all tests ?

PR #906, opened 2021-06-21
implemented external caching with cci-x/docker-registry-image-cache

agrebin:
## Purpose

This PR is to speedup docker image builds via implementing external cache method.

## For reviewers

i have performed may tests, trying to cover most of the variants that i was able to reproduce.
it costs about 10' downloading two cache images, one for current branch and one to be the latest, but overall i think it lowers the built time at least to half.
i have discovered that cacheable layers are all on the actual ctest image, since the taraxad image gets most of the stuff removed, so i retag ctest image to be the caching image.

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -68,6 +68,7 @@ jobs:
          repository: gcr.io/${GOOGLE_PROJECT_ID}
          images: >-
            ${IMAGE}-ctest:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}
+           ${IMAGE}-cache:latest
            ${IMAGE}-cache:${DOCKER_BRANCH_TAG}
          steps: 
            - docker-cache/build:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -54,7 +54,6 @@ jobs:
             echo ${GCLOUD_SERVICE_KEY} | docker login -u _json_key --password-stdin https://gcr.io
 
       - docker-cache/with-save-restore-images:
-         name: Build main image with external cache
          repository: gcr.io/${GOOGLE_PROJECT_ID}
          images: >-
            ${IMAGE}-cache:${DOCKER_BRANCH_TAG}
@@ -67,7 +66,6 @@ jobs:
                 docker build -t ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} .
 
       - docker-cache/with-save-restore-images:
-         name: Build cache image with eternal caching
          repository: gcr.io/${GOOGLE_PROJECT_ID}
          images: >-
            ${IMAGE}-ctest:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,14 +12,15 @@ parameters:
 orbs:
   git: opuscapita/git@0.0.3
   github-release: izumin5210/github-release@0.1.1
-
   gcp-gcr: circleci/gcp-gcr@0.13.0
+  docker-cache: cci-x/docker-registry-image-cache@0.2.12
 
 
 jobs:
    build-and-push-docker-image:
     environment:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -43,19 +44,52 @@ jobs:
...
+#         name: Docker Build taraxa-node
+#         command: |
+#           docker build -t ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} .
+
+      - docker-cache/with-save-restore-images:
+         repository: gcr.io/${GOOGLE_PROJECT_ID}
+         images: >-
+           ${IMAGE}:${DOCKER_BRANCH_TAG}
+           ${IMAGE}-cache:${DOCKER_BRANCH_TAG}
+           ${IMAGE}-cache:latest
```
rattrap, line 66: Why do we need this cache image ?

.circleci/config.yml
```diff
@@ -43,19 +44,52 @@ jobs:
...
-      - run:
-         name: Docker Build taraxa-node-ctest
+         name: Login into gcr
          command: |
-          docker build -t ${IMAGE}-ctest:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} \
-            --target build .
+            echo ${GCLOUD_SERVICE_KEY} | docker login -u _json_key --password-stdin https://gcr.io
+ 
+#      - run:
+#         name: Docker Build taraxa-node
```
rattrap, line 57: We shouldn't keep commented code here.

.circleci/config.yml
```diff
@@ -43,19 +44,52 @@ jobs:
...
+           - docker-cache/build:
+              command: |
+                docker build -t ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} .
+                 
+#      - run:
+#         name: Docker Build taraxa-node-ctest\
+#         command: |
+#          docker build -t ${IMAGE}-ctest:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} \
+#            --target build .
+      - docker-cache/with-save-restore-images:
```
rattrap, line 78: We should name these actions accordingly. The default name is "_DRIC: Restore cache from registry gcr.io/${GOOGLE_PROJECT_ID}_"

rattrap (changes requested):

.circleci/config.yml
sfrique: If we only save the actual cache/ctest image there, I think it's a good idea, so we can from time to time, clean up the registry and save some storage costs.

.circleci/config.yml
agrebin: hmm not sure if i can, i can name all the steps i explicitly declare , but i think the once you see with 'DRIC' are internal steps of the orb's command.

agrebin: As i expected the name is not supported on all commands check run 1447

agrebin: Tested adding name properties, but i had to roll them back , tthey're not supported.

.circleci/config.yml
agrebin: the Removed ${IMAGE}:${DOCKER_BRANCH_TAG} but the other two are required, we may choose not to use, the latest, but as we wred discussing, the latest is inteneded for the 'first commit in pr' while the one with the branch should be useful for the next runs, they should share more layers but when downloading they differ more than i expected.

rattrap:
You can also remove the docker login from the "Tag images" step.

JakubFornadel (changes requested):
Please read this doc and reword your commits according to the style we are sing, thank you. Just PM me if have any questions.

agrebin:
> Please read this doc and reword your commits according to the style we are using, thank you. Just PM me if have any questions.
> https://github.com/Taraxa-project/taraxa-node/blob/develop/doc/git_practices.md

Sure jakub i was completely unaware of it.

rattrap (approved):

sfrique (approved):

PR #909, opened 2021-06-22
Splitting PbftBlockPacket if over max size of 15MB

mfrankovi:
## Purpose

Splitting PbftBlockPacket if over max size of 15MB

## For reviewers

PbftBlockPacket which can contain multiple pbft packets together with the dag blocks and transactions was previously only split by pbft packets meaning that any pbft block packet over 15MB could not be split. With the changes in this PR splitting is done on the dag block level. This still means that no DAG block over 15MB can be sent.

## Changes

Changes in taraxa capability on sending and processing PBFT packet. Format is changed for any packet with pbft and dag blocks changing order in the packet. Pbft block is only sent once all dag blocks are sent

## Tests

Unit test added which tests syncing of large pbft block

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -1528,9 +1528,15 @@ void TaraxaCapability::sendPbftBlocks(NodeID const &_id, size_t height_to_sync, 
         dag_rlp.appendRaw(transactions[trx_idx]);
       }
 
+      // When checking if size limit exceeds MAX_PACKET_SIZE there are few bytes or rlp structure that is added
+      // for the pbft block and dag block. This should be just a couple of bytes but we enforece even stricter 128 limit
+      static const int RLP_OVERHEAD = 128;
+
       // Check if PBFT blocks need to be split and sent in multiple packets so we dont exceed
       // MAX_PACKET_SIZE (15 MB) limit
-      if (act_packet_size + dag_blocks_size + dag_rlp.out().size() > MAX_PACKET_SIZE) {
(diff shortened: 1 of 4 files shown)


src/network/taraxa_capability.cpp
```diff
@@ -1511,48 +1510,61 @@ void TaraxaCapability::sendPbftBlocks(NodeID const &_id, size_t height_to_sync,
...
+
+      dag_blocks_rlps.push_back(dag_rlp.out());
+      dag_blocks_size += dag_rlp.out().size();
     }
+    pbft_block_rlp.appendList(2);  // item #1 - list of dag blocks, item #2 pbft block rlp
+    pbft_block_rlp.appendList(dag_blocks_rlps.size());
+
+    for (auto &dag_block : dag_blocks_rlps) pbft_block_rlp.appendRaw(dag_block);
+
+    pbft_block_rlp.appendRaw(pbft_blocks[pbft_block_idx].rlp());
```
JakubFornadel, line 1567: are you sure this is ok ? Seems to me that this pbft block is added to the end of dga blocks array. 
 
Isn't this the correct sequence  ? :

  pbft_block_rlp.appendList(2);  // item #1 - list of dag blocks, item #2 pbft block rlp
  pbft_block_rlp.appendRaw(pbft_blocks[pbft_block_idx].rlp());
  
  pbft_block_rlp.appendList(dag_blocks_rlps.size());
  for (auto &dag_block : dag_blocks_rlps) pbft_block_rlp.appendRaw(dag_block);

justinsnapp (approved):
Testest in testnet and it worked well!

JakubFornadel (approved):

PR #912, opened 2021-06-23
feat: add additional thread for processing GetPbftBlockPacket

justinsnapp:
## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -577,6 +577,7 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
       }
       // If blocks_to_transfer is 0, send peer empty PBFT blocks for talking to peer syncing has completed
       syncing_tp_.post([host, _nodeID, height_to_sync, blocks_to_transfer, this] {
+        bool send_blocks = true;
         if (blocks_to_transfer > 0) {
           uint32_t total_wait_time = 0;
           while (tp_.num_pending_tasks() > MAX_NETWORK_QUEUE_TO_DROP_SYNCING) {
@@ -587,11 +588,12 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
               LOG(log_er_pbft_sync_) << "Node is busy with " << tp_.num_pending_tasks() << " network packets. Host "
                                      << _nodeID.abridged() << " will be disconnected";
               host->disconnect(_nodeID, p2p::UserReason);
+              send_blocks = false;
               break;
             }
           }
         }
-        sendPbftBlocks(_nodeID, height_to_sync, blocks_to_transfer);
+        if (send_blocks) sendPbftBlocks(_nodeID, height_to_sync, blocks_to_transfer);
       });
       break;
     }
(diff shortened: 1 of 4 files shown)


JakubFornadel:
Just a minor note, none of these these commits are feat. Think about it from end user perspective. He does not care if we serve sync blocks from threadpool or not. Feat is new feature - for example implementation of fees, possibility to download sync snapshot  would be a new feature. These changes are either chore, refactor or fix...

src/network/taraxa_capability.cpp
```diff
@@ -564,10 +564,33 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
         blocks_to_transfer =
             std::min((uint64_t)conf_.network_sync_level_size, (uint64_t)(my_chain_size - (height_to_sync - 1)));
       }
-
       LOG(log_dg_pbft_sync_) << "Will send " << blocks_to_transfer << " PBFT blocks to " << _nodeID;
+
+      // Stop syncing if too many sync requests or if the rest of the node is busy
```
JakubFornadel, line 569: Hm not really sure if I am okey with this fix. I believe this might bring us more trouble than improvements:

-  There is no mechanism for a new node to know if the node that he wants to sync with is "busy" in terms of serving syncing data. Real solution is choosing different node based on some data(maybe from status packet ?) + why should the node be disconnected here ? 
- It is not limitation of how many nodes can be syncing with this node. Its a limitation of how many syncing requests are scheduled... In some edge case it could disconnect even the single node that is syncing with this one
- as serving of syncing data is in separate thread now what performance-related problems should other parts of code have even if syncing thread is working on 100% ? There are literally 3 get calls to db so even locks should not be any problem...

mfrankovi: - I agree that the disconnect is kind of a dirty way of telling node that it should not sync with it. We already have this mechanism if a node we are syncing to disconnects that we switch syncing to another node for syncing so I kind of reused that. Although currently a node that is syncing really does not have any other role on the network other then trying to get synced so a disconnect to me actually makes sense. No need for it to have too many connections when syncing. There is an auto reconnect mechanism in p2p so once it is synced up it will reconnect to optimal number of nodes.
- Syncing requests are synchronous. One node will not send two getPbftBlockPackets in parallel, it will only send a second one once it gets a response to a first one. Any parallel getPbftBlockPacket packets being processed at the same time will match the number of nodes trying to sync. Excluding the possibility of a malicious node
- These 3 get calls are multiget query which can take a long time considering the amount of data they might be actually requesting(large number of dag blocks and transactions). If other parts of the code are also reading/writing to the db at the same time internal db processing/locking as well as any CPU usage in doing this

JakubFornadel: - hmm yes very dirty + can be applied only if your point #2 is true, which I believe is not...
- syncing requests are anything but synchronous. Yes they are synchronous in taraxa_capability but PBFT manager, which is running in separate thread is constantly restarting syncing with force flag - that will resend syncing request no matter the current syncing state...
- these are read locks, so its not that bad.  But yeah it would be definitely worth improving perf there, I just dont think that this should cause some real troubles, but I might be wrong with this point :)

PR #917, opened 2021-06-25
fix: network test large pbft block

mfrankovi:
## Purpose

sync_large_pbft_block was failing due to timing issues on creating pbft block. Fixed by using pbftmanager stop and start method to schedule creation of pbft block

## Version Notes ##
{ major | minor | patch }


justinsnapp (commented):
I think making a change to the GetBlocksPacket protocol should be its own separate pull request, and definitely such a change should warrant an increment of the network protocol version.

PR #920, opened 2021-06-29
chore(tests): check for nullptr in sync_large_pbft_block test

MatusKysel:
## Purpose

network_test core dumped for me locally


JakubFornadel (approved):

mfrankovi (approved):
I guess this is ok for a temporary fix but this means that we have some other bug because it should not be possible to not have dag block that is a part of the pbft block that was added to the pbft chain. Maybe @gaoqi8649 could take a look

MatusKysel:
> I guess this is ok for a temporary fix but this means that we have some other bug because it should not be possible to not have dag block that is a part of the pbft block that was added to the pbft chain. Maybe @gaoqi8649 could take a look

Test will FAIL so we are not hiding the error, but now rest of the testsuite will continue to run

rattrap:
https://app.circleci.com/pipelines/github/Taraxa-project/taraxa-node/1543/workflows/9eff8881-650c-4e8c-b435-747eb20cd097/jobs/1601

This fails on develop. Did you guys merge it without waiting for the tests?

PR #922, opened 2021-06-29
optimize syncing non finalized dag blocks

mfrankovi:
## Purpose

optimize syncing non finalized dag blocks

## For reviewers

Sending hashes with the getBlocksPacket to only retrieve non finalized blocks not yet known by the requesting node

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -401,7 +401,7 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
     }
     case GetBlocksPacket: {
       LOG(log_dg_dag_sync_) << "Received GetBlocksPacket with " << _r.itemCount() << " known blocks";
-      set<blk_hash_t> known_non_finalized_blocks;
+      std::unordered_set<blk_hash_t> known_non_finalized_blocks;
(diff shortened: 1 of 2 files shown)


justinsnapp (approved):

MatusKysel (approved):

src/network/taraxa_capability.cpp
```diff
@@ -400,12 +400,19 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
       break;
     }
     case GetBlocksPacket: {
-      LOG(log_dg_dag_sync_) << "Received GetBlocksPacket";
+      LOG(log_dg_dag_sync_) << "Received GetBlocksPacket with " << _r.itemCount() << " known blocks";
+      set<blk_hash_t> known_non_finalized_blocks;
```
MatusKysel, line 404: maybe use `std::unordered_map` so insert and search is O(1)

JakubFornadel: Why map ? as for insert& search complexity fix, unordered_set  should do it

justinsnapp: @JakubFornadel  changed it to unordered set.

PR #925, opened 2021-06-29
chore(pbft): implement exponential backoff for advanced steps in round beyond MAX_STEPS

justinsnapp:
## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -201,7 +201,7 @@ void PbftManager::update_dpos_state_() {
                    << pbft_chain_->getPbftChainSize() << ", have executed chain size "
                    << final_chain_->last_block_number();
       // Sleep one PBFT lambda time
-      thisThreadSleepForMilliSeconds(LAMBDA_ms);
+      thisThreadSleepForMilliSeconds(POLLING_INTERVAL_ms);
     }
   } while (!stopped_);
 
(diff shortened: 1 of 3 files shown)


justinsnapp:
@gaoqi8649  I was really surprised to see we still had a STEP_4_DELAY in there, no reason for that to exist now that execution is asynchronous to pbft.

src/consensus/pbft_manager.cpp
```diff
@@ -239,16 +239,20 @@ void PbftManager::setPbftStep(size_t const pbft_step) {
...
-  //   LAMBDA_ms = LAMBDA_ms_MIN;
-  // }
+  if (step_ > MAX_STEPS && LAMBDA_backoff_multiple < 8) {
+    // Note: We calculate the lambda for a step independently of prior steps
+    //       in case missed earlier steps.
+    std::default_random_engine generator;
+    std::uniform_int_distribution<u_long> distribution(0, step_ - MAX_STEPS);
+    auto lambda_random_count = distribution(generator);
+    LAMBDA_backoff_multiple = 2 * LAMBDA_backoff_multiple;
+    LAMBDA_ms = LAMBDA_ms_MIN * (LAMBDA_backoff_multiple + lambda_random_count);
```
gaoqi8649, line 249: Every time distribution will generate a number between [0, step_ - MAX_STEPS]. Looks like that cannot guaranty all nodes have same LAMBDA_ms in same steps.
Sorry, I don't understand the algorithm here

justinsnapp: Don't think it truly matters that nodes all have the same LAMBDA_ms once they get into the next voting steps, once they reset to next round they will al resume same lambda for the critical propose, soft vote and cert vote steps.   

The concept for adding jitter to binary exponential backoff algorithm (as I understand it) is to spread out the traffic so as to reduce the spikes in traffic in a network constrained situation in a distributed uncoordinate system. (see: https://devopedia.org/images/article/95/9334.1547359774.jpg )

gaoqi8649 (approved):
If don't require all nodes have same LAMBDA in same steps, that looks good to me

src/consensus/pbft_manager.cpp
```diff
@@ -239,16 +239,20 @@ void PbftManager::setPbftStep(size_t const pbft_step) {
...
-  //   // LOG(log_nf_) << "Surpassed max steps, relaxing lambda to " << LAMBDA_ms
-  //   //             << " ms in round " << getPbftRound() << ", step " << step_;
-  //   LAMBDA_ms = LAMBDA_ms_MIN;
-  // } else {
-  //   LAMBDA_ms = LAMBDA_ms_MIN;
-  // }
+  if (step_ > MAX_STEPS && LAMBDA_backoff_multiple < 8) {
+    // Note: We calculate the lambda for a step independently of prior steps
+    //       in case missed earlier steps.
+    std::default_random_engine generator;
```
MatusKysel, line 245: this needs to be global or class variable or you need to use some kind of seed (time?) every time you initialize it otherwise will get same number (probably 0) all the time from distribution(0, step_ - MAX_STEPS);

justinsnapp: @MatusKysel  ok I updated it as you wisely suggested, please re-review.

MatusKysel (approved):

PR #927, opened 2021-06-29
perf(votes): working in progress on return verified votes in reference

gaoqi8649:
## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -876,10 +876,10 @@ uint64_t PbftManager::roundDeterminedFromVotes_() {
   auto round = getPbftRound();
 
   for (auto const &v : votes_) {
-    if (v.getType() != next_vote_type) {
+    if (v.get().getType() != next_vote_type) {
       continue;
     }
-    std::pair<uint64_t, size_t> round_step = std::make_pair(v.getRound(), v.getStep());
+    std::pair<uint64_t, size_t> round_step = std::make_pair(v.get().getRound(), v.get().getStep());
     if (round_step.first >= round) {
       if (next_votes_tally_by_round_step.find(round_step) != next_votes_tally_by_round_step.end()) {
         next_votes_tally_by_round_step[round_step] += 1;
@@ -966,10 +966,17 @@ votesBundle PbftManager::blockWithEnoughVotes_(std::vector<Vote> const &votes) c
 }
 
 std::vector<Vote> PbftManager::getVotesOfTypeFromVotesForRoundAndStep_(PbftVoteTypes vote_type,
-                                                                       std::vector<Vote> &votes, uint64_t round,
+                                                                       std::vector<std::reference_wrapper<Vote>> &votes, uint64_t round,
                                                                        size_t step,
                                                                        std::pair<blk_hash_t, bool> blockhash) {
   std::vector<Vote> votes_of_requested_type;
+  // for (auto const& v : votes) {
+  //   if (v.getType() == vote_type && v.getRound() == round && v.getStep() == step &&
+  //       (blockhash.second == false || blockhash.first == v.getBlockHash())) {
+  //     votes_of_requested_type.emplace_back(v);
+  //   }
+  // }
+
   std::copy_if(votes.begin(), votes.end(), std::back_inserter(votes_of_requested_type),
                [vote_type, round, step, blockhash](Vote const &v) {
                  return (v.getType() == vote_type && v.getRound() == round && v.getStep() == step &&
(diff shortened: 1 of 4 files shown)


JakubFornadel:
While I was checking if we really can use references here due to possibility to have dangling references, I noticed that PbftManager class member "executed_pbft_block_" is basically useless, can we refactor it too ? See my comments for hints why it is not needed...

JakubFornadel (changes requested):
While I was checking if we really can use references here due to possibility to have dangling references, I noticed that PbftManager class member "executed_pbft_block_" is basically not needed, can we refactor it too ?

We are checking it [here](https://github.com/Taraxa-project/taraxa-node/blob/develop/src/consensus/pbft_manager.cpp#L300)  in "resetRound_()" but this condition will never be true  as "executed_pbft_block_" is set to true only if [this function call](https://github.com/Taraxa-project/taraxa-node/blob/develop/src/consensus/pbft_manager.cpp#L557) returns true, in which case "resetRound_()" is never called.

All other usages look like [this](https://github.com/Taraxa-project/taraxa-node/blob/develop/src/consensus/pbft_manager.cpp#L1362) (see usage few lines below in link). This code can be refactored to:

`
if (pushPbftBlock_(pbft_block_and_votes, dag_blocks_order.first)) {
      LOG(log_nf_) << node_addr_ << " push synced PBFT block " << pbft_block_hash << " in round " << round;
    } else {
      LOG(log_er_) << "Failed push PBFT block " << pbft_block_hash << " into chain";
      break;
    }

    // Remove from PBFT synced queue
    pbft_chain_->pbftSyncedQueuePopFront();
    if (executed_pbft_block_) {
      vote_mgr_->removeVerifiedVotes();
      update_dpos_state_();
      // update sortition_threshold and TWO_T_PLUS_ONE
      updateTwoTPlusOneAndThreshold_();
      db_->savePbftMgrStatus(PbftMgrStatus::executed_block, false);
      executed_pbft_block_ = false;
    }
    pbft_synced_queue_size = pbft_chain_->pbftSyncedQueueSize();
    if (pbft_last_observed_synced_queue_size_ != pbft_synced_queue_size) {
      LOG(log_dg_) << "PBFT synced queue still contains " << pbft_synced_queue_size
                   << " synced blocks that could not be pushed.";
    }
    pbft_last_observed_synced_queue_size_ = pbft_synced_queue_size;
`

gaoqi8649:
> While I was checking if we really can use references here due to possibility to have dangling references, I noticed that PbftManager class member "executed_pbft_block_" is basically not needed, can we refactor it too ?
> 
> We are checking it [here](https://github.com/Taraxa-project/taraxa-node/blob/develop/src/consensus/pbft_manager.cpp#L300) in "resetRound_()" but this condition will never be true as "executed_pbft_block_" is set to true only if [this function call](https://github.com/Taraxa-project/taraxa-node/blob/develop/src/consensus/pbft_manager.cpp#L557) returns true, in which case "resetRound_()" is never called.
> 
> All other usages look like [this](https://github.com/Taraxa-project/taraxa-node/blob/develop/src/consensus/pbft_manager.cpp#L1362) (see usage few lines below in link). This code can be refactored to:
> 
> ```
> if (pushPbftBlock_(pbft_block_and_votes, dag_blocks_order.first) == false) {
>   LOG(log_er_) << "Failed push PBFT block " << pbft_block_hash << " into chain";
>   break;
> }
> 
> LOG(log_nf_) << node_addr_ << " push synced PBFT block " << pbft_block_hash << " in round " << round;
> 
> // Remove from PBFT synced queue
> pbft_chain_->pbftSyncedQueuePopFront();
> vote_mgr_->removeVerifiedVotes();
> update_dpos_state_();
> // update sortition_threshold and TWO_T_PLUS_ONE
> updateTwoTPlusOneAndThreshold_();
> db_->savePbftMgrStatus(PbftMgrStatus::executed_block, false);
> executed_pbft_block_ = false;
> pbft_synced_queue_size = pbft_chain_->pbftSyncedQueueSize();
> 
> if (pbft_last_observed_synced_queue_size_ != pbft_synced_queue_size) {
>       LOG(log_dg_) << "PBFT synced queue still contains " << pbft_synced_queue_size
>                    << " synced blocks that could not be pushed.";
> }
> pbft_last_observed_synced_queue_size_ = pbft_synced_queue_size;
> ```

The reason is PBFT need to keep same DPOS state (same 2t+1 and threshold) in same round, we hit bugs before.
So cannot update in the middle of round https://github.com/Taraxa-project/taraxa-node/blob/develop/src/consensus/pbft_manager.cpp#L557
Need to update in the beginning of a new round https://github.com/Taraxa-project/taraxa-node/blob/develop/src/consensus/pbft_manager.cpp#L300

PR #940, opened 2021-07-02
chore(logs): use correct path for logs

MatusKysel:
## Purpose

new directory layout

```
.
+-- config.json
+-- wallet.json
+-- data/
|   +-- db/
|   +-- logs/
|   +-- net/
```

### DIFF
--- a/src/config/config.cpp
+++ b/src/config/config.cpp
@@ -90,7 +90,8 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value 
     json_file_name = string_or_object.asString();
   }
   auto const &root = string_or_object.isString() ? parsed_from_file : string_or_object;
-  db_path = getConfigDataAsString(root, {"db_path"});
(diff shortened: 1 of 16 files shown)


src/config/config.cpp
```diff
@@ -163,7 +163,7 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value
     if (auto path = getConfigData(root["logging"], {"log_path"}, true); !path.isNull()) {
       log_path = path.asString();
     } else {
-      log_path = db_path / "logs";
+      log_path = db_path / "../logs";
```
kstdl, line 166: Maybe we should have something like `datadir_path` filed in that config?

`auto net_file_path() const { return db_path / "net"; }`
There is that method in a header, so looks like the same problem with net_path

MatusKysel: agree

JakubFornadel (approved):

kstdl (approved):

PR #941, opened 2021-07-02
fix: ensure status message is sent first

mfrankovi:
## Purpose

Transactions that are sent on short intervals might get sent to the node before the status message. This will cause the node to ignore the packet and transactions but at the same time transactions would incorrectly be marked as known. The fix ensures status message is sent before sending the transactions

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -76,9 +76,11 @@ void TaraxaCapability::erasePeer(NodeID const &node_id) {
   peers_.erase(node_id);
 }
 
-void TaraxaCapability::insertPeer(NodeID const &node_id) {
+std::shared_ptr<TaraxaPeer> TaraxaCapability::insertPeer(NodeID const &node_id) {
   boost::unique_lock<boost::shared_mutex> lock(peers_mutex_);
-  peers_.emplace(std::make_pair(node_id, std::make_shared<TaraxaPeer>(node_id)));
+  auto ret = peers_.emplace(std::make_pair(node_id, std::make_shared<TaraxaPeer>(node_id)));
+  assert(ret.second);
+  return ret.first->second;
 }
 
 void TaraxaCapability::syncPeerPbft(NodeID const &_nodeID, unsigned long height_to_sync) {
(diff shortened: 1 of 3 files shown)


justinsnapp (approved):

src/network/taraxa_capability.cpp
```diff
@@ -94,7 +96,7 @@ void TaraxaCapability::sealAndSend(NodeID const &nodeID, unsigned packet_type, R
 
   auto peer = getPeer(nodeID);
 
-  if ((!peer) || (dag_mgr_ && !peer->passed_initial_ && packet_type != StatusPacket)) {
+  if ((!peer) || (dag_mgr_ && !peer->received_initial_status_ && packet_type != StatusPacket)) {
```
JakubFornadel, line 99: why is in this condition dag_mgr ? I would also add here sent_initial_status_ so the node does not send any other packet as it is not processed on the other side until it receives initial status packet too anyway. This condition seems more reasonable to me:
`if (!peer || ((!peer->sent_initial_status_ && !peer->received_initial_status_) && packet_type != StatusPacket)) {`

JakubFornadel:
Seems ok except the one note I had. But I wouldn't merge as there is actually pretty big chance that we are marking also something else as known (blocks, ...) and try to send it before initialization through status packets is done. In such case this packet is not send but we might mark some data as known. So it would be nice if we check all possible places and not just fix 1 problem that we know of but there might be actually more problems of the same character

src/network/taraxa_capability.cpp
justinsnapp: @JakubFornadel  dag_mgr is checked here to check if we have it initialized because in network tests we often don't have dag mgr initialized as part of the test.

JakubFornadel: hmm but that should not be handled here I guess. In general we have a lot of checks in code just because of tests, which should not be there or at least it should be somehow checked only if we are running tests through some #ifdef TESTS or something like that. Could be separate task though

src/network/taraxa_capability.hpp
```diff
@@ -86,7 +86,8 @@ class TaraxaPeer : public boost::noncopyable {
 
   void setAlive() { alive_check_count_ = 0; }
 
-  bool passed_initial_ = false;
+  bool received_initial_status_ = false;
+  bool sent_initial_status_ = false;
```
gaoqi8649, line 90: Looks like don't need sent_initial_status_. I think when node received_initial_status_ = true, that also means peer has send initial status.
I may wrong.

src/network/taraxa_capability.cpp
MatusKysel: We should add those guards there, there are lot of places we have test code

MatusKysel (approved):

PR #945, opened 2021-07-03
[REPLACED BY PR 965] fix(pbft): refactor the way we opt to give up on non-received next voted pbft block

justinsnapp:
## Purpose

Currently we had a fatal flaw in doing this based on round.   But in certain circumstances we could violate ability to make consensus for next voting if partitioning of the count of rounds have been waiting occurred.   Now we do it based on time, so that after some point in time > 2t+1 nodes will have timedout. 

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -379,11 +379,14 @@ void PbftManager::initialState_() {
     }
     previous_round_next_votes_->updateNextVotes(next_votes_in_previous_round, previous_round_2t_plus_1);
   }
+
+  next_voted_value_ = previous_round_next_votes_->getVotedValue();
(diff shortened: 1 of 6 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -532,6 +535,8 @@ void PbftManager::loopBackFinishState_() {
 bool PbftManager::stateOperations_() {
   pushSyncedPbftBlocksIntoChain_();
 
+  updateNextVotedValue_();
```
gaoqi8649, line 538: Next voted value should update at each new round

justinsnapp:
Replaced by PR #965

PR #947, opened 2021-07-05
fix: protect non finalized block map

mfrankovi:
## Purpose

Non finalized blocks were retrieved with reference which means is was accessed without any lock protection. Returning a copy instead

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -414,7 +414,12 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
         for (auto &block : level_blocks.second) {
           auto hash = blk_hash_t(block);
           if (known_non_finalized_blocks.count(hash) == 0) {
-            dag_blocks.emplace_back(db_->getDagBlock(hash));
+            if (auto blk = db_->getDagBlock(hash); blk) {
+              dag_blocks.emplace_back(blk);
+            } else {
+              LOG(log_er_dag_sync_) << "NonFinalizedBlock " << hash << " not in DB";
+              assert(false);
+            }
           }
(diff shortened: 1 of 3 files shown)


MatusKysel (approved):

src/network/taraxa_capability.cpp
```diff
@@ -411,7 +411,12 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
         for (auto &block : level_blocks.second) {
           auto hash = blk_hash_t(block);
           if (known_non_finalized_blocks.count(hash) == 0) {
-            dag_blocks.emplace_back(db_->getDagBlock(hash));
+            if (auto blk = db_->getDagBlock(hash); blk) {
+              dag_blocks.emplace_back(blk);
+            } else {
+              assert(false);
```
JakubFornadel, line 417: Lets switch this assert with LOG so we actually also get the log message before abort...

MatusKysel: I do not think we need to, assert is for debug and log for production. I doubt it will be saved somewhere on abort

JakubFornadel: not sure if I get this. Logs are for production as well as for debug. In case we abort here in debug and that abort is not printed in our boost logs, it would be nice to see the error message as last message, no ?

MatusKysel: abort will kill the node

JakubFornadel: so before killing the node it would be nice to print error message

justinsnapp (approved):

JakubFornadel (approved):

PR #950, opened 2021-07-06
chore(peers): make peer class more thread-safe

MatusKysel:
## Purpose

Small improvements to peer class

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -1021,8 +1021,8 @@ void TaraxaCapability::onNewTransactions(std::vector<taraxa::bytes> const &trans
     }
   }
   if (!fromNetwork || conf_.network_transaction_interval == 0) {
-    std::map<NodeID, std::vector<taraxa::bytes>> transactionsToSend;
-    std::map<NodeID, std::vector<trx_hash_t>> transactionsHashToSend;
+    std::unordered_map<NodeID, std::vector<taraxa::bytes>> transactionsToSend;
+    std::unordered_map<NodeID, std::vector<trx_hash_t>> transactionsHashToSend;
     {
       boost::shared_lock<boost::shared_mutex> lock(peers_mutex_);
       for (auto &peer : peers_) {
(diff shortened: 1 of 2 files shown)


src/network/taraxa_capability.cpp
```diff
@@ -981,7 +981,7 @@ void TaraxaCapability::onNewTransactions(std::vector<taraxa::bytes> const &trans
     for (auto &it : transactionsToSend) {
       sendTransactions(it.first, it.second);
     }
-    boost::unique_lock<boost::shared_mutex> lock(peers_mutex_);
+    boost::shared_lock<boost::shared_mutex> lock(peers_mutex_);
     for (auto &it : transactionsHashToSend) {
       for (auto &it2 : it.second) {
         peers_[it.first]->markTransactionAsKnown(it2);
```
JakubFornadel, line 987: Potential unprotected write into the peers_. If "it.first" is not present in "peers_", it is created here...

src/network/taraxa_capability.cpp
```diff
@@ -981,10 +981,10 @@ void TaraxaCapability::onNewTransactions(std::vector<taraxa::bytes> const &trans
     for (auto &it : transactionsToSend) {
       sendTransactions(it.first, it.second);
     }
-    boost::unique_lock<boost::shared_mutex> lock(peers_mutex_);
+    boost::shared_lock<boost::shared_mutex> lock(peers_mutex_);
     for (auto &it : transactionsHashToSend) {
       for (auto &it2 : it.second) {
-        peers_[it.first]->markTransactionAsKnown(it2);
+        if (peers_[it.first]) peers_[it.first]->markTransactionAsKnown(it2);
```
JakubFornadel, line 987: same mistake as before.... operator [] on map creates new element in case requested one does not exist. Use find instead https://en.cppreference.com/w/cpp/container/unordered_map/find

MatusKysel: thanks

justinsnapp (approved):

PR #954, opened 2021-07-06
fix: optimize vote processing

mfrankovi:
## Purpose

Optimized RLP votes processing.
Removed verify method in VrfPbftSortition which actually did not do anything since there was no check for return value.

These changes significantly reduce time of processing vote packets

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/consensus/vote.cpp
+++ b/src/consensus/vote.cpp
@@ -12,15 +12,14 @@ VrfPbftSortition::VrfPbftSortition(bytes const& b) {
   if (!rlp.isList()) {
     throw std::invalid_argument("VrfPbftSortition RLP must be a list");
   }
+  auto it = rlp.begin();
 
-  pk = rlp[0].toHash<vrf_pk_t>();
-  pbft_msg.type = PbftVoteTypes(rlp[1].toInt<uint>());
-  pbft_msg.round = rlp[2].toInt<uint64_t>();
-  pbft_msg.step = rlp[3].toInt<size_t>();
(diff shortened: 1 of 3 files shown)


gaoqi8649 (commented):
I just checked the code. I think CCL put the verify function there that is not for verification, that is for setup “output” and “threshold” value

gaoqi8649 (approved):
In vote verification has the verify() to setup output value. Should remove here

src/consensus/vote.cpp
```diff
@@ -311,7 +314,7 @@ std::vector<Vote> VoteManager::getVerifiedVotes(uint64_t const pbft_round, size_
     }
   }
 
-  for (auto const& v : votes_to_verify) {
+  for (auto& v : votes_to_verify) {
```
JakubFornadel, line 317: why did you remove const ?

src/consensus/vote.cpp
```diff
@@ -485,7 +487,7 @@ bool VoteManager::pbftBlockHasEnoughValidCertVotes(PbftBlockCert const& pbft_blo
   std::vector<Vote> valid_votes;
   auto first_cert_vote_round = pbft_block_and_votes.cert_votes[0].getRound();
 
-  for (auto const& v : pbft_block_and_votes.cert_votes) {
+  for (auto& v : pbft_block_and_votes.cert_votes) {
```
JakubFornadel, line 490: here too - remove const ?

MatusKysel (approved):

Issue #951, opened 2021-07-06
Implement protection from "invalid dag block" possible attack


JakubFornadel:
### Description
During syncing we do validate dag blocks in terms of it's tips/pivot. However for newly propagated block we do not process this check. As a result if someone send dag block with non-existent tip it would get into unverified_queue and block "block_worker" thread as it cannot build in-memory dag in case it encounters invalid dag block... There is no mechanism to solve such situations and block_worker thread just keeps trying to process the same invalid block over and over again(original approach) or just indefinitely waits for it's missing tip to arrive(Miho's fix). In both cases the node gets stuck... In case we receive block that does not pass this check, we should probably not add it to the unverified_queue at all

Check to be processed in newly propagated block:
https://github.com/Taraxa-project/taraxa-node/blob/develop/src/network/taraxa_capability.cpp#L125

Miho's fix:
https://github.com/Taraxa-project/taraxa-node/pull/931/files

What to think about:

- is Miho's fix reasonable ? It might be in terms of saving processor but in terms of solving the bug it is not I guess. If we want to keep it, lets put there at least some log message so we know we are stuck there, if not let's revert it
- Can we receive new dag block and not have received all of its tips/pivot ? If yes, it is invalid block but not maliciously invaldit so dropping it might not be the best solution at this case. Can we somehow save it and process it later maybe ?

PR #960, opened 2021-07-07
fix(NewBlockPacket): add behaviour on NewBlock that is missing pivot …

MatusKysel:
…or/and tips

## Purpose
In case we are receive block that is missing pivot/tips we should drop it and re-request it we all the missing blocks

resolves: https://github.com/Taraxa-project/taraxa-node/issues/951

### DIFF
--- a/src/network/taraxa_capability.cpp
+++ b/src/network/taraxa_capability.cpp
@@ -124,35 +124,54 @@ void TaraxaCapability::sealAndSend(NodeID const &nodeID, unsigned packet_type, R
   });
 }
 
-std::pair<bool, blk_hash_t> TaraxaCapability::checkDagBlockValidation(DagBlock const &block) {
+std::pair<bool, std::vector<blk_hash_t>> TaraxaCapability::checkDagBlockValidation(DagBlock const &block) {
+  std::vector<blk_hash_t> missing_blks;
+
   if (dag_blk_mgr_->getDagBlock(block.getHash())) {
     // The DAG block exist
-    return std::make_pair(true, blk_hash_t());
+    return std::make_pair(true, missing_blks);
   }
 
   level_t expected_level = 0;
   for (auto const &tip : block.getTips()) {
     auto tip_block = dag_blk_mgr_->getDagBlock(tip);
     if (!tip_block) {
-      LOG(log_nf_dag_sync_) << "Block " << block.getHash().toString() << " has a missing tip " << tip.toString();
-      return std::make_pair(false, tip);
+      LOG(log_er_dag_sync_) << "Block " << block.getHash().toString() << " has a missing tip " << tip.toString();
+      missing_blks.push_back(tip);
+    } else {
+      expected_level = std::max(tip_block->getLevel(), expected_level);
     }
-    expected_level = std::max(tip_block->getLevel(), expected_level);
   }
+
   auto pivot = block.getPivot();
   auto pivot_block = dag_blk_mgr_->getDagBlock(pivot);
   if (!pivot_block) {
-    LOG(log_nf_) << "Block " << block.getHash().toString() << " has a missing pivot " << pivot.toString();
-    return std::make_pair(false, pivot);
+    LOG(log_er_dag_sync_) << "Block " << block.getHash().toString() << " has a missing pivot " << pivot.toString();
+    missing_blks.push_back(pivot);
   }
+
+  if (missing_blks.size()) return std::make_pair(false, missing_blks);
+
   expected_level = std::max(pivot_block->getLevel(), expected_level);
   expected_level++;
   if (expected_level != block.getLevel()) {
-    throw InvalidDataException(std::string("Invalid block level ") + std::to_string(block.getLevel()) + " for block " +
-                               block.getHash().toString() + ". Expected level " + std::to_string(expected_level));
+    LOG(log_er_dag_sync_) << "Invalid block level " << block.getLevel() << " for block " << block.getHash()
+                          << " . Expected level " << expected_level;
+    return std::make_pair(false, missing_blks);
   }
 
-  return std::make_pair(true, blk_hash_t());
+  return std::make_pair(true, missing_blks);
+}
+
+void TaraxaCapability::requestBlocks(const NodeID &_nodeID, std::vector<blk_hash_t> const &blocks,
+                                     GetBlocksPacketRequestType mode) {
+  LOG(log_nf_dag_sync_) << "Sending GetBlocksPacket";
+  RLPStream s(blocks.size() + 1);  // Mode + block itself
+  s << mode;                       // Send mode first
+  for (auto blk : blocks) {
+    s << blk;
+  }
+  sealAndSend(_nodeID, GetBlocksPacket, s);
 }
 
 void TaraxaCapability::onConnect(weak_ptr<Session> session, u256 const &) {
@@ -350,6 +369,12 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
           LOG(log_dg_dag_prp_) << "Received NewBlock " << block.getHash().toString() << "that is already known";
           break;
         }
+        if (auto status = checkDagBlockValidation(block); !status.first) {
+          LOG(log_wr_dag_prp_) << "Received NewBlock " << block.getHash().toString() << " missing pivot or/and tips";
+          status.second.push_back(block.getHash());
+          requestBlocks(_nodeID, status.second);
+          break;
+        }
       }
 
       packet_stats.is_unique_ = true;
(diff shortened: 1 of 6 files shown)


src/network/taraxa_capability.cpp
```diff
@@ -400,17 +425,24 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
...
+      for (size_t i = 1; i < _r.itemCount(); ++i) {
+        blocks_hashes.emplace(_r[i]);
+      }
+
       const auto &blocks = dag_mgr_->getNonFinalizedBlocks();
       for (auto &level_blocks : blocks) {
         for (auto &block : level_blocks.second) {
           auto hash = blk_hash_t(block);
-          if (known_non_finalized_blocks.count(hash) == 0) {
+          if (blocks_hashes.count(hash) == (mode == MissingHashes)) {
```
mfrankovi, line 445: I am just wandering is there a possibility we get into some never ending loop of requesting blocks and not receiving them. I mean in theory it should not happen but if for some reason we have a bug where a node is missing something it should not be missing, we might flood the networking with a loop of getBlocksPacket and BlocksPacket. I think you should check here if the node has the missinghashes blocks. If it does not have them, I would just log an error and not send back the BlocksPacket to break such a loop

MatusKysel: I think this will not happened, because I am not processing block with missing tip/pivot, so if you do not send me what I need I will not re-request it again. If I add check there you wanted, what will happened if you send me blocks in wrong order and I needed pivot is last one, then I can to ask you again to send it to me because you will mark I have them

src/network/taraxa_capability.cpp
```diff
@@ -400,17 +425,25 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
...
+        blocks_hashes.emplace(_r[i]);
+      }
+
       const auto &blocks = dag_mgr_->getNonFinalizedBlocks();
       for (auto &level_blocks : blocks) {
         for (auto &block : level_blocks.second) {
           auto hash = blk_hash_t(block);
-          if (known_non_finalized_blocks.count(hash) == 0) {
+          if (blocks_hashes.count(hash) == (mode == MissingHashes) && !peer->isBlockKnown(hash)) {
+            peer->markBlockAsKnown(hash);
```
mfrankovi, line 446: I am not sure we should use markBlockAsKnown functionality for getBlocksPacket especially for mode KnownHashes. I believe it can create issues in some cases. At least we need a discussion about it

src/network/taraxa_capability.cpp
mfrankovi: But I thought we request for the block that had a missing tip/pivot as well since we drop it in first attempt. If we do not do that how do we ever get the original block again

src/network/taraxa_capability.cpp
MatusKysel: This was completely wrong

src/network/taraxa_capability.cpp
MatusKysel: Let's discuss this over tomorrow's call

mfrankovi (approved):

src/network/taraxa_capability.cpp
```diff
@@ -122,35 +122,53 @@ void TaraxaCapability::sealAndSend(NodeID const &nodeID, unsigned packet_type, R
...
+    missing_blks.push_back(pivot);
   }
+
+  if (missing_blks.size()) return std::make_pair(false, missing_blks);
+
   expected_level = std::max(pivot_block->getLevel(), expected_level);
   expected_level++;
   if (expected_level != block.getLevel()) {
     throw InvalidDataException(std::string("Invalid block level ") + std::to_string(block.getLevel()) + " for block " +
                                block.getHash().toString() + ". Expected level " + std::to_string(expected_level));
```
JakubFornadel, line 157: is this even caught somewhere ?

src/network/taraxa_capability.cpp
```diff
@@ -122,35 +122,53 @@ void TaraxaCapability::sealAndSend(NodeID const &nodeID, unsigned packet_type, R
...
     }
-    expected_level = std::max(tip_block->getLevel(), expected_level);
   }
+
   auto pivot = block.getPivot();
   auto pivot_block = dag_blk_mgr_->getDagBlock(pivot);
   if (!pivot_block) {
-    LOG(log_nf_) << "Block " << block.getHash().toString() << " has a missing pivot " << pivot.toString();
-    return std::make_pair(false, pivot);
+    LOG(log_nf_dag_sync_) << "Block " << block.getHash().toString() << " has a missing pivot " << pivot.toString();
```
JakubFornadel, line 147: Why info and not error or warning at least ?

src/network/taraxa_capability.cpp
```diff
@@ -347,6 +365,13 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
           LOG(log_dg_dag_prp_) << "Received NewBlock " << block.getHash().toString() << "that is already known";
           break;
         }
+        if (auto status = checkDagBlockValidation(block); !status.first) {
+          LOG(log_wr_dag_prp_) << "Received NewBlock " << block.getHash().toString()
+                               << "but we are dropping it as we are missing pivot or/and tips";
```
JakubFornadel, line 370: missing white space after hash and lets keep messages as short as possible, why not just write "Received invalid dag block - missing pivot and/or tips" ? All this part "but we are dropping it as we are" does not add any meaning/value to the message

src/network/taraxa_capability.cpp
```diff
@@ -400,21 +425,36 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
...
       std::vector<std::shared_ptr<DagBlock>> dag_blocks;
+      auto mode = static_cast<GetBlocksPacketRequestType>(_r[0].toInt<unsigned>());
+
+      if (mode == MissingHashes)
+        LOG(log_dg_dag_sync_) << "Received GetBlocksPacket with " << _r.itemCount() - 1 << " missing blocks";
+      else
+        LOG(log_dg_dag_sync_) << "Received GetBlocksPacket with " << _r.itemCount() - 1 << " known blocks";
+
+      for (size_t i = 1; i < _r.itemCount(); ++i) {
+        blocks_hashes.emplace(_r[i]);
```
JakubFornadel, line 438: lets use iteration instead of random access as rlp does not support random access

src/network/taraxa_capability.cpp
MatusKysel: I do not know, maybe it should crash whole node

src/network/taraxa_capability.cpp
MatusKysel: I think later is it logged as warning

src/network/taraxa_capability.cpp
MatusKysel: Ok, I will change this

src/network/taraxa_capability.cpp
MatusKysel: so why do not we remove operator[] from RLP?

src/network/taraxa_capability.cpp
JakubFornadel: What does it matter ? If something is warning/error it should be logged with appropriate level everywhere

src/network/taraxa_capability.cpp
```diff
@@ -400,21 +425,36 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
...
+            if (blocks_hashes.count(hash) == 1) dag_blocks.emplace_back(db_->getDagBlock(hash));
+          } else {
+            if (blocks_hashes.count(hash) == 0) dag_blocks.emplace_back(db_->getDagBlock(hash));
           }
         }
       }
+
+      // This means that someone requested more hashes that we actually have -> do not send anything
+      if (mode == MissingHashes && dag_blocks.size() != blocks_hashes.size()) {
+        LOG(log_nf_dag_sync_) << "Node " << _nodeID << " requested uknow DAG block from us";
```
JakubFornadel, line 455: - maybe at least warning ? 
- typo in "uknow"
- it's not "us" I guess  :D and again last 2 words useless info without any additional value

src/network/taraxa_capability.hpp
```diff
@@ -43,6 +43,13 @@ enum SubprotocolPacketType : ::byte {
   PacketCount
 };
 
+enum GetBlocksPacketRequestType : ::byte {
+
+  MissingHashes = 0x0,
+  KnownHashes,
+  RequestCount
```
JakubFornadel, line 50: what is this count good for ?

src/network/taraxa_capability.cpp
MatusKysel: I would not mark it as warning as this is not issue of this node.

src/network/taraxa_capability.cpp
```diff
@@ -400,21 +425,36 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
...
       const auto &blocks = dag_mgr_->getNonFinalizedBlocks();
       for (auto &level_blocks : blocks) {
         for (auto &block : level_blocks.second) {
           auto hash = blk_hash_t(block);
-          if (known_non_finalized_blocks.count(hash) == 0) {
-            dag_blocks.emplace_back(db_->getDagBlock(hash));
+          if (mode == MissingHashes) {
+            if (blocks_hashes.count(hash) == 1) dag_blocks.emplace_back(db_->getDagBlock(hash));
+          } else {
+            if (blocks_hashes.count(hash) == 0) dag_blocks.emplace_back(db_->getDagBlock(hash));
```
JakubFornadel, line 448: it is possible to call requestBlocks with "RequestCount" mode, even though it does not make sense but it is possible and this would not work then. Do we need that RequestCount ?

src/network/taraxa_capability.cpp
JakubFornadel: well my bad it supports it, but it is highly inefficient as it iterates the whole rlp from beginning when using [] until it gets to the index

src/network/taraxa_capability.hpp
MatusKysel: for the future use

MatusKysel: but I can remove it

src/network/taraxa_capability.cpp
JakubFornadel: so if I send you malicious dag block with invalid level, that should crash the node ? I dont think so

src/network/taraxa_capability.hpp
JakubFornadel: not really sure what future use that might be, but I might not see something. Either keep it but handle possible values of it correctly in other places or remove it....

JakubFornadel (approved):

PR #961, opened 2021-07-07
chore(circleci): Removing build of special ctest image, making ctest …

agrebin:
## Purpose

Ctest build takes a long time and is not taking advantage of cache some times, this PR is to try to improve that 

## For reviewers
Ctest image is the one that actually is being used as cache, will try to improve this behaviour
## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -422,21 +422,21 @@ jobs:
               command: |
                 docker build -t ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} .
 
-      - docker-cache/with-save-restore-images:
-         repository: gcr.io/${GOOGLE_PROJECT_ID}
-         images: >-
-           ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} 
-           ${IMAGE}-ctest:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}
-           ${IMAGE}-cache:latest
-           ${IMAGE}-cache:${DOCKER_BRANCH_TAG}
-         steps: 
-           - docker-cache/build:
-              command: |
-               docker build -t ${IMAGE}-ctest:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} \
-                 --target build .
-           - docker-cache/tag: 
-               tags: >-
-                  ${IMAGE}-ctest:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} ${IMAGE}-cache:${DOCKER_BRANCH_TAG}
+#      - docker-cache/with-save-restore-images:
+#         repository: gcr.io/${GOOGLE_PROJECT_ID}
+#         images: >-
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -412,31 +412,41 @@ jobs:
             echo ${GCLOUD_SERVICE_KEY} | docker login -u _json_key --password-stdin https://gcr.io
 
       - docker-cache/with-save-restore-images:
+         parallelism: 8
          repository: gcr.io/${GOOGLE_PROJECT_ID}
          images: >-
            ${IMAGE}-cache:${DOCKER_BRANCH_TAG}
            ${IMAGE}-cache:latest
-           ${IMAGE}-develop:latest
+           #${IMAGE}-develop:latest
```
rattrap, line 420: `#` doesn't work here. Just remove the line.

```
Error parsing reference: "#taraxa-node-develop:latest" is not a valid repository/tag: invalid reference format
```

Dockerfile
```diff
@@ -93,9 +93,9 @@ RUN mkdir $BUILD_OUTPUT_DIR && cd $BUILD_OUTPUT_DIR \
     ../ \
     && make -j$(nproc) all \
     # Copy CMake generated Testfile to be able to trigger ctest from bin directory
-    && cp tests/CTestTestfile.cmake bin/ \
+    && cp tests/CTestTestfile.cmake bin/ 
     # keep only required shared libraries and final binaries
-    && find . -maxdepth 1 ! -name "lib" ! -name "bin" -exec rm -rfv {} \;
+    # && find . -maxdepth 1 ! -name "lib" ! -name "bin" -exec rm -rfv {} \;
```
rattrap, line 98: I don't think commenting this helps.

rattrap (changes requested):
Yeah. Have no idea why/if this works haha.
And I don't think you're building the "regular" image anymore. Building or pushing it. I check in Google Cloud Registry

.circleci/config.yml
agrebin: No, yes, it is being built, it is just not cached (saved/pushed as cache)

Dockerfile
agrebin: no, youŕe right it is a leftover from a test, fixed it.

.circleci/config.yml
```diff
@@ -412,33 +412,24 @@ jobs:
             echo ${GCLOUD_SERVICE_KEY} | docker login -u _json_key --password-stdin https://gcr.io
 
       - docker-cache/with-save-restore-images:
+         parallelism: 8
          repository: gcr.io/${GOOGLE_PROJECT_ID}
          images: >-
            ${IMAGE}-cache:${DOCKER_BRANCH_TAG}
            ${IMAGE}-cache:latest
-           ${IMAGE}-develop:latest
+           ${IMAGE}-ctest:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM}
```
sfrique, line 420: Having the `${CIRCLE_BUIL_NUM}` make sure this is unique, does it make sense to add unique images to the cache?

sfrique (approved):
Other than the command, looks fine.

PR #963, opened 2021-07-08
perf(dag): remove std::string from DAG graph

MatusKysel:
## Purpose

Removed std::string from DAG related code, so we do not need to do conversion string<->hash all the time

Tests were just modified because zero hash is not supported  

Please check also aleth PR https://github.com/Taraxa-project/taraxa-aleth/pull/26


JakubFornadel (commented):
Seems like a great improvement but lets wait for call if there is not some hidden reason why we used strings instead of hashes...

JakubFornadel (approved):

justinsnapp (approved):

gaoqi8649 (approved):

PR #972, opened 2021-07-10
fix(pbft): give up on soft voted value if never receive valid dag blocks for it

justinsnapp:
## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -152,6 +152,17 @@ void PbftManager::resume() {
   daemon_ = std::make_unique<std::thread>([this]() { continuousOperation_(); });
 }
 
+// Only to be used for tests...
+void PbftManager::setMaxWaitForSoftVotedBlock_ms(uint64_t wait_ms) {
+  max_wait_for_soft_voted_block_steps_ms_ = wait_ms;
+}
+
+// Only to be used for tests...
+void PbftManager::setMaxWaitForNextVotedBlock_ms(uint64_t wait_ms) {
+  max_wait_for_next_voted_block_steps_ms_ = wait_ms;
+}
+
+// Only to be used for tests...
 void PbftManager::resumeSingleState() {
   if (!stopped_.load()) daemon_->join();
   stopped_ = false;
@@ -171,6 +182,7 @@ void PbftManager::resumeSingleState() {
   daemon_ = std::make_unique<std::thread>([this]() { doNextState_(); });
 }
 
+// Only to be used for tests...
 void PbftManager::doNextState_() {
   auto initial_state = state_;
 
@@ -202,7 +214,10 @@ void PbftManager::doNextState_() {
     }
 
     setNextState_();
-    if (state_ == initial_state) sleep_();
+    if (state_ != initial_state) {
+      return;
+    }
+    sleep_();
   }
 }
 
@@ -422,7 +437,11 @@ void PbftManager::sleep_() {
 
 void PbftManager::initialState_() {
   // Initial PBFT state
+
+  // Time constants...
   LAMBDA_ms = LAMBDA_ms_MIN;
+  max_wait_for_soft_voted_block_steps_ms_ = MAX_WAIT_FOR_SOFT_VOTED_BLOCK_STEPS * 2 * LAMBDA_ms;
+  max_wait_for_next_voted_block_steps_ms_ = MAX_WAIT_FOR_NEXT_VOTED_BLOCK_STEPS * 2 * LAMBDA_ms;
 
   auto round = db_->getPbftMgrField(PbftMgrRoundStep::PbftRound);
   auto step = db_->getPbftMgrField(PbftMgrRoundStep::PbftStep);
@@ -497,6 +516,7 @@ void PbftManager::initialState_() {
 
   // Used for detecting timeout due to malicious node
   // failing to gossip PBFT block or DAG blocks
+  // Requires that soft_voted_block_for_this_round_ already be initialized from db
   initializeVotedValueTimeouts_();
 
   executed_pbft_block_ = db_->getPbftMgrStatus(PbftMgrStatus::executed_block);
@@ -710,14 +730,19 @@ bool PbftManager::updateSoftVotedBlockForThisRound_() {
 void PbftManager::initializeVotedValueTimeouts_() {
   time_began_waiting_next_voted_block_ = std::chrono::system_clock::now();
   time_began_waiting_soft_voted_block_ = std::chrono::system_clock::now();
+
+  // Concern: Requires that soft_voted_block_for_this_round_ already be initialized from db
   if (soft_voted_block_for_this_round_.first != NULL_BLOCK_HASH && soft_voted_block_for_this_round_.second) {
     last_soft_voted_value_ = soft_voted_block_for_this_round_.first;
   }
 }
 
-void PbftManager::setLastSoftVotedValue(blk_hash_t soft_voted_value) { checkSoftVotedValueChange_(soft_voted_value); }
+void PbftManager::setLastSoftVotedValue(blk_hash_t soft_voted_value) {
+  // TODO: Add a check for some kind of guard to ensure this is only called from within a test
+  updateLastSoftVotedValue_(soft_voted_value);
+}
 
-void PbftManager::checkSoftVotedValueChange_(blk_hash_t const new_soft_voted_value) {
+void PbftManager::updateLastSoftVotedValue_(blk_hash_t const new_soft_voted_value) {
   if (new_soft_voted_value != last_soft_voted_value_) {
     time_began_waiting_soft_voted_block_ = std::chrono::system_clock::now();
   }
@@ -818,7 +843,7 @@ void PbftManager::identifyBlock_() {
 
       auto place_votes = placeVote_(leader_block.first, soft_vote_type, round, step_);
 
-      checkSoftVotedValueChange_(leader_block.first);
+      updateLastSoftVotedValue_(leader_block.first);
 
       if (place_votes) {
         LOG(log_nf_) << "Soft votes " << place_votes << " voting block " << leader_block.first << " at round " << round;
@@ -827,7 +852,10 @@ void PbftManager::identifyBlock_() {
   } else if (round >= 2 && previous_round_next_voted_value_ != NULL_BLOCK_HASH) {
     auto place_votes = placeVote_(previous_round_next_voted_value_, soft_vote_type, round, step_);
 
-    checkSoftVotedValueChange_(previous_round_next_voted_value_);
+    // Generally this value will either be the same as last soft voted value from previous round
+    // but a node could have observed a previous round next voted value that differs from what they
+    // soft voted.
+    updateLastSoftVotedValue_(previous_round_next_voted_value_);
 
     if (place_votes) {
       LOG(log_nf_) << "Soft votes " << place_votes << " voting block " << previous_round_next_voted_value_
@@ -1591,13 +1619,13 @@ bool PbftManager::giveUpSoftVotedBlock_() {
   unsigned long elapsed_wait_soft_voted_block_in_ms =
       std::chrono::duration_cast<std::chrono::milliseconds>(soft_voted_block_wait_duration).count();
 
-  if (elapsed_wait_soft_voted_block_in_ms > MAX_WAIT_FOR_SOFT_VOTED_BLOCK_STEPS * 2 * LAMBDA_ms) {
+  if (elapsed_wait_soft_voted_block_in_ms > max_wait_for_soft_voted_block_steps_ms_) {
     LOG(log_dg_) << "Have been waiting " << elapsed_wait_soft_voted_block_in_ms << "ms for soft voted block "
                  << last_soft_voted_value_ << ", giving up on this value.";
     return true;
   } else {
     LOG(log_tr_) << "Have only been waiting " << elapsed_wait_soft_voted_block_in_ms << "ms for soft voted block "
-                 << last_soft_voted_value_ << "(after " << MAX_WAIT_FOR_SOFT_VOTED_BLOCK_STEPS * 2 * LAMBDA_ms
+                 << last_soft_voted_value_ << "(after " << max_wait_for_soft_voted_block_steps_ms_
                  << "ms will give up on this value)";
   }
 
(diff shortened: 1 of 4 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -1524,9 +1558,8 @@ bool PbftManager::giveUpNextVotedBlock_() {
     }
     // Read from DB pushing into unverified queue
     pbft_chain_->pushUnverifiedPbftBlock(pbft_block);
-  }
-
-  if (pbft_block) {
+  } else {
```
gaoqi8649, line 1561: I think cannot change here. The previous approach could handle the PBFT block doesn't in the unverified queue but in the DB

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+  pbft_chain->pushUnverifiedPbftBlock(propose_pbft_block);
+
+  // Generate bogus vote
+  auto round = 1;
+  auto step = 4;
+  auto weighted_index = 0;
+  auto propose_vote = pbft_mgr->generateVote(pbft_block_hash, next_vote_type, round, step, weighted_index);
+  vote_mgr->addVerifiedVote(propose_vote);
+
+  // pbft_mgr->updateSoftVotedBlockForThisRound_(std::make_pair(pbft_block_hash, true));
```
gaoqi8649, line 215: please cleanup

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+  // Generate bogus DAG anchor for PBFT block
+  // auto dag_anchor = blk_hash_t("1234567890000000000000000000000000000000000000000000000000000000");
+  // auto last_pbft_block_hash = pbft_chain->getLastPbftBlockHash();
+  // auto propose_pbft_period = pbft_chain->getPbftChainSize() + 1;
+  // auto beneficiary = nodes[0]->getAddress();
+  auto node_sk = nodes[0]->getSecretKey();
+  // auto propose_pbft_block =
+  //    std::make_shared<PbftBlock>(last_pbft_block_hash, dag_anchor, propose_pbft_period, beneficiary, node_sk);
+  // auto pbft_block_hash = propose_pbft_block->getBlockHash();
+  // pbft_chain->pushUnverifiedPbftBlock(propose_pbft_block);
```
gaoqi8649, line 273: please cleanup

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+  auto pbft_block_hash = blk_hash_t("0000000100000000000000000000000000000000000000000000000000000000");
+  auto propose_vote = pbft_mgr->generateVote(pbft_block_hash, next_vote_type, round, step, weighted_index);
+  vote_mgr->addVerifiedVote(propose_vote);
+
+  // pbft_mgr->updateSoftVotedBlockForThisRound_(std::make_pair(pbft_block_hash, true));
+  pbft_mgr->start();
+
+  // Vote at the bogus PBFT block hash
+  EXPECT_HAPPENS({60s, 50ms}, [&](auto &ctx) {
+    auto soft_vote_value = blk_hash_t(0);
```
gaoqi8649, line 288: propose value?

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+    auto votes = vote_mgr->getVerifiedVotes();
+    for (auto const &v : votes) {
+      if (soft_vote_type == v.getType() && v.getBlockHash() == pbft_block_hash) {
```
gaoqi8649, line 291: propose value?

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+        soft_vote_value = v.getBlockHash();
```
gaoqi8649, line 292: propose value

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+        break;
+      }
+    }
+    WAIT_EXPECT_EQ(ctx, soft_vote_value, pbft_block_hash)
```
gaoqi8649, line 296: propose value

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+  });
+
+  auto start_round = pbft_mgr->getPbftRound();
+
+  cout << "After some time, terminate voting on the missing proposed block " << pbft_block_hash << endl;
+  // After some rounds, terminate the bogus value and vote on NULL_BLOCK_HASH since no new DAG blocks generated
+  EXPECT_HAPPENS({60s, 50ms}, [&](auto &ctx) {
+    auto soft_vote_value = pbft_block_hash;
```
gaoqi8649, line 304: propose value

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+    auto votes = vote_mgr->getVerifiedVotes();
+    for (auto const &v : votes) {
+      if (next_vote_type == v.getType() && v.getBlockHash() == blk_hash_t(0)) {
```
gaoqi8649, line 307: propose value

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+        soft_vote_value = v.getBlockHash();
```
gaoqi8649, line 308: propose value

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+        break;
+      }
+    }
+    WAIT_EXPECT_EQ(ctx, soft_vote_value, blk_hash_t(0))
```
gaoqi8649, line 312: propose value

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+  });
+
+  cout << "Wait ensure node is still advancing in rounds... " << endl;
+  EXPECT_HAPPENS({60s, 50ms}, [&](auto &ctx) { WAIT_EXPECT_NE(ctx, start_round, pbft_mgr->getPbftRound()) });
+}
+
+TEST_F(PbftManagerTest, DISABLED_terminate_missing_voted_pbft_block) {
```
gaoqi8649, line 319: Disabled the unit test, does it cannot pass?
And please comment what is situation the unit test testing for. Looks like this is a duplicate unit test as terminate_missing_proposed_pbft_block

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,197 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+  auto pbft_chain1 = nodes[0]->getPbftChain();
+  auto vote_mgr1 = nodes[0]->getVoteManager();
+
+  // Generate bogus votes
+  auto round = 1;
+  auto step = 4;
+  auto weighted_index = 0;
+  auto pbft_block_hash = blk_hash_t("0000000100000000000000000000000000000000000000000000000000000000");
+
+  // pbft_mgr->updateSoftVotedBlockForThisRound_(std::make_pair(pbft_block_hash, true));
```
gaoqi8649, line 340: please clean up

tests/pbft_manager_test.cpp
justinsnapp: no, want to check the soft voted value

tests/pbft_manager_test.cpp
justinsnapp: no, want to check the soft voted value

tests/pbft_manager_test.cpp
justinsnapp: no, want to check the soft voted value

tests/pbft_manager_test.cpp
justinsnapp: no, want to check the soft voted value, (which in a single node case will be the propose value I suppose)

src/consensus/pbft_manager.cpp
justinsnapp: You are right, fixed it.

src/consensus/pbft_manager.cpp
```diff
@@ -582,7 +668,63 @@ bool PbftManager::stateOperations_() {
...
+
+void PbftManager::initializeVotedValueTimeouts_() {
+  time_began_waiting_next_voted_block_ = std::chrono::system_clock::now();
+  time_began_waiting_soft_voted_block_ = std::chrono::system_clock::now();
+  if (soft_voted_block_for_this_round_.first != NULL_BLOCK_HASH && soft_voted_block_for_this_round_.second) {
+    last_soft_voted_value_ = soft_voted_block_for_this_round_.first;
+  }
+}
+
+void PbftManager::setLastSoftVotedValue(blk_hash_t soft_voted_value) { checkSoftVotedValueChange_(soft_voted_value); }
```
gaoqi8649, line 733: The function did the same thing with checkSoftVotedValueChange_, could remove and call checkSoftVotedValueChange_ directly

src/consensus/pbft_manager.cpp
```diff
@@ -582,7 +668,63 @@ bool PbftManager::stateOperations_() {
...
+
+void PbftManager::checkSoftVotedValueChange_(blk_hash_t const new_soft_voted_value) {
```
gaoqi8649, line 735: This is not only check. Maybe change name to updateSoftVotedValue is better?

src/consensus/pbft_manager.cpp
```diff
@@ -675,12 +732,18 @@ void PbftManager::identifyBlock_() {
...
+      checkSoftVotedValueChange_(leader_block.first);
+
       if (place_votes) {
         LOG(log_nf_) << "Soft votes " << place_votes << " voting block " << leader_block.first << " at round " << round;
       }
     }
   } else if (round >= 2 && previous_round_next_voted_value_ != NULL_BLOCK_HASH) {
     auto place_votes = placeVote_(previous_round_next_voted_value_, soft_vote_type, round, step_);
+
+    checkSoftVotedValueChange_(previous_round_next_voted_value_);
```
gaoqi8649, line 845: I don't understand why need checkSoftVotedValueChange_ here. Here will always place previous round next voted value. I think the value will never change here

src/consensus/pbft_manager.cpp
```diff
@@ -126,6 +126,87 @@ void PbftManager::run() {
   // Initialize PBFT status
   initialState_();
 
+  continuousOperation_();
+}
+
+// Only to be used for tests...
```
gaoqi8649, line 132: I think resume() is not for test, and resumeSingleState() is for test only?

src/consensus/pbft_manager.cpp
```diff
@@ -126,6 +126,87 @@ void PbftManager::run() {
...
+void PbftManager::resume() {
+  if (!stopped_.load()) daemon_->join();
+  stopped_ = false;
+
+  // Will only appear in testing...
```
gaoqi8649, line 137: If resume() is not for test, need remove this

src/consensus/pbft_manager.cpp
```diff
@@ -675,12 +817,18 @@ void PbftManager::identifyBlock_() {
...
+      checkSoftVotedValueChange_(leader_block.first);
+
       if (place_votes) {
         LOG(log_nf_) << "Soft votes " << place_votes << " voting block " << leader_block.first << " at round " << round;
       }
     }
   } else if (round >= 2 && previous_round_next_voted_value_ != NULL_BLOCK_HASH) {
     auto place_votes = placeVote_(previous_round_next_voted_value_, soft_vote_type, round, step_);
+
+    checkSoftVotedValueChange_(previous_round_next_voted_value_);
```
gaoqi8649, line 845: I don't understand here. Here will always propose previous round next voted value, never change.
Maybe you want to terminate the soft voted value if after max waiting time?

src/consensus/pbft_manager.hpp
```diff
@@ -79,6 +81,9 @@ class PbftManager {
       std::shared_ptr<std::vector<std::pair<blk_hash_t, std::vector<bool>>>> trx_overlap_table);
   size_t getPbftCommitteeSize() const { return COMMITTEE_SIZE; }
   u_long getPbftInitialLambda() const { return LAMBDA_ms_MIN; }
+  void setLastSoftVotedValue(blk_hash_t soft_voted_value);
```
gaoqi8649, line 84: Should remove

tests/pbft_manager_test.cpp
```diff
@@ -181,6 +181,185 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
   }
 }
 
+// Test that after some amount of elapsed time will not continue soft voting for same value
+TEST_F(PbftManagerTest, terminate_soft_voting_pbft_block) {
```
gaoqi8649, line 185: I think we should not design unit test like this. This is not simulate the real process in PBFT manager.
And I think we don't need this. Since terminate soft voted value because the value include a bogus anchor. So the unit test looks like testing same thing with terminate_bogus_dag_anchor

justinsnapp: @gaoqi8649  there is a fundamental difference between the two tests, in that in one case we have received the block (but its "bogus") and in the other we have never received the block.

src/consensus/pbft_manager.cpp
justinsnapp: My thinking was to add a guard once we have one (or do we already have one) so that can ensure this public function is only called from within a test.

src/consensus/pbft_manager.cpp
justinsnapp: @gaoqi8649 , I believe generally this value will either be the same as last soft voted value from previous round, but a node could have observed a previous round next voted value that differs from what they had individually soft voted.

src/consensus/pbft_manager.cpp
justinsnapp: Why do you say that, where is it used besides in the test?  I just added it myself for this test (and future tests).  ha.

gaoqi8649 (approved):
In order to write an unit test terminate_soft_voting_pbft_block, add many test only code in PBFT manager daemon. I don't like the design
We have cleaned so many test only code already, and should avoid to add test only code in main code base. But I don't have a better design for the unit test
Others looks good to me

justinsnapp:
> In order to write an unit test terminate_soft_voting_pbft_block, add many test only code in PBFT manager daemon. I don't like the design
> We have cleaned so many test only code already, and should avoid to add test only code in main code base. But I don't have a better design for the unit test
> Others looks good to me

@gaoqi8649, I agree.  I think some re-write though of pbft_manager to enable easy testing of state machine transitions would be beneficial.   I believer after you are done refactoring the vote manager class then pbft_manager can be refactored to be a state machine that receives only events (timeouts, and sufficient vote totals for values) and uses those events to update the state.  Then tests within pbft manager test can be well re-written to test specific state changes.

PR #982, opened 2021-07-16
Syncing

MatusKysel:


src/network/syncing_state.cpp
```diff
@@ -0,0 +1,61 @@
...
+bool SyncingState::is_actively_syncing() const {
+  auto now = std::chrono::steady_clock::now();
+
+  std::shared_lock lock(time_mutex_);
+  return std::chrono::duration_cast<std::chrono::seconds>(now - last_received_sync_packet_time_) <=
+         SYNCING_INACTIVITY_THRESHOLD;
+}
+
+void SyncingState::set_peer_malicious() {
+  std::shared_lock lock(peer_mutex_);
```
MatusKysel, line 49: comment

.circleci/config.yml
```diff
@@ -235,8 +235,8 @@ commands:
               helm upgrade -i pr-<< parameters.pr-name >> taraxa-node/ --namespace blockchain-prnet -f taraxa-node/values.yaml -f taraxa-node/prnet.values.yaml \
                 --set image.repository=gcr.io/jovial-meridian-249123/taraxa-node \
                 --set image.tag=${PR_TAG} \
-                --set explorer.configEnvs.rpcHttpProvider="http://pr-<< parameters.pr-name >>-taraxa-node:7777" \
-                --set explorer.configEnvs.rpcWsProvider="http://pr-<< parameters.pr-name >>-taraxa-node:8777" \
+                --set explorer.configEnvs.rpcHttpProvider="http://pr-<< parameters.pr-name >>-taraxa-node-1.pr-<< parameters.pr-name >>-taraxa-node:7777" \
+                --set explorer.configEnvs.rpcWsProvider="http://pr-<< parameters.pr-name >>-taraxa-node-1.pr-<< parameters.pr-name >>-taraxa-node:8777" \
```
JakubFornadel, line 239: hmm this is your change ?

CMakeLists.txt
```diff
@@ -3,7 +3,7 @@ cmake_minimum_required(VERSION 3.16)
 # Set current version of the project
 set(TARAXA_MAJOR_VERSION 1)
 set(TARAXA_MINOR_VERSION 9)
-set(TARAXA_PATCH_VERSION 1)
+set(TARAXA_PATCH_VERSION 2)
```
JakubFornadel, line 6: btw we dont really check patch version anywhere, we only send it in status packet and log it... is that intended behaviour ?

.circleci/config.yml
MatusKysel: check the commits it's Leo's

CMakeLists.txt
MatusKysel: yes, it was requirement from Justin not to check patch version anywhere (as it should not be breaking change)

src/consensus/pbft_manager.cpp
```diff
@@ -662,7 +662,7 @@ bool PbftManager::stateOperations_() {
   // CHECK IF WE HAVE RECEIVED 2t+1 CERT VOTES FOR A BLOCK IN OUR CURRENT
   // ROUND.  IF WE HAVE THEN WE EXECUTE THE BLOCK
   // ONLY CHECK IF HAVE *NOT* YET EXECUTED THIS ROUND...
-  if (state_ == certify_state && !have_executed_this_round_) {
+  if (!is_syncing_() && state_ == certify_state && !have_executed_this_round_) {
```
JakubFornadel, line 665: why not move this "!is_syncing_()" check higher ? For example getVerifiedVotes call sets some internal data that is used in resetRound_() ? Because if not it seems like it is useless call

src/final_chain/replay_protection_service.cpp
```diff
@@ -76,7 +76,7 @@ class ReplayProtectionServiceImpl : public virtual ReplayProtectionService {
       } else {
         sender_state = loadSenderState(senderStateKey(sender_addr));
         if (!sender_state) {
-          sender_states[sender_addr] = sender_states_dirty[sender_addr] = s_ptr(new SenderState(trx.nonce));
+          sender_states[sender_addr] = sender_states_dirty[sender_addr] = std::make_shared<SenderState>(trx.nonce);
```
JakubFornadel, line 79: cool :)

src/network/syncing_state.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+void SyncingState::set_pbft_syncing(bool syncing, const std::optional<dev::p2p::NodeID>& peer_id) {
+  assert((syncing && peer_id) || !syncing);
+  pbft_syncing_ = syncing;
+
+  if (peer_id) {
+    set_peer(peer_id.value());
+  }
+  // Reset last sync packet time when syncing is restarted/fresh syncing flag is set
+  set_last_sync_packet_time();
```
JakubFornadel, line 23: this does not seem ok ?? I guess it should be called only at the end of pbft packet processing

src/network/taraxa_capability.cpp
```diff
@@ -546,21 +561,26 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
...
         if (block.getLevel() > peer->dag_level_) peer->dag_level_ = block.getLevel();
         dag_blk_mgr_->insertBroadcastedBlockWithTransactions(block, newTransactions);
+      }
 
-        if (iBlock + transactionCount + 1 >= itemCount) break;
+      if (is_final_sync_packet) {
+        syncing_state_.set_dag_syncing(false);
+        LOG(log_nf_dag_sync_) << "Received final DagBlocksSyncPacket with blocks: " << received_dag_blocks_str;
+      } else {
+        syncing_state_.set_last_sync_packet_time();
```
JakubFornadel, line 577: I guess I would double check where we call this and if we are sure we call it only on places where we should

CMakeLists.txt
justinsnapp: @JakubFornadel  well these version checks are only for our own convenience during testnet stage.  During mainnet we will only be able to defend checking protocol version, because of course anyone can implement their own node design.    So I suggested that we only check major and minor version and leave patch free to be increased without requiring other nodes to upgrade to interoperate, for example for performance and other chores.

JakubFornadel:
Seems ok in terms of making syncing state thread safe as it was not before, synchronous -> fixing error messages, although I am still not so sure the perf as when I was trying it the original syncing was faster as it requests next syncing block faster than this synchronous version but lets believe your measurements :)

As of our final goal for refactor the main challenge is still there:

- separate initial syncing from consensus triggered syncing
- block processing of all packets during initial syncing
- implement automatic snapshots to be downloaded and then just sync from that

JakubFornadel:
Giving approve as I wont be here in the evening but please check those few comments I had

JakubFornadel (approved):

src/network/syncing_state.cpp
MatusKysel: it should be there but only if we are setting syncing to true, now it should be fixed

src/consensus/pbft_manager.cpp
MatusKysel: moved

src/network/taraxa_capability.cpp
MatusKysel: it is called when syncing packet is received and it passes all check. And the other palace is when we are setting syncing to true

kstdl (approved):

PR #985, opened 2021-07-17
perf(votes): rewrite verified votes table

gaoqi8649:
## Purpose

Fix verified votes consume lots of memory issue when PBFT cannot make consensus on voting

## For reviewers

1. Only save verified votes hash in verified table
2. PBFT manager will not get all verified votes for current round, only get number of votes 

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1050,19 +1050,19 @@ void PbftManager::secondFinish_() {
   loop_back_finish_state_ = elapsed_time_in_round_ms_ > end_time_for_step;
 }
 
-std::shared_ptr<Vote> PbftManager::generateVote(blk_hash_t const &blockhash, PbftVoteTypes type, uint64_t round, size_t step,
-                               size_t weighted_index) {
+std::shared_ptr<Vote> PbftManager::generateVote(blk_hash_t const &blockhash, PbftVoteTypes type, uint64_t round,
+                                                size_t step, size_t weighted_index) {
   // sortition proof
   VrfPbftMsg msg(type, round, step, weighted_index);
   VrfPbftSortition vrf_sortition(vrf_sk_, msg);
   Vote vote(node_sk_, vrf_sortition, blockhash);
 
-  return std::shared_ptr<Vote>(vote);
+  return std::make_shared<Vote>(vote);
 }
 
 size_t PbftManager::placeVote_(taraxa::blk_hash_t const &blockhash, PbftVoteTypes vote_type, uint64_t round,
                                size_t step) {
-  vector<Vote> votes;
+  std::vector<std::shared_ptr<Vote>> votes;
 
   for (size_t weighted_index(0); weighted_index < weighted_votes_count_; weighted_index++) {
     if (step == 1 && weighted_index > 0) {
@@ -1394,7 +1394,7 @@ std::pair<vec_blk_t, bool> PbftManager::comparePbftBlockScheduleWithDAGblocks_(P
 }
 
 bool PbftManager::pushCertVotedPbftBlockIntoChain_(taraxa::blk_hash_t const &cert_voted_block_hash,
-                                                   std::vector<Vote> const &cert_votes_for_round) {
+                                                   std::vector<std::shared_ptr<Vote>> const &cert_votes_for_round) {
   auto pbft_block = getUnfinalizedBlock_(cert_voted_block_hash);
   if (!pbft_block) {
     LOG(log_nf_) << "Can not find the cert voted block hash " << cert_voted_block_hash << " in both pbft queue and DB";
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -352,11 +352,8 @@ void PbftManager::resetStep_() {
 
 bool PbftManager::resetRound_() {
   bool restart = false;
-  // Check if we are synced to the right step ...
-  uint64_t consensus_pbft_round = roundDeterminedFromVotes_();
-  // Check should be always true...
+  auto consensus_pbft_round = vote_mgr_->roundDeterminedFromVotes(TWO_T_PLUS_ONE);
   auto round = getPbftRound();
-  assert(consensus_pbft_round >= round);
 
   if (consensus_pbft_round > round) {
     LOG(log_nf_) << "From votes determined round " << consensus_pbft_round;
@@ -658,19 +655,15 @@ bool PbftManager::stateOperations_() {
   LOG(log_tr_) << "PBFT current round is " << round;
   LOG(log_tr_) << "PBFT current step is " << step_;
 
-  // Get votes
-  votes_ = vote_mgr_->getVerifiedVotes(round, sortition_threshold_, getDposTotalVotesCount(),
-                                       [this](auto const &addr) { return dpos_eligible_vote_count_(addr); });
-
-  LOG(log_tr_) << "There are " << votes_.size() << " total votes in round " << round;
+  // Periodically verify unverified votes
+  vote_mgr_->verifyVotes(round, sortition_threshold_, getDposTotalVotesCount(),
+                         [this](auto const &addr) { return dpos_eligible_vote_count_(addr); });
 
   // CHECK IF WE HAVE RECEIVED 2t+1 CERT VOTES FOR A BLOCK IN OUR CURRENT
   // ROUND.  IF WE HAVE THEN WE EXECUTE THE BLOCK
   // ONLY CHECK IF HAVE *NOT* YET EXECUTED THIS ROUND...
   if (state_ == certify_state && !have_executed_this_round_) {
-    std::vector<Vote> cert_votes_for_round = getVotesOfTypeFromVotesForRoundAndStep_(
-        cert_vote_type, votes_, round, 3, std::make_pair(NULL_BLOCK_HASH, false));
-    auto voted_block_hash_with_cert_votes = blockWithEnoughVotes_(cert_votes_for_round);
+    auto voted_block_hash_with_cert_votes = vote_mgr_->getVotesBundleByRoundAndStep(round, 3, TWO_T_PLUS_ONE);
     if (voted_block_hash_with_cert_votes.enough) {
       LOG(log_dg_) << "PBFT block " << voted_block_hash_with_cert_votes.voted_block_hash << " has enough certed votes";
       // put pbft block into chain
@@ -840,7 +833,8 @@ void PbftManager::identifyBlock_() {
 
   if (round == 1 || (round >= 2 && giveUpNextVotedBlock_())) {
     // Identity leader
-    std::pair<blk_hash_t, bool> leader_block = identifyLeaderBlock_(votes_);
+    auto proposal_votes = vote_mgr_->getProposalVotes(round);
+    std::pair<blk_hash_t, bool> leader_block = identifyLeaderBlock_(proposal_votes);
     if (leader_block.second) {
       db_->savePbftMgrVotedValue(PbftMgrVotedValue::own_starting_value_in_round, leader_block.first);
       own_starting_value_for_round_ = leader_block.first;
@@ -1059,126 +1053,14 @@ void PbftManager::secondFinish_() {
   loop_back_finish_state_ = elapsed_time_in_round_ms_ > end_time_for_step;
 }
 
-// There is a quorum of next-votes and set determine that round p should be the current round...
-uint64_t PbftManager::roundDeterminedFromVotes_() {
-  // <<vote_round, vote_step>, count>, <round, step> store in reverse order
-  std::map<std::pair<uint64_t, size_t>, size_t, std::greater<std::pair<uint64_t, size_t>>>
-      next_votes_tally_by_round_step;
-  auto round = getPbftRound();
-
-  for (auto const &v : votes_) {
-    if (v.getType() != next_vote_type) {
-      continue;
-    }
-    std::pair<uint64_t, size_t> round_step = std::make_pair(v.getRound(), v.getStep());
-    if (round_step.first >= round) {
-      if (next_votes_tally_by_round_step.find(round_step) != next_votes_tally_by_round_step.end()) {
-        next_votes_tally_by_round_step[round_step] += 1;
-      } else {
-        next_votes_tally_by_round_step[round_step] = 1;
-      }
-    }
-  }
-
-  for (auto const &rs_votes : next_votes_tally_by_round_step) {
-    if (rs_votes.second >= TWO_T_PLUS_ONE) {
-      std::vector<Vote> next_votes_for_round_step = getVotesOfTypeFromVotesForRoundAndStep_(
-          next_vote_type, votes_, rs_votes.first.first, rs_votes.first.second, std::make_pair(NULL_BLOCK_HASH, false));
-      auto voted_block_hash_with_next_votes = blockWithEnoughVotes_(next_votes_for_round_step);
-      if (voted_block_hash_with_next_votes.enough) {
-        LOG(log_dg_) << "Found sufficient next votes in round " << rs_votes.first.first << ", step "
-                     << rs_votes.first.second << ", PBFT 2t+1 " << TWO_T_PLUS_ONE;
-        // Update next votes
-        previous_round_next_votes_->updateNextVotes(voted_block_hash_with_next_votes.votes, TWO_T_PLUS_ONE);
-        auto next_votes = previous_round_next_votes_->getNextVotes();
-
-        auto batch = db_->createWriteBatch();
-        db_->addPbft2TPlus1ToBatch(rs_votes.first.first, TWO_T_PLUS_ONE, batch);
-        db_->addNextVotesToBatch(rs_votes.first.first, next_votes, batch);
-        if (round > 1) {
-          db_->removeNextVotesToBatch(round - 1, batch);
-        }
-        db_->commitWriteBatch(batch);
-
-        return rs_votes.first.first + 1;
-      }
-    }
-  }
-
-  return round;
-}
-
-// Assumption is that all votes are in the same round, step and of same type
-votesBundle PbftManager::blockWithEnoughVotes_(std::vector<Vote> const &votes) const {
-  if (votes.empty()) {
-    return votesBundle();
-  }
-
-  // <voted block hash, votes>
-  std::map<blk_hash_t, std::vector<Vote>> blockhash_votes_map;
-  auto vote_type = votes[0].getType();
-  auto vote_round = votes[0].getRound();
-  auto vote_step = votes[0].getStep();
-
-  for (Vote const &v : votes) {
-    if (v.getType() != vote_type) {
-      LOG(log_er_) << "Vote has a different type with " << unsigned(vote_type) << ". VOTE: " << v;
-      assert(false);
-    } else if (v.getRound() != vote_round) {
-      LOG(log_er_) << "Vote has a different round with " << vote_round << ". VOTE: " << v;
-      assert(false);
-    } else if (v.getStep() != vote_step) {
-      LOG(log_er_) << "Next phase vote has a different step with " << vote_step << ". VOTE: " << v;
-      assert(false);
-    }
-
-    auto voted_block_hash = v.getBlockHash();
-    if (blockhash_votes_map.count(voted_block_hash)) {
-      blockhash_votes_map[voted_block_hash].emplace_back(v);
-    } else {
-      std::vector<Vote> voted_block_hash_votes{v};
-      blockhash_votes_map[voted_block_hash] = voted_block_hash_votes;
-    }
-
-    for (auto const &blockhash_votes : blockhash_votes_map) {
-      if (blockhash_votes.second.size() == TWO_T_PLUS_ONE) {
-        LOG(log_dg_) << "Find voted block hash " << blockhash_votes.first << " vote type " << unsigned(vote_type)
-                     << " in round " << vote_round << " step " << vote_step << " has " << blockhash_votes.second.size()
-                     << " votes";
-        return votesBundle(true, blockhash_votes.first, blockhash_votes.second);
-      } else {
-        LOG(log_tr_) << "Don't have enough votes. voted block hash " << blockhash_votes.first << " vote type "
-                     << unsigned(vote_type) << " for round " << vote_round << " step " << vote_step << " has "
-                     << blockhash_votes.second.size() << " votes (2TP1 = " << TWO_T_PLUS_ONE << ")";
-      }
-    }
-  }
-
-  return votesBundle();
-}
-
-std::vector<Vote> PbftManager::getVotesOfTypeFromVotesForRoundAndStep_(PbftVoteTypes vote_type,
-                                                                       std::vector<Vote> &votes, uint64_t round,
-                                                                       size_t step,
-                                                                       std::pair<blk_hash_t, bool> blockhash) {
-  std::vector<Vote> votes_of_requested_type;
-  std::copy_if(votes.begin(), votes.end(), std::back_inserter(votes_of_requested_type),
-               [vote_type, round, step, blockhash](Vote const &v) {
-                 return (v.getType() == vote_type && v.getRound() == round && v.getStep() == step &&
-                         (blockhash.second == false || blockhash.first == v.getBlockHash()));
-               });
-
-  return votes_of_requested_type;
-}
-
-Vote PbftManager::generateVote(blk_hash_t const &blockhash, PbftVoteTypes type, uint64_t round, size_t step,
+std::shared_ptr<Vote> PbftManager::generateVote(blk_hash_t const &blockhash, PbftVoteTypes type, uint64_t round, size_t step,
                                size_t weighted_index) {
   // sortition proof
   VrfPbftMsg msg(type, round, step, weighted_index);
   VrfPbftSortition vrf_sortition(vrf_sk_, msg);
   Vote vote(node_sk_, vrf_sortition, blockhash);
 
-  return vote;
+  return std::shared_ptr<Vote>(vote);
 }
 
 size_t PbftManager::placeVote_(taraxa::blk_hash_t const &blockhash, PbftVoteTypes vote_type, uint64_t round,
@@ -1390,8 +1272,10 @@ std::vector<std::vector<uint>> PbftManager::createMockTrxSchedule(
 std::pair<blk_hash_t, bool> PbftManager::identifyLeaderBlock_(std::vector<Vote> const &votes) {
   auto round = getPbftRound();
   LOG(log_dg_) << "Into identify leader block, in round " << round;
+
   // each leader candidate with <vote_signature_hash, pbft_block_hash>
   std::vector<std::pair<vrf_output_t, blk_hash_t>> leader_candidates;
+
   for (auto const &v : votes) {
     if (v.getRound() == round && v.getType() == propose_vote_type) {
       // We should not pick any null block as leader (proposed when
@@ -1821,8 +1705,8 @@ std::shared_ptr<PbftBlock> PbftManager::getUnfinalizedBlock_(blk_hash_t const &b
 void PbftManager::countVotes_() {
   auto round = getPbftRound();
   while (!monitor_stop_) {
-    auto verified_votes = vote_mgr_->getVerifiedVotes();
-    auto unverified_votes = vote_mgr_->getUnverifiedVotes();
+    auto verified_votes = db_->getVerifiedVotes();
+    auto unverified_votes = db_->getUnverifiedVotes();
     std::vector<Vote> votes;
     votes.reserve(verified_votes.size() + unverified_votes.size());
     votes.insert(votes.end(), std::make_move_iterator(verified_votes.begin()),
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1796,8 +1796,8 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
 
   // Check cert vote matches
   for (auto const &vote : sync_block.first.cert_votes) {
-    if (vote.getBlockHash() != pbft_block_hash) {
-      LOG(log_er_) << "Invalid cert votes block hash " << vote.getBlockHash() << " instead of " << pbft_block_hash
+    if (vote->getBlockHash() != pbft_block_hash) {
+      LOG(log_er_) << "Invalid cert votes block hash " << vote->getBlockHash() << " instead of " << pbft_block_hash
                    << " from peer " << sync_block.second.abridged() << " received, stop syncing.";
       clearSyncBlockQueue();
       net->handleMaliciousSyncPeer(sync_queue_.front().second);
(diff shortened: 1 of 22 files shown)


src/consensus/vote.cpp
```diff
@@ -368,12 +382,10 @@ std::vector<Vote> VoteManager::getVerifiedVotes(uint64_t const pbft_round, size_
   for (auto const& v : future_unverifiable_votes) {
     removeUnverifiedVote(v.getRound(), v.getHash());
   }
-
-  return getVerifiedVotes();
 }
 
 // cleanup votes < pbft_round
-void VoteManager::cleanupVotes(uint64_t pbft_round) {
+void VoteManager::cleanupVotes(uint64_t const& pbft_round) {
```
MatusKysel, line 388: please remove `const&` it is not necessary and it could be slower then just passing a copy of uint64_t

src/consensus/vote.cpp
```diff
@@ -453,12 +467,11 @@ void VoteManager::cleanupVotes(uint64_t pbft_round) {
   for (auto const& v_hash : remove_verified_votes_hash) {
     db_->removeVerifiedVoteToBatch(v_hash, batch);
   }
-
   db_->commitWriteBatch(batch);
 }
 
-bool VoteManager::voteValidation(taraxa::Vote& vote, size_t const dpos_total_votes_count,
-                                 size_t const sortition_threshold) const {
+bool VoteManager::voteValidation(taraxa::Vote& vote, size_t const& dpos_total_votes_count,
```
MatusKysel, line 473: same here, there is not need for const & for size_t args

src/network/rpc/Test.cpp
```diff
@@ -329,15 +329,16 @@ Json::Value Test::get_votes(const Json::Value & /*param1*/) {
   Json::Value res;
   try {
     if (auto node = full_node_.lock()) {
-      std::shared_ptr<VoteManager> vote_mgr = node->getVoteManager();
+      auto db = node->getDB();
 
-      auto verified_votes = vote_mgr->getVerifiedVotes();
-      auto unverified_votes = vote_mgr->getUnverifiedVotes();
+      auto verified_votes = db->getVerifiedVotes();
```
MatusKysel, line 334: I do not understand why we are now get all votes from DB and not from vote_mgr. DB is much slower as vote_mgr keeps everything in memory

gaoqi8649: Since we don't keep verified votes in vote_mgr anymore, and just keep verified votes hash in vote_mgr for saving memory

src/consensus/vote.cpp
gaoqi8649: Yeah, you right. Fixed, thanks

src/consensus/vote.cpp
gaoqi8649: Fixed

src/consensus/pbft_manager.cpp
```diff
@@ -1691,8 +1574,8 @@ bool PbftManager::giveUpNextVotedBlock_() {
 void PbftManager::countVotes_() {
```
JakubFornadel, line 1574: what do we use this function for ?

src/consensus/vote.cpp
```diff
@@ -163,15 +158,15 @@ void VoteManager::addUnverifiedVotes(std::vector<Vote> const& votes) {
   }
 }
 
-void VoteManager::removeUnverifiedVote(uint64_t const& pbft_round, vote_hash_t const& vote_hash) {
+void VoteManager::removeUnverifiedVote(uint64_t const pbft_round, vote_hash_t const& vote_hash) {
```
JakubFornadel, line 161: this makes a copy of uint64_t so const does not make any sense here

src/consensus/vote.cpp
```diff
@@ -213,44 +208,66 @@ uint64_t VoteManager::getVerifiedVotesSize() const {
...
-      upgradeLock_ locked(lock);
-      verified_votes_[pbft_round][hash] = vote;
     } else {
-      std::unordered_map<vote_hash_t, Vote> votes{std::make_pair(hash, vote)};
+      std::unordered_set<vote_hash_t> votes({hash});
+      std::unordered_map<blk_hash_t, std::unordered_set<vote_hash_t>> voted_values{std::make_pair(voted_value, votes)};
+
       upgradeLock_ locked(lock);
-      verified_votes_[pbft_round] = votes;
+      verified_votes_[round][step] = voted_values;
```
JakubFornadel, line 251: std::move(vited_values)

src/consensus/vote.cpp
```diff
@@ -213,44 +208,66 @@ uint64_t VoteManager::getVerifiedVotesSize() const {
...
-    upgradableLock_ lock(verified_votes_access_);
-    std::map<uint64_t, std::unordered_map<vote_hash_t, Vote>>::const_iterator found_round =
-        verified_votes_.find(pbft_round);
-    if (found_round != verified_votes_.end()) {
-      std::unordered_map<vote_hash_t, Vote>::const_iterator found_vote = found_round->second.find(hash);
-      if (found_vote != found_round->second.end()) {
-        LOG(log_dg_) << "Vote " << hash << " is in verified map already";
-        return;
+  upgradableLock_ lock(verified_votes_access_);
+  if (verified_votes_.count(round)) {
```
JakubFornadel, line 229: how about using if statement with initializer and find instead of count:
`if (auto verified_votes_it = verified_votes_.find(round); verified_votes_it != verified_votes_.end())`

This way you dont need to always lookup roung by `verified_votes_[round]`, but you will use  verified_votes_it. Same logic applies for step, voted_value....

src/consensus/vote.cpp
```diff
@@ -213,44 +208,66 @@ uint64_t VoteManager::getVerifiedVotesSize() const {
...
+
+        } else {
+          upgradeLock_ locked(lock);
+          verified_votes_[round][step][voted_value].insert(hash);
+        }
+      } else {
+        std::unordered_set<vote_hash_t> votes({hash});
+
+        upgradeLock_ locked(lock);
+        verified_votes_[round][step][voted_value] = votes;
```
JakubFornadel, line 244: std::move(votes)

src/consensus/vote.cpp
```diff
@@ -213,44 +208,66 @@ uint64_t VoteManager::getVerifiedVotesSize() const {
...
+
       upgradeLock_ locked(lock);
-      verified_votes_[pbft_round] = votes;
+      verified_votes_[round][step] = voted_values;
     }
+  } else {
+    std::unordered_set<vote_hash_t> votes({hash});
+    std::unordered_map<blk_hash_t, std::unordered_set<vote_hash_t>> voted_values{std::make_pair(voted_value, votes)};
+    std::map<size_t, std::unordered_map<blk_hash_t, std::unordered_set<vote_hash_t>>> steps{
+        std::make_pair(step, voted_values)};
```
JakubFornadel, line 257: std::move....

src/consensus/vote.cpp
```diff
@@ -368,12 +382,10 @@ std::vector<Vote> VoteManager::getVerifiedVotes(uint64_t const pbft_round, size_
   for (auto const& v : future_unverifiable_votes) {
     removeUnverifiedVote(v.getRound(), v.getHash());
   }
-
-  return getVerifiedVotes();
 }
 
 // cleanup votes < pbft_round
-void VoteManager::cleanupVotes(uint64_t pbft_round) {
+void VoteManager::cleanupVotes(uint64_t const pbft_round) {
```
JakubFornadel, line 388: const ?

src/consensus/vote.cpp
```diff
@@ -478,8 +491,10 @@ bool VoteManager::voteValidation(taraxa::Vote& vote, size_t const dpos_total_vot
   return true;
 }
 
-bool VoteManager::pbftBlockHasEnoughValidCertVotes(PbftBlockCert& pbft_block_and_votes, size_t dpos_total_votes_count,
-                                                   size_t sortition_threshold, size_t pbft_2t_plus_1) const {
+bool VoteManager::pbftBlockHasEnoughValidCertVotes(PbftBlockCert& pbft_block_and_votes,
+                                                   size_t const dpos_total_votes_count,
```
JakubFornadel, line 495: const ?

src/consensus/vote.cpp
```diff
@@ -549,6 +564,106 @@ std::string VoteManager::getJsonStr(std::vector<Vote> const& votes) {
...
+    sharedLock_ lock(verified_votes_access_);
+    proposal_voted_value_map = verified_votes_[pbft_round][1];
+  }
+
+  std::vector<Vote> proposal_votes;
+  for (auto const& voted_value : proposal_voted_value_map) {
+    for (auto const& v_hash : voted_value.second) {
+      auto vote = db_->getVerifiedVote(v_hash);
+      if (vote) {
+        proposal_votes.emplace_back(*vote);
```
JakubFornadel, line 579: why would copy votes here ? Is there some reason why not to return std::vector<std::shared_ptr<Vote>> ???

src/consensus/vote.cpp
```diff
@@ -549,6 +564,106 @@ std::string VoteManager::getJsonStr(std::vector<Vote> const& votes) {
   return ptroot.toStyledString();
 }
 
+std::vector<Vote> VoteManager::getProposalVotes(uint64_t const pbft_round) {
+  std::unordered_map<blk_hash_t, std::unordered_set<vote_hash_t>> proposal_voted_value_map;
+  {
+    sharedLock_ lock(verified_votes_access_);
+    proposal_voted_value_map = verified_votes_[pbft_round][1];
```
JakubFornadel, line 571: why would you create a copy of whole std::unordered_map<blk_hash_t, std::unordered_set<vote_hash_t>> ? Is that because you did not want to lock verified_votes_access_ for the whole processing functionality below ? Because this might a lot of memory to copy

src/consensus/vote.cpp
```diff
@@ -549,6 +564,106 @@ std::string VoteManager::getJsonStr(std::vector<Vote> const& votes) {
...
+  std::vector<vote_hash_t> votes_hash;
+
+  {
+    sharedLock_ lock(verified_votes_access_);
+    if (verified_votes_.count(round)) {
+      if (verified_votes_[round].count(step)) {
+        for (auto const& voted_value : verified_votes_[round][step]) {
+          if (voted_value.second.size() >= two_t_plus_one) {
+            voted_block_hash = voted_value.first;
+            votes_hash.reserve(two_t_plus_one);
```
JakubFornadel, line 601: I guess this is some mistake ? Why call reserve inside for ?

JakubFornadel, line 601: I guess this is some mistake. Why would you reserve for the same size over and over again inside for loop ?

src/consensus/vote.cpp
```diff
@@ -549,6 +564,106 @@ std::string VoteManager::getJsonStr(std::vector<Vote> const& votes) {
...
+  if (!votes_hash.empty()) {
+    LOG(log_nf_) << "Found enough " << votes_hash.size() << " votes at voted value " << voted_block_hash
+                 << " for round " << round << " step " << step;
+
+    std::vector<Vote> votes;
+    votes.reserve(two_t_plus_one);
+    for (auto const& v_hash : votes_hash) {
+      auto vote = db_->getVerifiedVote(v_hash);
+      assert(vote);
+      votes.emplace_back(*vote);
```
JakubFornadel, line 627: same question as here: 
https://github.com/Taraxa-project/taraxa-node/pull/985/files#r680926590

src/consensus/vote.hpp
```diff
@@ -251,4 +191,82 @@ class NextVotesForPreviousRound {
   LOG_OBJECTS_DEFINE
 };
 
+struct votesBundle {
```
JakubFornadel, line 194: types start with capital letter

src/consensus/vote.hpp
```diff
@@ -251,4 +191,82 @@ class NextVotesForPreviousRound {
...
+  bool enough;
+  blk_hash_t voted_block_hash;
+  std::vector<Vote> votes;  // exactly 2t+1 votes
+
+  votesBundle() : enough(false), voted_block_hash(blk_hash_t(0)) {}
+  votesBundle(bool const enough_, blk_hash_t const& voted_block_hash_, std::vector<Vote> const& votes_)
```
JakubFornadel, line 200: - uff - bool const ?  
- you just exchanged formatting here, "enough_" with "_" should be name of class member, not ctor parameter...

src/consensus/pbft_manager.cpp
```diff
@@ -655,19 +652,15 @@ bool PbftManager::stateOperations_() {
 
   if (is_syncing_()) return true;
 
-  // Get votes
-  votes_ = vote_mgr_->getVerifiedVotes(round, sortition_threshold_, getDposTotalVotesCount(),
-                                       [this](auto const &addr) { return dpos_eligible_vote_count_(addr); });
-
-  LOG(log_tr_) << "There are " << votes_.size() << " total votes in round " << round;
+  // Periodically verify unverified votes
+  vote_mgr_->verifyVotes(round, sortition_threshold_, getDposTotalVotesCount(),
```
JakubFornadel, line 656: so, if I remember correctly one of the problems I pointed out is that inside `getVerifiedVotes` was called `getUnverifiedVotes()`, which would always make copies of votes. Because this is called very often it would reallocate Terabytes of memory evry hour. Now it is just moved to the `verifyVotes` if I see correctly. What happened with this PR, which was supposed to fix it ? 
https://github.com/Taraxa-project/taraxa-node/pull/927

JakubFornadel (changes requested):
There is a lot of places in new code that should be improved in terms of efficiency. I think we should really come up with a solution to avoid making copies of votes - maybe using shared_ptrs or references ? 

+ I dont see where is the perf. gain/ improvement here ? Lets talk about it on call...

src/consensus/pbft_manager.cpp
gaoqi8649: We use this function to debug number of votes for each step

src/consensus/vote.cpp
gaoqi8649: Fixed, thanks

src/consensus/vote.cpp
gaoqi8649: fixed, thanks

src/consensus/vote.cpp
gaoqi8649: Fixed, thanks

src/consensus/vote.cpp
gaoqi8649: fixed, thanks

src/consensus/vote.cpp
gaoqi8649: fixed, thanks

src/consensus/vote.cpp
gaoqi8649: fixed, thanks

src/consensus/vote.cpp
gaoqi8649: fixed, thanks

src/consensus/vote.cpp
gaoqi8649: Yeah, because don't want to lock verified_votes_access_.
This will not a lot of memory to copy. Since only allow to vote one single vote in proposal phase per each node. All 3 boot nodes occupied 2t+1 votes, 3 boot nodes only vote 3 votes at proposal phase. And there are only t-1 nodes will get sortition in maximum. So the max number of proposal votes should be t - 1 + 3 = t + 2 = 1/3 committee + 2 = 335
Even after remove centralized by boot nodes, the max number of proposal votes for each round is committee size, that is 1000. So this is not a lot of memory

src/consensus/vote.cpp
gaoqi8649: Because only need 2t+1 number of votes to make consensus

src/consensus/vote.hpp
gaoqi8649: Fixed start capital letter, Thanks!
In my opinion, struct is better for creating value types, and class is better for making use of features. I prefer struct here

src/network/rpc/Test.cpp
gaoqi8649: Move back verified votes in memory, fixed

src/consensus/vote.cpp
gaoqi8649: fixed, thanks

src/consensus/vote.cpp
gaoqi8649: Since moved back vote struct in memory, don't need retrieve vote from DB. Remove copy table

src/consensus/vote.cpp
gaoqi8649: fixed, thanks

gaoqi8649:
> There is a lot of places in new code that should be improved in terms of efficiency. I think we should really come up with a solution to avoid making copies of votes - maybe using shared_ptrs or references ?
> 
> * I dont see where is the perf. gain/ improvement here ? Lets talk about it on call...

I fixed the comments on the review board. Please review again, Thanks!

src/consensus/pbft_manager.cpp
```diff
@@ -843,7 +834,8 @@ void PbftManager::identifyBlock_() {
 
   if (round == 1 || (round >= 2 && giveUpNextVotedBlock_())) {
     // Identity leader
-    std::pair<blk_hash_t, bool> leader_block = identifyLeaderBlock_(votes_);
+    auto proposal_votes = vote_mgr_->getProposalVotes(round);
+    std::pair<blk_hash_t, bool> leader_block = identifyLeaderBlock_(proposal_votes);
```
MatusKysel, line 838: std::move(proposal_votes)

src/consensus/vote.cpp
```diff
@@ -100,6 +105,7 @@ VoteManager::~VoteManager() { daemon_->join(); }
 void VoteManager::setNetwork(std::weak_ptr<Network> network) { network_ = move(network); }
 
 void VoteManager::retreieveVotes_() {
+  LOG(log_si_) << "Retrieve unverified votes from DB";
```
MatusKysel, line 108: are you sure about this level?

src/consensus/vote.cpp
```diff
@@ -230,41 +226,92 @@ uint64_t VoteManager::getUnverifiedVotesSize() const {
   return size;
 }
 
+std::vector<Vote> VoteManager::getVerifiedVotes() {
+  std::vector<Vote> votes;
```
MatusKysel, line 230: isn't this same thing, that @JakubFornadel pointed out where we are doing a lot of allocations? Why we are not using shared_ptr also here?

src/consensus/vote.cpp
```diff
@@ -570,6 +614,82 @@ std::string VoteManager::getJsonStr(std::vector<Vote> const& votes) {
...
+    for (auto const& v : voted_value.second) {
+      proposal_votes.emplace_back(std::make_shared<Vote>(v.second));
+    }
+  }
+
+  return proposal_votes;
+}
+
+VotesBundle VoteManager::getVotesBundleByRoundAndStep(uint64_t round, size_t step, size_t two_t_plus_one) {
+  std::vector<Vote> votes;
```
MatusKysel, line 631: here you again copying votes, can we use shared_ptr here ?

src/consensus/vote.hpp
```diff
@@ -191,9 +259,12 @@ class VoteManager {
   using upgradableLock_ = boost::upgrade_lock<boost::shared_mutex>;
   using upgradeLock_ = boost::upgrade_to_unique_lock<boost::shared_mutex>;
 
-  // <pbft_round, <vote_hash, vote>>
+  // <pbft round, <vote hash, vote>>
   std::map<uint64_t, std::unordered_map<vote_hash_t, Vote>> unverified_votes_;
-  std::map<uint64_t, std::unordered_map<vote_hash_t, Vote>> verified_votes_;
+
+  // <PBFT round, <PBFT step, <voted value, <vote hash, vote>>>>
+  std::map<uint64_t, std::map<size_t, std::unordered_map<blk_hash_t, std::unordered_map<vote_hash_t, Vote>>>>
```
MatusKysel, line 266: do we need to have this steps and round in order? we are accessing it by step/round, so I would suggest  unordered_map for performance. Even for `unverified_votes_`

src/consensus/pbft_manager.cpp
JakubFornadel: yes we did and we all agreed that we should not be making copies of votes here but we still do so what did you want to say by "Have discussed at the meeting" in terms of solving this ?

src/consensus/vote.cpp
```diff
@@ -570,6 +614,82 @@ std::string VoteManager::getJsonStr(std::vector<Vote> const& votes) {
...
+
+  sharedLock_ lock(verified_votes_access_);
+  auto found_round_it = verified_votes_.find(round);
+  if (found_round_it != verified_votes_.end()) {
+    auto found_step_it = found_round_it->second.find(step);
+    if (found_step_it != found_round_it->second.end()) {
+      for (auto const& voted_value : found_step_it->second) {
+        if (voted_value.second.size() >= two_t_plus_one) {
+          auto voted_block_hash = voted_value.first;
+          votes.reserve(two_t_plus_one);
```
JakubFornadel, line 641: I already asked but this reserve was somewhere else so for reminding the conversation I asked:

I guess this is some mistake. Why would you reserve always for the same size inside for loop ?

@gaoqi8649 gaoqi8649 6 days ago Author Member
Because only need 2t+1 number of votes to make consensus

So again it does not make any sense to reserve vector size in loop. What are you trying to achieve by this ?

src/consensus/vote.cpp
```diff
@@ -570,6 +614,82 @@ std::string VoteManager::getJsonStr(std::vector<Vote> const& votes) {
...
+          size_t count = 0;
+          // Only copy 2t+1 votes
+          while (count != two_t_plus_one) {
+            votes.emplace_back(it->second);
+            it++;
+            count++;
+          }
+          LOG(log_nf_) << "Found enough " << votes.size() << " votes at voted value " << voted_block_hash
+                       << " for round " << round << " step " << step;
+          return VotesBundle(true, voted_block_hash, votes);
```
JakubFornadel, line 653: again you are copying the whole votes vector here, which will copy all the votes inside. Can you please go over you code (just to make sure) that you are using move anywhere you can to avoid useless copies like this ?

src/consensus/pbft_manager.cpp
```diff
@@ -1284,16 +1165,18 @@ std::vector<std::vector<uint>> PbftManager::createMockTrxSchedule(
   return blocks_trx_modes;
 }
 
-std::pair<blk_hash_t, bool> PbftManager::identifyLeaderBlock_(std::vector<Vote> const &votes) {
+std::pair<blk_hash_t, bool> PbftManager::identifyLeaderBlock_(std::vector<std::shared_ptr<Vote>> votes) {
```
JakubFornadel, line 1168: even worse than before, instead of const ref to vector now it is a copy of vector

src/consensus/vote.cpp
```diff
@@ -570,6 +614,82 @@ std::string VoteManager::getJsonStr(std::vector<Vote> const& votes) {
   return ptroot.toStyledString();
 }
 
+std::vector<std::shared_ptr<Vote>> VoteManager::getProposalVotes(uint64_t pbft_round) {
+  std::vector<std::shared_ptr<Vote>> proposal_votes;
+
+  sharedLock_ lock(verified_votes_access_);
+  for (auto const& voted_value : verified_votes_[pbft_round][1]) {
+    for (auto const& v : voted_value.second) {
+      proposal_votes.emplace_back(std::make_shared<Vote>(v.second));
```
JakubFornadel, line 623: I know we were talking a lot about shared pointers and references but I am not sure if you completely understand what we want to achieve by that. What do you think this vector of shared_ptrs solves in terms of efficiency if you are still making a copy of vote here std::make_shared<Vote>(v.second) as you stored Votes objects in verified_votes_ ?

src/consensus/pbft_manager.cpp
MatusKysel: is it better to call `vote_mgr_->getProposalVotes(round)` inside this function as it is used only once like this:
```    
auto proposal_votes = vote_mgr_->getProposalVotes(round);
std::pair<blk_hash_t, bool> leader_block = identifyLeaderBlock_(proposal_votes);
```

src/consensus/vote.cpp
gaoqi8649: Sorry, I don't understand how to do here.
I thought std::make_shared<Vote>(v.second) is makeing a shared pointer to the vote object. 
Could you please give me an example? How should I do here? Thanks

src/consensus/vote.cpp
gaoqi8649: yeah, since retreieveVotes_ only run in the beginning of node start. The log only print one time per each reboot. I set it in log_si_ for checking the thread is running in the beginning

src/consensus/vote.hpp
gaoqi8649: The round and step need in order.
For round, if a bootstrapping node join. We need to check the round in descending to find the current round in the network. If we can find enough votes in round 1000, then we could jump to round 1000 directly.
For step, we need to check in ascending to check enough votes to go to next round. For example, like node0 at step 5 and node1 at step 100. So we will get enough votes at step 5 because node0 hasn't vote in the further step

src/consensus/vote.cpp
gaoqi8649: This will only reserve one time and will return VotesBundle(true, voted_block_hash, votes)

src/consensus/pbft_manager.cpp
gaoqi8649: Fixed by future_unverifiable_votes, verified_votes and getUnverifiedVotes return std::vector<std::shared_ptr<Vote>> in verifyVotes

src/consensus/pbft_manager.cpp
gaoqi8649: Move vote_mgr_->getProposalVotes(round) in identifyLeaderBlock_ function

src/consensus/vote.cpp
gaoqi8649: Fixed, thanks

src/consensus/pbft_manager.cpp
gaoqi8649: Moved vote_mgr_->getProposalVotes(round) inside identifyLeaderBlock_, thanks

src/consensus/vote.cpp
```diff
@@ -220,48 +215,101 @@ uint64_t VoteManager::getUnverifiedVotesSize() const {
...
-  for (it = unverified_votes_.begin(); it != unverified_votes_.end(); ++it) {
+  for (auto it = unverified_votes_.begin(); it != unverified_votes_.end(); ++it) {
     size += it->second.size();
   }
 
   return size;
 }
 
+std::vector<std::shared_ptr<Vote>> VoteManager::getVerifiedVotes() {
+  std::vector<std::shared_ptr<Vote>> votes;
```
MatusKysel, line 226: reserve size based on `verified_votes_` size

src/consensus/vote.cpp
```diff
@@ -545,29 +591,105 @@ bool VoteManager::pbftBlockHasEnoughValidCertVotes(SyncBlock& pbft_block_and_vot
...
+    ptvote["step"] = Json::Value::UInt64(v->getStep());
     ptvotes.append(ptvote);
   }
   ptroot["votes"] = ptvotes;
 
   return ptroot.toStyledString();
 }
 
+std::vector<std::shared_ptr<Vote>> VoteManager::getProposalVotes(uint64_t pbft_round) {
+  std::vector<std::shared_ptr<Vote>> proposal_votes;
```
MatusKysel, line 616: same here, reserve size based on `verified_votes_[pbft_round][1]` size

MatusKysel (approved):

src/consensus/pbft_manager.cpp
```diff
@@ -1059,131 +1050,19 @@ void PbftManager::secondFinish_() {
...
-                               size_t weighted_index) {
+std::shared_ptr<Vote> PbftManager::generateVote(blk_hash_t const &blockhash, PbftVoteTypes type, uint64_t round,
+                                                size_t step, size_t weighted_index) {
   // sortition proof
   VrfPbftMsg msg(type, round, step, weighted_index);
   VrfPbftSortition vrf_sortition(vrf_sk_, msg);
   Vote vote(node_sk_, vrf_sortition, blockhash);
 
-  return vote;
+  return std::make_shared<Vote>(vote);
```
JakubFornadel, line 1060: retturn std::make_shared<Vote>(node_sk_, vrf_sortition, blockhash) and delete line 1058

src/consensus/vote.cpp
```diff
@@ -95,102 +100,92 @@ VoteManager::VoteManager(addr_t node_addr, std::shared_ptr<DbStorage> db, std::s
...
-
     // Rebroadcast our own next votes in case we were partitioned...
-    if (v.getVoterAddr() == node_addr_ && v.getStep() >= FIRST_FINISH_STEP &&
+    if (v->getVoterAddr() == node_addr_ && v->getStep() >= FIRST_FINISH_STEP &&
         db_->getPbftMgrField(PbftMgrRoundStep::PbftStep) > EXTENDED_PARTITION_STEPS) {
-      vector<Vote> votes = {v};
+      std::vector<std::shared_ptr<Vote>> votes = {v};
       if (auto net = network_.lock()) {
-        net->onNewPbftVotes(move(votes));
+        net->onNewPbftVotes(std::move(votes));
```
JakubFornadel, line 131: Why is this called inside for loop and not after that loop is done ?

src/consensus/vote.cpp
gaoqi8649: We don't have variable "size" for record unverified/verifed votes in class, since votes will change very often.
And getVerifiedVotesSize() need go through PBFT rounds, steps, and voted value for counting verfied_votes size. And I think that is not good for getting size first and reserve it

gaoqi8649: Fixed, thanks

src/consensus/vote.cpp
gaoqi8649: Fixed, thanks

src/consensus/pbft_manager.cpp
gaoqi8649: Fixed, thanks

src/consensus/vote.cpp
gaoqi8649: When nodes get partition, that don't need all verified votes. Partitioned nodes will only need 2t+1 next votes in one step  to jump to the correct PBFT round. Sending them one by one will be faster for partitioned nodes jump to correct PBFT round.
This need to improve later for sending votes in bundle: https://github.com/Taraxa-project/taraxa-node/issues/1158

gaoqi8649:
> There is a lot of places in new code that should be improved in terms of efficiency. I think we should really come up with a solution to avoid making copies of votes - maybe using shared_ptrs or references ?
> 
> * I dont see where is the perf. gain/ improvement here ? Lets talk about it on call...

I think I have fixed all review comments. Please review again. Thanks!

src/consensus/vote.cpp
MatusKysel: no like this, this is just overhead. after the lock you can do `votes.reserve(verified_votes_.size());`

src/consensus/vote.cpp
MatusKysel: move it under to lock

gaoqi8649: yeah, fixed

MatusKysel (approved):

JakubFornadel (approved):

PR #986, opened 2021-07-19
fix(logs): fix automatic log clean-up

MatusKysel:
## Purpose

because we did not specify `target` directory log cleaning up did not work


JakubFornadel:
Does this keep logs folder at output.max_size in a way that once we hit that limit it starts to delete oldest files ?

MatusKysel:
> Does this keep logs folder at output.max_size in a way that once we hit that limit it starts to delete oldest files ?

yes

JakubFornadel (approved):

kstdl (approved):

PR #1023, opened 2021-07-27
Add information about prnet to commetns in pr when deployed

agrebin:
## Purpose

After deploying a prnet , pipeline shoudl return a message with information about the prnet deployed

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -311,13 +311,15 @@ jobs:
       - pr_comment:
          when: on_success
          comment: "\
-           <p>PR NET pr-<< pipeline.parameters.pr_number >> Successfully deployed!</p>\ 
-           <p>explorer=https://explorer-pr-<< pipeline.parameters.pr_number >> </p> \
-           <p>RPC=rpc-pr-<< pipeline.parameters.pr_number >>.prnet.taraxa.io </p>\
-           <p>Boot Nodes</p>\
-           <p>Boot Node 1=${PRNET_IPS[0]}</p>\ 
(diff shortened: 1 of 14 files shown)


.circleci/config.yml
```diff
@@ -311,7 +310,14 @@ jobs:
          pr-name: << pipeline.parameters.pr_number >>
       - pr_comment:
          when: on_success
-         comment: "Deployed prnet successfully"
+         comment: "\
+           <p>PR NET pr-<< pipeline.parameters.pr_number >> Successfully deployed!</p>\ 
+           <p>explorer=https://explorer-pr-<< pipeline.parameters.pr_number >> </p> \
```
sfrique, line 315: We are missing the last part of the domain no? 
I saw on the previous comments that the link was clickable, I think it would be best if it was possible to click.
like: https://explorer-pr-1023.prnet.taraxa.io

.circleci/config.yml
```diff
@@ -311,7 +310,14 @@ jobs:
...
+           <p>RPC=rpc-pr-<< pipeline.parameters.pr_number >>.prnet.taraxa.io </p>\
```
sfrique, line 316: The RPC is also an ingress and has HTTPS on front
but I do not know if the ledger team wants this var with HTTPS on front, please check with them

agrebin: /deploy

sfrique (approved):

rattrap (approved):

PR #1024, opened 2021-07-27
perf(vote_mgr): retrieve votes from db on launch from separate thread

justinsnapp:
## Purpose

Right now if a node has been stalled or partitioned for a while, upon restart it will block pbft manager consensus while vote manage retrieves votes from the db.   Instead we can asynchronously retrieve the votes and populate them into the vote manager.

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -352,7 +352,10 @@ void PbftManager::setPbftStep(size_t const pbft_step) {
   }
 }
 
-void PbftManager::resetStep_() { setPbftStep(1); }
+void PbftManager::resetStep_() {
+  startingStepInRound_ = 1;
+  setPbftStep(1);
+}
 
 bool PbftManager::resetRound_() {
   bool restart = false;
@@ -462,6 +465,8 @@ void PbftManager::initialState_() {
     LOG(log_er_) << "Unexpected condition at round " << round << " step " << step;
     assert(false);
   }
+  // This is used to offset endtime for second finishing step...
+  startingStepInRound_ = step;
   setPbftStep(step);
   setPbftRound(round);
 
@@ -635,7 +640,8 @@ void PbftManager::loopBackFinishState_() {
   db_->commitWriteBatch(batch);
   next_voted_soft_value_ = false;
   next_voted_null_block_hash_ = false;
-  next_step_time_ms_ = step_ * LAMBDA_ms;
+  assert(step_ >= startingStepInRound_);
+  next_step_time_ms_ = (1 + step_ - startingStepInRound_) * LAMBDA_ms;
   last_step_clock_initial_datetime_ = current_step_clock_initial_datetime_;
   current_step_clock_initial_datetime_ = std::chrono::system_clock::now();
 }
@@ -969,7 +975,8 @@ void PbftManager::secondFinish_() {
   // Odd number steps from 5 are in second finish
   auto round = getPbftRound();
   LOG(log_tr_) << "PBFT second finishing state at step " << step_ << " in round " << round;
-  auto end_time_for_step = (step_ + 1) * LAMBDA_ms - POLLING_INTERVAL_ms;
+  assert(step_ >= startingStepInRound_);
+  auto end_time_for_step = (2 + step_ - startingStepInRound_) * LAMBDA_ms - POLLING_INTERVAL_ms;
 
   updateSoftVotedBlockForThisRound_();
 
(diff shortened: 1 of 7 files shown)


src/consensus/vote.cpp
```diff
@@ -105,8 +120,15 @@ VoteManager::VoteManager(addr_t node_addr, std::shared_ptr<DbStorage> db, std::s
   {
     auto verified_votes = db_->getVerifiedVotes();
```
gaoqi8649, line 121: I remember that retrieve all verified votes from DB that will take a long time.
I don't know if RocksDB support get number of items. Emm... that should not work. Because in the meantime, also write votes into verified votes DB column.  And I think DB writing items are not by order guarantee

gaoqi8649 (approved):
Looks good to me

src/consensus/vote.cpp
```diff
@@ -86,11 +86,26 @@ VoteManager::VoteManager(addr_t node_addr, std::shared_ptr<DbStorage> db, std::s
     : db_(db), pbft_chain_(pbft_chain), final_chain_(final_chain) {
   LOG_OBJECTS_CREATE("VOTE_MGR");
   // Retrieve votes from DB
+  daemon_ = std::make_unique<std::thread>([this]() { retreieveVotes_(); });
+
+  current_period_final_chain_block_hash_ = final_chain_->block_header()->hash;
+}
+
+VoteManager::~VoteManager() { daemon_->join(); }
```
gaoqi8649, line 94: I think we can stop the daemon by end of retrieveVotes_()

justinsnapp: That would be in the same thread..??

src/consensus/vote.cpp
```diff
@@ -86,40 +86,61 @@ VoteManager::VoteManager(addr_t node_addr, std::shared_ptr<DbStorage> db, std::s
...
+
+void VoteManager::retreieveVotes_() {
+  auto unverified_votes = db_->getUnverifiedVotes();
+  auto unverified_votes_itr = unverified_votes.begin();
+  while (unverified_votes_itr != unverified_votes.end()) {
+    {
+      uniqueLock_ lock(unverified_votes_access_);
+      auto cntr = 0;
+      while (cntr++ < 100000 && unverified_votes_itr != unverified_votes.end()) {
```
MatusKysel, line 103: make this 100000 some constexpr class variable

MatusKysel (approved):

src/consensus/vote.cpp
```diff
@@ -114,6 +121,15 @@ void VoteManager::retreieveVotes_() {
   for (auto const& v : verified_votes) {
     auto pbft_round = v.getRound();
     auto hash = v.getHash();
+    
+    //Rebroadcast our own next votes in case we were partitioned...
+    if (v.getVoterAddr() == node_addr_ && v.getStep() > FIRST_FINISH_STEP && db_->getPbftMgrField(PbftMgrRoundStep::PbftStep) > EXTENDED_PARTITION_STEPS) {
```
gaoqi8649, line 126: v.getStep() >= FIRST_FINISH_STEP, should include step 4

justinsnapp: Ok fixed.

src/consensus/pbft_manager.cpp
```diff
@@ -635,7 +640,8 @@ void PbftManager::loopBackFinishState_() {
   db_->commitWriteBatch(batch);
   next_voted_soft_value_ = false;
   next_voted_null_block_hash_ = false;
-  next_step_time_ms_ = step_ * LAMBDA_ms;
+  assert(step_ >= startingStepInRound_);
+  next_step_time_ms_ = (1 + step_ - startingStepInRound_) * LAMBDA_ms;
```
gaoqi8649, line 644: Does step_ always equal to startingStepInRound_? If yes, next_step_time_ms_ always is LAMBDA_ms time

src/consensus/pbft_manager.cpp
```diff
@@ -969,7 +975,8 @@ void PbftManager::secondFinish_() {
   // Odd number steps from 5 are in second finish
   auto round = getPbftRound();
   LOG(log_tr_) << "PBFT second finishing state at step " << step_ << " in round " << round;
-  auto end_time_for_step = (step_ + 1) * LAMBDA_ms - POLLING_INTERVAL_ms;
+  assert(step_ >= startingStepInRound_);
+  auto end_time_for_step = (2 + step_ - startingStepInRound_) * LAMBDA_ms - POLLING_INTERVAL_ms;
```
gaoqi8649, line 979: Same here. If step_ always equal to startingStepInRound_, end_time_for_step is always 2 * LAMBDA_ms - POLLING_INTERVAL_ms

PR #1026, opened 2021-07-28
Mid prio handlers

MatusKysel:


src/network/tarcap/packets_handler/handlers/blocks_packet_handler.cpp
```diff
@@ -0,0 +1,57 @@
...
+  const bool is_final_sync_packet = (*it++).toInt<unsigned>();
+
+  for (; it != packet_rlp.end();) {
+    DagBlock block(*it++);
+    peer_->markBlockAsKnown(block.getHash());
+
+    std::vector<Transaction> new_transactions;
+    for (size_t i = 0; i < block.getTrxs().size(); i++) {
+      Transaction transaction(*it++);
+      new_transactions.push_back(transaction);
```
JakubFornadel, line 27: new_transactions.push_back(transaction) could be made more efficient with std::move(transaction)

src/network/tarcap/packets_handler/handlers/blocks_packet_handler.cpp
```diff
@@ -0,0 +1,57 @@
...
+      status.second.push_back(block.getHash());
+      syncing_state_->requestBlocks(packet_data.from_node_id_, status.second,
+                                    GetBlocksPacketRequestType::MissingHashes);
+      continue;
+    }
+
+    LOG(log_dg_) << "Storing block " << block.getHash().toString() << " with " << new_transactions.size()
+                 << " transactions";
+    if (block.getLevel() > peer_->dag_level_) peer_->dag_level_ = block.getLevel();
+    dag_blk_mgr_->insertBroadcastedBlockWithTransactions(block, new_transactions);
```
JakubFornadel, line 45: when I look at it now, we are probably marking txs and block as known even if received invalid(malicious) block. Maybe we could even return bool flag from  insertBroadcastedBlockWithTransactions and mark block as well as txs as known only if true was returned...

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
```diff
@@ -0,0 +1,230 @@
...
+      return;
+    }
+  }
+
+  const auto transactions_count = packet_rlp.itemCount() - 1;
+  LOG(log_dg_) << "Received NewBlockPacket " << transactions_count;
+
+  std::vector<Transaction> new_transactions;
+  for (size_t i_transaction = 1; i_transaction < transactions_count + 1; i_transaction++) {
+    Transaction transaction(packet_rlp[i_transaction].data().toBytes());
```
JakubFornadel, line 62: this "handlers separation" might be a good task to get rid of all rlp random access like here "packet_rlp[i_transaction]"

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
```diff
@@ -0,0 +1,230 @@
...
+  } else if (peers_state_->test_blocks_.find(hash) == peers_state_->test_blocks_.end() &&
+             block_requestes_set_.count(hash) == 0) {
+    block_requestes_set_.insert(hash);
+    requestBlock(packet_data.from_node_id_, hash);
+  }
+}
+
+inline void DagPacketsHandler::processGetNewBlockPacket(const PacketData &packet_data, const dev::RLP &packet_rlp) {
+  blk_hash_t const hash(packet_rlp[0]);
+  peer_->markBlockAsKnown(hash);
```
JakubFornadel, line 91: sendBlock could potentially throw and exception.... lets move this to the end of  processGetNewBlockPacket function

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+  peer_->markBlockAsKnown(hash);
+  if (block.getLevel() > peer_->dag_level_) peer_->dag_level_ = block.getLevel();
+  onNewBlockReceived(block, new_transactions);
+}
+
+inline void DagPacketsHandler::processNewBlockHashPacket(const PacketData &packet_data, const dev::RLP &packet_rlp) {
+  blk_hash_t const hash(packet_rlp[0]);
+  LOG(log_dg_) << "Received NewBlockHashPacket " << hash.toString();
+  peer_->markBlockAsKnown(hash);
```
JakubFornadel, line 75: I guess even this should be at the end of function

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
```diff
@@ -0,0 +1,230 @@
...
+    new_transactions.push_back(transaction);
+    peer_->markTransactionAsKnown(transaction.getHash());
+  }
+
+  peer_->markBlockAsKnown(hash);
+  if (block.getLevel() > peer_->dag_level_) peer_->dag_level_ = block.getLevel();
+  onNewBlockReceived(block, new_transactions);
+}
+
+inline void DagPacketsHandler::processNewBlockHashPacket(const PacketData &packet_data, const dev::RLP &packet_rlp) {
```
JakubFornadel, line 72: I think that in general we could get rid of whole "NewBlockHashPacket" packet. When I checked it is only send inside "onNewBlockVerified" to "in some way randomized" set of peers that do not have this block already(or at least we did not mark it as known fro them) and I am not sure what is the purpose of it ? If such peer received only block hash but does not have it locally it will instantly request our node to send him full block. I guess it has some purpose but I am not really sure what it is ??? @justinsnapp @mfrankovi you might know ?

src/network/tarcap/packets_handler/handlers/blocks_packet_handler.cpp
MatusKysel: but is there a problem? we mark, that specific block is invalid for specific peer. I do not see a problem there

src/network/tarcap/packets_handler/handlers/blocks_packet_handler.cpp
MatusKysel: there are two places like this I will fix it

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
```diff
@@ -0,0 +1,230 @@
...
+  blk_hash_t const hash(packet_rlp[0]);
+  peer_->markBlockAsKnown(hash);
+  LOG(log_dg_) << "Received GetNewBlockPacket" << hash.toString();
+
+  if (dag_blk_mgr_) {
+    auto block = db_->getDagBlock(hash);
+    if (block) {
+      sendBlock(packet_data.from_node_id_, *block);
+    } else
+      LOG(log_nf_) << "NO NEW PACKET: " << hash.toString();
```
JakubFornadel, line 99: hmm lets have here some proper message, this is very generic and does not say what actually hapenned

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
MatusKysel: this is not an issue actually if you access elements in ascending order it is exactly same as iterator

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
```diff
@@ -0,0 +1,230 @@
...
+namespace taraxa::network::tarcap {
+
+DagPacketsHandler::DagPacketsHandler(std::shared_ptr<PeersState> peers_state,
+                                     std::shared_ptr<SyncingState> syncing_state,
+                                     std::shared_ptr<TransactionManager> trx_mgr,
+                                     std::shared_ptr<DagBlockManager> dag_blk_mgr, std::shared_ptr<DbStorage> db,
+                                     uint16_t network_min_dag_block_broadcast, uint16_t network_max_dag_block_broadcast,
+                                     const addr_t &node_addr)
+    : PacketHandler(std::move(peers_state), node_addr, "DAG_BLOCK_PH"),
+      syncing_state_(syncing_state),
```
JakubFornadel, line 16: all shared pointer were copied into the ctor so should be moved here into the class members, otherwise we make 1 extra unnecessary copy (I know compiler might optimize it), but lets not rely on complier...

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
JakubFornadel: hmm I did not really check but @mfrankovi suggested that when using random access it always iterates from 0 until it gets to the index you want. If true, it does not matter if it is in order or not...

src/network/tarcap/packets_handler/handlers/blocks_packet_handler.cpp
JakubFornadel: we dont mark it as invalid but as known, but yeah maybe we can do that even if the block we received here is invalid. Not 100% sure though

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
MatusKysel: it is even noted in the header, so we should be alright
```
  /// @note if used to access items in ascending order, this is efficient.
  RLP operator[](size_t _i) const;
```

MatusKysel: it is more complicated, it will mark last index so in ascending order is exactly the same as iterator, if you really do random access it will iterate from start

src/network/tarcap/packets_handler/handlers/blocks_packet_handler.cpp
MatusKysel: I think this is ok, invalid block will be marked for specific peer, but never broadcasted or inserted anywhere. so it does not affect anything

JakubFornadel: yeah you are right probably. But in functions that send data to the peer and mark it as known should mark everything at the end of function I guess

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
JakubFornadel: ah okey, good to know :)

src/network/tarcap/packets_handler/handlers/blocks_packet_handler.cpp
MatusKysel: yes, that was fixed

src/network/tarcap/packets_handler/handlers/dag_packets_handler.cpp
JakubFornadel: comment moved to https://github.com/Taraxa-project/taraxa-node/issues/1038

JakubFornadel (approved):

PR #1037, opened 2021-07-30
fix(pbft): do not return early from state operations when syncing

justinsnapp:
## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1647,10 +1647,19 @@ bool PbftManager::giveUpNextVotedBlock_() {
     return true;
   }
 
-  if (previous_round_next_voted_value_ == last_soft_voted_value_ && giveUpSoftVotedBlock_() &&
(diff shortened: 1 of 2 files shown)


gaoqi8649 (approved):

src/consensus/pbft_manager.cpp
```diff
@@ -659,8 +661,6 @@ bool PbftManager::stateOperations_() {
   LOG(log_tr_) << "PBFT current round is " << round;
   LOG(log_tr_) << "PBFT current step is " << step_;
 
-  if (is_syncing_()) return true;
```
JakubFornadel, line 662: so what problems did this cause ?

justinsnapp: @JakubFornadel  it causes the node to have trouble from recovering from a pbft state out of sync caused sync request.    We should add a check like this back in once we differentiate from bootstrapping and block syncing from pbft syncing.

PR #1039, opened 2021-08-03
Tarcap summary and send txs

JakubFornadel:
## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/network/tarcap/new_taraxa_capability.cpp
+++ b/src/network/tarcap/new_taraxa_capability.cpp
@@ -6,6 +6,7 @@
 #include "consensus/pbft_manager.hpp"
 #include "consensus/vote.hpp"
 #include "dag/dag.hpp"
+#include "network/tarcap/node_stats.hpp"
 #include "network/tarcap/packets_handler/handlers/blocks_packet_handler.hpp"
 #include "network/tarcap/packets_handler/handlers/dag_packets_handler.hpp"
 #include "network/tarcap/packets_handler/handlers/get_blocks_packet_handler.hpp"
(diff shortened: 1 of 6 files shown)


src/network/tarcap/new_taraxa_capability.cpp
```diff
@@ -121,26 +97,37 @@ TaraxaCapability::TaraxaCapability(std::weak_ptr<dev::p2p::Host> _host, NetworkC
...
-                                  [this] { peers_state_->packets_stats_.logStats(); });
+                                  [peers_state = peers_state_] { peers_state->packets_stats_.logStats(); });
   }
 
-  // TODO: implement logNodeStats()
-  //  const auto summary_interval_ms = 5 * 6 * lambda_ms_min;
-  //  periodic_events_tp_.post_loop( { summary_interval_ms }, [this] { logNodeStats(); });
+  // Logs node stats periodic event
+  const auto node_stats_log_interval = 5 * 6 * lambda_ms_min;
+  // node_stats is shared_ptr as it is probably less expensive to move periodically than object with uint64_t members
```
MatusKysel, line 126: this comment is misleading, I am ok with shared_ptr but you are not periodically moving any  "object with uint64_t members" but whole anonymous function and it's capture is evaluated only once

src/network/tarcap/node_stats.cpp
```diff
@@ -0,0 +1,216 @@
...
+}
+
+uint64_t NodeStats::getNodeStatsLogInterval() const { return stats_log_interval_; }
+
+uint64_t NodeStats::syncTimeSeconds() const { return stats_log_interval_ * syncing_interval_count_ / 1000; }
+
+void NodeStats::logNodeStats() {
+  bool is_syncing = syncing_state_->is_syncing();
+
+  dev::p2p::NodeID max_pbft_round_nodeID;
```
MatusKysel, line 39: naming -> max_pbft_round_node_id

src/network/tarcap/node_stats.cpp
```diff
@@ -0,0 +1,216 @@
...
+      // Find max peer PBFT round
+      if (peer.second->pbft_round_ > peer_max_pbft_round) {
+        peer_max_pbft_round = peer.second->pbft_round_;
+        max_pbft_round_nodeID = peer.first;
+      }
+    }
+  }
+
+  // Local dag info...
+  auto local_max_level_in_dag = dag_mgr_->getMaxLevel();
```
MatusKysel, line 72: probably everything cloud be `const`

src/network/tarcap/new_taraxa_capability.cpp
```diff
@@ -121,26 +97,37 @@ TaraxaCapability::TaraxaCapability(std::weak_ptr<dev::p2p::Host> _host, NetworkC
...
+  auto node_stats = std::make_shared<NodeStats>(peers_state_, syncing_state_, pbft_chain, pbft_mgr, dag_mgr,
+                                                dag_blk_mgr, vote_mgr, trx_mgr, node_stats_log_interval);
+  periodic_events_tp_.post_loop({node_stats->getNodeStatsLogInterval()},
+                                [node_stats = std::move(node_stats)]() mutable { node_stats->logNodeStats(); });
```
MatusKysel, line 130: does it need to be mutable when using shr_ptr? I do know just asking

src/network/tarcap/new_taraxa_capability.cpp
JakubFornadel: so what is this if not recursive loop where you periodically move whole lambda function, which contains object with uint64_t members ? https://github.com/Taraxa-project/taraxa-node/blob/tarcap_summary_and_send_txs/src/util/thread_pool.cpp#L69

Move operator on lambda internally calls move operator on all captured objects

JakubFornadel: `class Test {
  public:
    Test() = default;
    Test(const Test&) { std::cout << "C"; }
    Test(Test&&) { std::cout << "M"; }
};

int main()
{
    Test test;

    std::function<void()> lambda = [test = std::move(test)]() { std::cout << "lambda"; };
    //auto lambda = [test = std::move(test)]() { std::cout << "lambda"; };
    auto lambda1 = std::move(lambda);
    auto lambda2 = std::move(lambda1);
    auto lambda3 = std::move(lambda2);
}`

hmm when used std::function it prints MM, when using auto it prints MMMM, interesting

MatusKysel: https://web.mst.edu/~nmjxv3/articles/lambdas.html please read
Even
```
std::function<void()> lambda = [test = std::move(test)]() { std::cout << "lambda"; };
```

will print it twice, this of `test` as anonymous class variable

JakubFornadel: thanks for writing me what I just wrote a commet above :D That was a comment that was supposed to say yes you are right std::function does not call move constructor multiple times, auto = []() {} does

src/network/tarcap/node_stats.cpp
JakubFornadel: everything is copy paste but yeah I agree that I can refactor bad naming here...

JakubFornadel: fixed

src/network/tarcap/node_stats.cpp
JakubFornadel: fixed

src/network/tarcap/new_taraxa_capability.cpp
JakubFornadel: fixed, no shared_ptr used anymore

MatusKysel (approved):

PR #1044, opened 2021-08-04
Feature/rework tests

kstdl:
## Purpose

Originally it was a PR fixing an [issue](https://github.com/Taraxa-project/taraxa-node/issues/924). But some other fails was detected on macOS builds during the tests. So it is converted to a feature request.

## Related Github issues

https://github.com/Taraxa-project/taraxa-node/issues/924

PR #1046, opened 2021-08-04
chore(tests): fix full_node_test

kstdl:
## Purpose

Fix some of fails in a full_node_test including accidentally changed `save_network_to_file test`

### DIFF
--- a/tests/full_node_test.cpp
+++ b/tests/full_node_test.cpp
@@ -1176,7 +1176,7 @@ TEST_F(FullNodeTest, two_nodes_run_two_transactions) {
 
 TEST_F(FullNodeTest, save_network_to_file) {
   auto node_cfgs = make_node_cfgs(3);
-  auto nodes = launch_nodes(node_cfgs);
+  { auto nodes = launch_nodes(node_cfgs); }
   {
     FullNode::Handle node2(node_cfgs[1], true);
     FullNode::Handle node3(node_cfgs[2], true);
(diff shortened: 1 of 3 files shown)


tests/full_node_test.cpp
```diff
@@ -1176,7 +1176,7 @@ TEST_F(FullNodeTest, two_nodes_run_two_transactions) {
 
 TEST_F(FullNodeTest, save_network_to_file) {
   auto node_cfgs = make_node_cfgs(3);
-  auto nodes = launch_nodes(node_cfgs);
+  { auto nodes = launch_nodes(node_cfgs); }
```
JakubFornadel, line 1179: what is this nodes even good for ? Seems it is not used in whole test ?

MatusKysel: here you create network and destroy it so next time will be loaded from file

MatusKysel: I do not know when the scope was removed as on develop it still there https://github.com/Taraxa-project/taraxa-node/blob/develop/tests/full_node_test.cpp#L1179

kstdl: It was accidentally removed here https://github.com/Taraxa-project/taraxa-node/pull/1044/files#diff-cb712a3395b4b23506c8277c7056adf0f3cf679e3e7f04bb3de00f3fb1a68dd4L1179. 
So just I returned it. You can also read it in the PR description :)

JakubFornadel: OK I get it now, although without your comments I would be totally lost on what is actually going on on this line - I dont know maybe a comment would be nice but that might be just me

MatusKysel (approved):

PR #1048, opened 2021-08-04
refactor(full_node): remove register_s_ptr and related stuff

kstdl:
## Purpose

Remove register_s_ptr, FullNodeHandle and rework creation of nodes in the tests

## For reviewers 

commit `chore(tests): fix full_node_test` will be removed after https://github.com/Taraxa-project/taraxa-node/pull/1046 merging


src/consensus/block_proposer.hpp
```diff
@@ -96,7 +96,7 @@ class BlockProposer : public std::enable_shared_from_this<BlockProposer> {
   void start();
   void stop();
   std::shared_ptr<BlockProposer> getShared();
-  void setNetwork(std::weak_ptr<Network> network) { network_ = move(network); }
+  void setNetwork(std::shared_ptr<Network> network) { network_ = move(network); }
```
JakubFornadel, line 99: are you sure about this one ? We have some weak_ptr's because of shared_ptr's circular dependency issue.

src/final_chain/final_chain.cpp
```diff
@@ -78,9 +78,8 @@ class FinalChainImpl final : public FinalChain {
   future<shared_ptr<FinalizationResult>> finalize(NewBlock new_blk,
                                                   finalize_precommit_ext precommit_ext = {}) override {
     auto p = make_shared<promise<shared_ptr<FinalizationResult>>>();
-    executor_([this, new_blk = move(new_blk), precommit_ext = move(precommit_ext), p]() mutable {
-      p->set_value(finalize_(move(new_blk), precommit_ext));
-    });
+    executor_([this, s = shared_from_this(), new_blk = move(new_blk), precommit_ext = move(precommit_ext),
```
JakubFornadel, line 81: so what changed that "s = shared_from_this()" is now required here and before we did not need it ?

src/consensus/block_proposer.hpp
kstdl: I think in some commit after I returned it back, but forgot to include it during the splitting. Will check that

src/network/rpc/Test.cpp
```diff
@@ -102,26 +98,18 @@ Json::Value Test::create_test_coin_transactions(const Json::Value &param1) {
...
+      // get trx receiving time stamp
+      uint i = 0;
+      while (i < number) {
+        auto now = getCurrentTimeMilliSeconds();
+        val_t value = val_t(100);
+        auto trx = taraxa::Transaction(i + nonce, value, 1000, 0, bytes(), sk, receiver);
+        LOG(log_time) << "Transaction " << trx.getHash() << " received at: " << now;
+        node->getTransactionManager()->insertTransaction(trx, false, true);
+        thisThreadSleepForMicroSeconds(delay);
+        i++;
```
JakubFornadel, line 110: you are changing the same stuff as in https://github.com/Taraxa-project/taraxa-node/pull/1046/files. Please clean up

src/consensus/block_proposer.hpp
JakubFornadel: are you sure then about the splitting then ?

JakubFornadel: I think it is better in separate PR's but dont want to merge fixes with bugs...

src/network/rpc/Test.cpp
kstdl: Will be rebase after merging the PR and this change will gone

src/consensus/block_proposer.hpp
kstdl: It wont have conflicts because it is just have the commit from other one PR and it will be removed after it is merge. This splitting is done to made a review easier and PRs are merging to a feature, not a develop. So after all  that merges I will just compare all the changes to original one I made

src/network/rpc/Test.cpp
JakubFornadel: but why would you do it like this ? Those PR's are independent arent they ?

src/taraxad/taraxad_version.hpp
```diff
@@ -0,0 +1,4 @@
+#pragma once
+
+// get version of cmake project
+#define TARAXAD_VERSION "1.9.0"
```
JakubFornadel, line 4: This is old file that we dont use anymore I think

tests/dag_block_test.cpp
```diff
@@ -173,7 +173,7 @@ TEST_F(DagBlockTest, sign_verify) {
 
 TEST_F(DagBlockTest, push_and_pop) {
   auto node_cfgs = make_node_cfgs(1);
-  FullNode::Handle node(node_cfgs[0]);
+  auto node = create_nodes(1).front();
```
JakubFornadel, line 176: I dont know, you are creating node_cfgs which are used later in test. Seemed much more intuitive to see the same config being used in test and also in FullNode creation compared to this where you create configs for tests and then inside create_nodes create again new configs that should and must be the same. Seems like a nice opportunity for future bugs/bad design.

tests/dag_block_test.cpp
```diff
@@ -218,8 +218,7 @@ TEST_F(DagBlockTest, overlap) {
 }
 
 TEST_F(DagBlockMgrTest, proposal_period) {
-  auto node_cfgs = make_node_cfgs(1);
-  FullNode::Handle node(node_cfgs[0]);
+  auto node = create_nodes(1).front();
```
JakubFornadel, line 221: How are these changes related to removing "register_s_ptr and related stuff" ?

JakubFornadel: Aaah I see now you replace FullNode::Handle , which is some weird derived shared_ptr. My bad

src/final_chain/final_chain.cpp
kstdl: There was a situations when FinalChain was already destructed when we entering this lambda

src/network/rpc/Test.cpp
kstdl: Will remove this commit from there. There is no sufficient reason for it 🥴

tests/dag_block_test.cpp
kstdl: Totally agree. Will fix that

JakubFornadel (approved):

PR #1058, opened 2021-08-06
fix(pbft): make sure don't give up a cert voted value

justinsnapp:
## Purpose

Introduced a consensus error where could give up on previously cert voted (but not pushed) value in giveUpNextVotedBlock_()

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -915,18 +915,22 @@ void PbftManager::certifyBlock_() {
 
       if (executed_soft_voted_block_for_this_round || unverified_soft_vote_block_for_this_round_is_valid) {
         // generate cert vote
-        // comparePbftBlockScheduleWithDAGblocks_ has checked the cert voted block exist
-        auto cert_voted_block = pbft_chain_->getUnverifiedPbftBlock(soft_voted_block_for_this_round_.first);
-        auto batch = db_->createWriteBatch();
-        db_->addPbftCertVotedBlockHashToBatch(round, soft_voted_block_for_this_round_.first, batch);
-        db_->addPbftCertVotedBlockToBatch(*cert_voted_block, batch);
-        db_->commitWriteBatch(batch);
-        cert_voted_values_for_round_[round] = soft_voted_block_for_this_round_.first;
-        should_have_cert_voted_in_this_round_ = true;
         auto place_votes = placeVote_(soft_voted_block_for_this_round_.first, cert_vote_type, round, step_);
         if (place_votes) {
           LOG(log_nf_) << "Cert votes " << place_votes << " voting " << soft_voted_block_for_this_round_.first
                        << " in round " << round;
+
+          // comparePbftBlockScheduleWithDAGblocks_ has checked the cert voted block exist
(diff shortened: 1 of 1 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -915,22 +915,18 @@ void PbftManager::certifyBlock_() {
...
       if (executed_soft_voted_block_for_this_round || unverified_soft_vote_block_for_this_round_is_valid) {
         // generate cert vote
+        // comparePbftBlockScheduleWithDAGblocks_ has checked the cert voted block exist
+        auto cert_voted_block = pbft_chain_->getUnverifiedPbftBlock(soft_voted_block_for_this_round_.first);
+        auto batch = db_->createWriteBatch();
+        db_->addPbftCertVotedBlockHashToBatch(round, soft_voted_block_for_this_round_.first, batch);
+        db_->addPbftCertVotedBlockToBatch(*cert_voted_block, batch);
+        db_->commitWriteBatch(batch);
+        cert_voted_values_for_round_[round] = soft_voted_block_for_this_round_.first;
+        should_have_cert_voted_in_this_round_ = true;
```
gaoqi8649, line 925: I don't think it correct. If failed in vote sortition, the node will think it has voted already. Certify state is pulling state, then will not be able to vote again

src/consensus/pbft_manager.cpp
```diff
@@ -982,25 +978,23 @@ void PbftManager::secondFinish_() {
 
   if (!next_voted_soft_value_ && soft_voted_block_for_this_round_.second &&
       soft_voted_block_for_this_round_.first != NULL_BLOCK_HASH) {
+    db_->savePbftMgrStatus(PbftMgrStatus::next_voted_soft_value, true);
+    next_voted_soft_value_ = true;
```
gaoqi8649, line 982: Same here

src/consensus/pbft_manager.cpp
```diff
@@ -982,25 +978,23 @@ void PbftManager::secondFinish_() {
...
-      db_->savePbftMgrStatus(PbftMgrStatus::next_voted_soft_value, true);
-      next_voted_soft_value_ = true;
     }
   }
 
   if (!next_voted_null_block_hash_ && round >= 2 &&
       (giveUpNextVotedBlock_() || step_ > MAX_WAIT_FOR_NEXT_VOTED_BLOCK_STEPS) &&
       (cert_voted_values_for_round_.find(round) == cert_voted_values_for_round_.end())) {
+    db_->savePbftMgrStatus(PbftMgrStatus::next_voted_null_block_hash, true);
+    next_voted_null_block_hash_ = true;
```
gaoqi8649, line 994: Same here

src/consensus/pbft_manager.cpp
justinsnapp: Good points.

src/consensus/pbft_manager.cpp
justinsnapp: Fixed

src/consensus/pbft_manager.cpp
justinsnapp: Fixed

gaoqi8649 (approved):

PR #1062, opened 2021-08-09
Rework final chain executor

kstdl:
## Purpose

1. Remove unnecessary mutex from thread_pool and it locks
2. Remove `executor_` field from `FinalChain` and replace it with direct `executor_thread_.post`. We don't need `strand` here and executor implementation caused some side-effects that caused tests failing

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/final_chain/final_chain.cpp
+++ b/src/final_chain/final_chain.cpp
@@ -22,7 +22,6 @@ class FinalChainImpl final : public FinalChain {
 
   size_t transaction_count_hint_;
   util::ThreadPool executor_thread_{1};
-  util::task_executor_t executor_ = executor_thread_.strand();
 
   atomic<uint64_t> num_executed_dag_blk_ = 0;
   atomic<uint64_t> num_executed_trx_ = 0;
@@ -78,8 +77,8 @@ class FinalChainImpl final : public FinalChain {
   future<shared_ptr<FinalizationResult>> finalize(NewBlock new_blk,
                                                   finalize_precommit_ext precommit_ext = {}) override {
     auto p = make_shared<promise<shared_ptr<FinalizationResult>>>();
-    executor_([this, s = shared_from_this(), new_blk = move(new_blk), precommit_ext = move(precommit_ext),
-               p]() mutable { p->set_value(finalize_(move(new_blk), precommit_ext)); });
+    executor_thread_.post([this, s = shared_from_this(), new_blk = move(new_blk), precommit_ext = move(precommit_ext),
+                           p]() mutable { p->set_value(finalize_(move(new_blk), precommit_ext)); });
     return p->get_future();
   }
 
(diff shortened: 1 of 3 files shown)


JakubFornadel:
yeah those locks did not hurt anybody but dont need to be there either :) So when should strand be used ? Just a question out of curiosity and laziness to google it :)

JakubFornadel (approved):

MatusKysel (approved):

MatusKysel:
> yeah those locks did not hurt anybody but dont need to be there either :) So when should strand be used ? Just a question out of curiosity and laziness to google it :)



> yeah those locks did not hurt anybody but dont need to be there either :) So when should strand be used ? Just a question out of curiosity and laziness to google it :)

it was used by `util::task_executor_t executor_ = executor_thread_.strand();` so you did not need execute `executor_.post()` just `executor_` so useless

MatusKysel:
> > > yeah those locks did not hurt anybody but dont need to be there either :) So when should strand be used ? Just a question out of curiosity and laziness to google it :)
> > 
> > 
> > it was used by `util::task_executor_t executor_ = executor_thread_.strand();` so you did not need execute `executor_.post()` just `executor_` so useless
> 
> @MatusKysel No it is not just that. It posted tasks on boost::asio::make_strand(ioc_) and in normal post it is posted on normal boost::asio::io_context. So the question was when should be strand used

you're right. This is what I found `The io_context::strand class provides the ability to post and dispatch handlers with the guarantee that none of those handlers will execute concurrently. ` so maybe it should be there, so we are sure each block is execute correctly

src/final_chain/final_chain.cpp
```diff
@@ -78,8 +77,8 @@ class FinalChainImpl final : public FinalChain {
   future<shared_ptr<FinalizationResult>> finalize(NewBlock new_blk,
                                                   finalize_precommit_ext precommit_ext = {}) override {
     auto p = make_shared<promise<shared_ptr<FinalizationResult>>>();
-    executor_([this, s = shared_from_this(), new_blk = move(new_blk), precommit_ext = move(precommit_ext),
```
MatusKysel, line 81: correct me if I am wrong, but I think this should stay there along with changes related to it

kstdl: Yeah. Strand needed exactly for that you posted above. 
I don't see a situation in which we can run finalize concurrently. Executor thread pool working on only a one thread and I think there is no reason to add some more.
Another one argument there I thought was that there is no way for some block to be processed at the same time during normal operation. But I just checked that it could be met during a sync.

So if we want that code to be safe for increasing that count of threads in a pool that logic with a strand should be returned back. But it should be reworked. Cause with a current implementation we could met with an issue that some code will try to execute after FinalChain destruction.

PR #1064, opened 2021-08-09
perf(dag): use interal dag block cache instead of db

MatusKysel:
## Purpose

We already have internal cache for DAG blocks, but we do not use it and get them all the time from DB. With this approach we use it as much as possible

### DIFF
--- a/src/dag/dag.cpp
+++ b/src/dag/dag.cpp
@@ -283,11 +283,13 @@ void PivotTree::getGhostPath(blk_hash_t const &vertex, std::vector<blk_hash_t> &
 }
 
 DagManager::DagManager(blk_hash_t const &genesis, addr_t node_addr, std::shared_ptr<TransactionManager> trx_mgr,
-                       std::shared_ptr<PbftChain> pbft_chain, std::shared_ptr<DbStorage> db) try
+                       std::shared_ptr<PbftChain> pbft_chain, std::shared_ptr<DagBlockManager> dag_blk_mgr,
+                       std::shared_ptr<DbStorage> db) try
     : pivot_tree_(std::make_shared<PivotTree>(genesis, node_addr)),
       total_dag_(std::make_shared<Dag>(genesis, node_addr)),
       trx_mgr_(trx_mgr),
       pbft_chain_(pbft_chain),
+      dag_blk_mgr_(dag_blk_mgr),
       db_(db),
       anchor_(genesis),
       period_(0),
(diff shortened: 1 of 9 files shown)


mfrankovi (commented):
We actually before even had more memory caches all around but by testing we noticed that rocksdb actually has an excellent memory cache capabilities and data retrieved from the db which is relatively recently inserted or recently queried for, is retrieved very quickly from rocksdb internal memory cache and no major improvement was made by having our own custom memory caches. For these kind of changes I would recommend having some benchmark to prove that we are actually gaining some performance and that memory cache is needed. If so, the changes are good, if there is no improvement, in that case I would recommend completely removing the memory cache to simplify our code

MatusKysel:
> We actually before even had more memory caches all around but by testing we noticed that rocksdb actually has an excellent memory cache capabilities and data retrieved from the db which is relatively recently inserted or recently queried for, is retrieved very quickly from rocksdb internal memory cache and no major improvement was made by having our own custom memory caches. For these kind of changes I would recommend having some benchmark to prove that we are actually gaining some performance and that memory cache is needed. If so, the changes are good, if there is no improvement, in that case I would recommend completely removing the memory cache to simplify our code

Good point, I will try to measure it so we know for sure

src/dag/dag.cpp
```diff
@@ -283,11 +283,13 @@ void PivotTree::getGhostPath(blk_hash_t const &vertex, std::vector<blk_hash_t> &
 }
 
 DagManager::DagManager(blk_hash_t const &genesis, addr_t node_addr, std::shared_ptr<TransactionManager> trx_mgr,
-                       std::shared_ptr<PbftChain> pbft_chain, std::shared_ptr<DbStorage> db) try
+                       std::shared_ptr<PbftChain> pbft_chain, std::shared_ptr<DagBlockManager> dag_blk_mgr,
+                       std::shared_ptr<DbStorage> db) try
```
JakubFornadel, line 287: what is "try" keyword for btw ?

gaoqi8649 (commented):
I remember DagBlockManager only save unfinalized DAG blocks. After finalization, will remove the DAG blocks from DagBlockManager queue and add into DB.
Like, we are finalized at period N, and get some unfinalized DAG blocks. Some of the unfinalized DAG blocks' pivot/tips may have finalized already. Those blocks may have been removed from  DagBlockManager queue already. How do we handle the case?

gaoqi8649 (approved):

PR #1066, opened 2021-08-09
chore(dag): do not send blocks requests in loop

MatusKysel:


src/network/taraxa_capability.cpp
```diff
@@ -112,8 +112,8 @@ bool TaraxaCapability::sealAndSend(NodeID const &nodeID, unsigned packet_type, R
   return true;
 }
 
-std::pair<bool, std::vector<blk_hash_t>> TaraxaCapability::checkDagBlockValidation(DagBlock const &block) {
-  std::vector<blk_hash_t> missing_blks;
+std::pair<bool, std::unordered_set<blk_hash_t>> TaraxaCapability::checkDagBlockValidation(DagBlock const &block) {
+  std::unordered_set<blk_hash_t> missing_blks;
```
JakubFornadel, line 116: We need ordering of requested blocks, which is not guaranteed in unordered_set, is it ?

MatusKysel: ordering will be guaranteed by node that is sending requested blocks, as we are sending them by levels:
```
      const auto &blocks = dag_mgr_->getNonFinalizedBlocks();
      for (auto &level_blocks : blocks) {
```

JakubFornadel: hmm so 2 questions:

- I dont really see any perf gain here in using map vs vector, why is it even used ?
- that code you copy pasted here is actually pretty unefficient when I look at it now:

why are even iterating over non finalized blocks like this ? It seems useless. Most of the cases when requesting missing blocks it is very few and we are iterating over all non finalized dag blocks(which could be potentially hundreds or even thousands in case of network being stuck). Why not iterate over requested missing blocks and just check non-ordered blocks map if it exists there. In such case, which is much more efficient compared to this loop the ordering guarantee is gone and you need vector back ?

JakubFornadel: aa okey getNonFinalizedBlocks() does not return true map that could be used, so that answers my question #2.

MatusKysel: set is used so we do need to check for duplicates

JakubFornadel: hmm, ok I get it now. Thanks

JakubFornadel (approved):

PR #1081, opened 2021-08-17
Fix consensus issue with giving up soft voted value that could be potentially cert voted

justinsnapp:
## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1739,12 +1739,6 @@ bool PbftManager::giveUpNextVotedBlock_() {
     return true;
   }
 
-  if (step_ > MAX_WAIT_FOR_NEXT_VOTED_BLOCK_STEPS) {
-    LOG(log_nf_) << "In round " << round << " step " << step_ << ", have exceeded max waiting step "
-                 << MAX_WAIT_FOR_NEXT_VOTED_BLOCK_STEPS << ". Give up voted value " << previous_round_next_voted_value_;
-    return true;
-  }
-
   auto pbft_block = pbft_chain_->getUnverifiedPbftBlock(previous_round_next_voted_value_);
   if (!pbft_block) {
     pbft_block = db_->getPbftCertVotedBlock(previous_round_next_voted_value_);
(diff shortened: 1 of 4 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -1638,6 +1657,19 @@ bool PbftManager::giveUpSoftVotedBlock_() {
...
+  if (!pbft_block) {
+    pbft_block = db_->getPbftCertVotedBlock(previous_round_next_voted_value_);
+  }
+
+  if (pbft_block) {
+    // Have a block, but is it valid?
+    if (!checkPbftBlockValid_(previous_round_next_voted_value_)) {
+      // Received the block, but not valid
+      return true;
+    }
```
gaoqi8649, line 1670: Do we need add:
else {
    return false;
}

If receive the block already and the block is valid, should continue with the value. Don't terminate

gaoqi8649 (approved):

src/consensus/pbft_manager.cpp
justinsnapp: @gaoqi8649  I thought so at first too, but being valid in this sense (ie a well signed block) isn't sufficient because it can be a malicious block that points to a bogus DAG anchor that doesn't exist.  In that case we DO want to give up.

src/consensus/pbft_manager.cpp
```diff
@@ -1638,6 +1657,19 @@ bool PbftManager::giveUpSoftVotedBlock_() {
   unsigned long elapsed_wait_soft_voted_block_in_ms =
       std::chrono::duration_cast<std::chrono::milliseconds>(soft_voted_block_wait_duration).count();
 
+  auto pbft_block = pbft_chain_->getUnverifiedPbftBlock(previous_round_next_voted_value_);
+  if (!pbft_block) {
+    pbft_block = db_->getPbftCertVotedBlock(previous_round_next_voted_value_);
+  }
+
+  if (pbft_block) {
```
JakubFornadel, line 1665: This should be "else" branch of "if (!pbft_block) " condition from line 1660

JakubFornadel (approved):

PR #1087, opened 2021-08-19
chore(dag): fix dag block issue

MatusKysel:


src/storage/db_storage.cpp
```diff
@@ -219,6 +219,18 @@ std::shared_ptr<DagBlock> DbStorage::getDagBlock(blk_hash_t const& hash) {
   return nullptr;
 }
 
+bool DbStorage::checkDagBlock(blk_hash_t const& hash) {
```
mfrankovi, line 222: We have transactionInDb and pbftBlockInDb methods, maybe dagBlockInDb would be a better name than checkDagBlock

src/node/full_node.cpp
```diff
@@ -221,7 +221,7 @@ void FullNode::start() {
         received_blocks_++;
       }
 
-      if (dag_blk_mgr_->pivotAndTipsAvailable(blk)) {
+      if (dag_mgr_->pivotAndTipsAvailable(blk)) {
```
mfrankovi, line 224: Maybe put check inside of the addDagBlock and have it return a bool. But this is fine as well

mfrankovi (approved):

src/node/full_node.cpp
MatusKysel: addDagBlock is used also elsewhere so I think I can not do it

src/storage/db_storage.cpp
MatusKysel: good point I will fix it

justinsnapp (approved):

PR #1097, opened 2021-08-23
build: remove optimization when building in debug mode

JakubFornadel:
## Version Notes ##
{ major | minor | patch }


MatusKysel (changes requested):
If we are disabling optimization, then we should introduce `CMAKE_CXX_FLAGS_RELWITHDEBINFO` and use `-DCMAKE_BUILD_TYPE=RELWITHDEBINFO` in our docker builds, so dev/testnet is not so slow

JakubFornadel:
> If we are disabling optimization, then we should introduce `CMAKE_CXX_FLAGS_RELWITHDEBINFO` and use `-DCMAKE_BUILD_TYPE=RELWITHDEBINFO` in our docker builds, so dev/testnet is not so slow

If you feel like changing it, go ahead. If not I will look at it after merging my branch with Miho's

MatusKysel (approved):

kstdl:
Do we need this true debug build in the CI and executable builds? 
If not we should also change it in `Dockerfile` and circle CI config files

MatusKysel:
> Do we need this true debug build in the CI and executable builds?
> If not we should also change it in `Dockerfile` and circle CI config files

it is changes in Docker file to new `RelWithDebInfo` and circle CI is using Release

kstdl:
@MatusKysel shouldn't we use the same build type for the binaries release(circle CI) as for docker image build? Discussed it some time ago as far as I remember

MatusKysel:
@kstdl for now we want to have debug info in docker, in case our nodes crash, so we can check what is the problem. In future it will be probably Release

PR #1098, opened 2021-08-23
Fix for syncing state

mfrankovi:
Fixed a bug that dag_syncing state would be left incorrectly to true if an empty BlocksPacket was not sent back
Enable gossip of any proposed block even if in syncing state. 
Enable gossip any block if not in pbft syncing state
Prevent gossip of pbft blocks that are missing dag anchor

### DIFF
--- a/src/dag/dag_block_manager.cpp
+++ b/src/dag/dag_block_manager.cpp
@@ -234,7 +234,7 @@ std::pair<size_t, size_t> DagBlockManager::getDagBlockQueueSize() const {
   return res;
 }
 
-DagBlock DagBlockManager::popVerifiedBlock(bool level_limit, uint64_t level) {
+std::pair<std::shared_ptr<DagBlock>, bool> DagBlockManager::popVerifiedBlock(bool level_limit, uint64_t level) {
   uLock lock(shared_mutex_for_verified_qu_);
   if (level_limit) {
     while ((verified_qu_.empty() || verified_qu_.begin()->first >= level) && !stopped_) {
@@ -245,12 +245,14 @@ DagBlock DagBlockManager::popVerifiedBlock(bool level_limit, uint64_t level) {
       cond_for_verified_qu_.wait(lock);
     }
   }
-  if (stopped_) return DagBlock();
+  if (stopped_) return {std::make_shared<DagBlock>(), false};
 
-  auto blk = verified_qu_.begin()->second.front();
+  auto blk = std::make_shared<DagBlock>(verified_qu_.begin()->second.front());
   verified_qu_.begin()->second.pop_front();
   if (verified_qu_.begin()->second.empty()) verified_qu_.erase(verified_qu_.begin());
-  return blk;
+  auto status = blk_status_.get(blk->getHash());
+
+  return {blk, status.second && status.first == BlockStatus::proposed};
 }
 
 void DagBlockManager::pushVerifiedBlock(DagBlock const &blk) {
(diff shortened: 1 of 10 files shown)


MatusKysel (approved):

src/network/taraxa_capability.cpp
```diff
@@ -1074,9 +1074,10 @@ void TaraxaCapability::sendSyncedMessage() {
   }
 }
 
-void TaraxaCapability::onNewBlockVerified(DagBlock const &block) {
-  // If node is syncing this is an old block that has been verified - no block goosip is needed
-  if (syncing_state_.is_syncing()) {
+void TaraxaCapability::onNewBlockVerified(DagBlock const &block, bool proposed) {
+  // If node is pbft syncing and block is not proposed by us, this is an old block that has been verified - no block
+  // goosip is needed
```
JakubFornadel, line 1079: if node syncing, node should not propose anything as it does not have current data ?

src/dag/dag_block_manager.cpp
```diff
@@ -245,12 +245,14 @@ DagBlock DagBlockManager::popVerifiedBlock(bool level_limit, uint64_t level) {
       cond_for_verified_qu_.wait(lock);
     }
   }
-  if (stopped_) return DagBlock();
+  if (stopped_) return {std::make_shared<DagBlock>(), false};
```
JakubFornadel, line 248: maybe nullptr could be returned here and caller should check the value to nullptr ?

src/network/taraxa_capability.cpp
mfrankovi: I did this as a quick fix but I would propose completely remove the dag syncing state, if we are pbft synced, we are synced. There is always a chance that DAG structure is not exactly the same throughout network and it can be partitioned for some amount of time. Currently that constant call to restartSyncing from pbft manager puts us often into dag_syncing state which than would stop block gossip if I left it as it was before. This code is for gossiping and not proposing.

src/dag/dag_block_manager.cpp
mfrankovi: sure, I will change this

src/dag/dag_block_manager.cpp
```diff
@@ -245,12 +245,14 @@ DagBlock DagBlockManager::popVerifiedBlock(bool level_limit, uint64_t level) {
...
 
-  auto blk = verified_qu_.begin()->second.front();
+  auto blk = std::make_shared<DagBlock>(verified_qu_.begin()->second.front());
```
JakubFornadel, line 250: This calls DagBlock copy ctor thus making unnecessary copy. std::move could be used here if we add move semantics support to DagBlock class:
`  DagBlock(const DagBlock&) = default;
  DagBlock& operator=(const DagBlock&) = default;
  DagBlock(DagBlock&&) = default;
  DagBlock& operator=(DagBlock&&) = default;`

src/network/taraxa_capability.cpp
JakubFornadel: if restartSyncing triggers only dag syncing, it does not block gossiping because syncing_state_.is_pbft_syncing() returns true only if we are syncing pbft blocks. And if we are truly pbft syncing, I think it should never happen that the node propose new dag/or pbft block so we should not get into this function with variables:  
- proposed == true and 
- syncing_state_.is_pbft_syncing() == true.

JakubFornadel: So seems to me that changing condition from `if (syncing_state_.is_syncing())` to `if (syncing_state_.is_pbft_syncing())` might be enough and what you wanted to achieve by checking `!proposed` could be handled during proposing by checking pbft syncing and if we are syncing, nothing should be proposed ?

mfrankovi: there is a chance for race condition, we propose a block and only at that point we realize we are not pbft synced, we still should gossip it. Proposing is disabled by not feeding transactions to non-synced nodes but we could add additional protection

PR #1106, opened 2021-08-25
Issue 1079. fix order of data updating during pushPbftBlock_

kstdl:


conanfile.py
```diff
@@ -14,7 +14,7 @@ class TaraxaConan(ConanFile):
 
     def requirements(self):
         self.requires("boost/1.76.0")
-        self.requires("cppcheck/2.3")
+        # self.requires("cppcheck/2.3")
```
MatusKysel, line 17: why this is disabled?

src/consensus/pbft_manager.cpp
```diff
@@ -1586,7 +1586,8 @@ void PbftManager::finalize_(PbftBlock const &pbft_block, vector<h256> finalized_
 
 bool PbftManager::pushPbftBlock_(PbftBlockCert const &pbft_block_cert_votes, vec_blk_t const &dag_blocks_order,
                                  bool sync) {
-  auto const &pbft_block_hash = pbft_block_cert_votes.pbft_blk->getBlockHash();
+  auto pbft_block = pbft_block_cert_votes.pbft_blk;
```
MatusKysel, line 1589: you could use just hash here and even in `getJsonStrForBlock` as you are not using anything except that

tests/network_test.cpp
```diff
@@ -189,7 +189,12 @@ TEST_F(NetworkTest, DISABLED_sync_large_pbft_block) {
 
   auto pbft_blocks1 = nodes[0]->getDB()->getPbftBlock(1);
   auto pbft_blocks2 = nodes2[0]->getDB()->getPbftBlock(1);
-  EXPECT_EQ(pbft_blocks1, pbft_blocks2);
+  EXPECT_EQ(pbft_blocks1->rlp(true), pbft_blocks2->rlp(true));
```
MatusKysel, line 192: you are modifying disabled test, so if this is the fix for it, maybe you should enabled it, if not do we need this test?

MatusKysel:
also you committed older version of aleth

conanfile.py
kstdl: added by accident during changes moving from another branch

tests/network_test.cpp
kstdl: I will move all that changes to a separate branch where this is fixed

src/consensus/pbft_manager.cpp
kstdl: Don't getting what about is in the first part of your message, because pbft_block using below in some lines. And I just moved that line above
And good point about `getJsonStrForBlock`, changed that

MatusKysel: sorry my bad I did not notice

MatusKysel (approved):

JakubFornadel (approved):

justinsnapp (approved):

PR #1107, opened 2021-08-25
chore:split transactions into multiple packets

mfrankovi:
Splitting transactions packet to send only 1000 transactions in one packet not to block transactions processing. This is only an issue when dag blocks production is stuck and transactions in verified queue grow to large numbers.

Increased the MAX_CHECK_ALIVE_COUNT from 5 to 20 effectively increasing the liveness timeout from 30 seconds to 2 minutes. There is a plan to completely remove this timeout since there is a timeout mechanism in the p2p library.


MatusKysel (approved):

src/network/taraxa_capability.cpp
```diff
@@ -1155,11 +1155,21 @@ void TaraxaCapability::sendTransactions(NodeID const &_id, std::vector<taraxa::b
...
     trx_bytes.insert(trx_bytes.end(), std::begin(transaction), std::end(transaction));
+    transaction_counter++;
+    if (transaction_counter == MAX_TRANSACTIONS_IN_PACKET) {
+      s.appendRaw(trx_bytes, transaction_counter);
+      transaction_counter = 0;
+      sealAndSend(_id, TransactionPacket, move(s));
+      trx_bytes.clear();
+    }
+  }
+  if (transaction_counter) {
```
JakubFornadel, line 1169: Just a note: I think with do while(transaction_counter <= transactions.size()) you could get rid of this whole condition and two sealAndSend calls

mfrankovi: I changed the implementation to actually send max 1000 transactions each interval (current this is 100ms) because only splitting the packets still basically sends all transactions at once

PR #1117, opened 2021-08-27
chore(pbft): Fix dumping error logs

gaoqi8649:
## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -901,10 +901,14 @@ void PbftManager::certifyBlock_() {
 
       bool unverified_soft_vote_block_for_this_round_is_valid = false;
       if (!executed_soft_voted_block_for_this_round) {
-        if (checkPbftBlockValid_(soft_voted_block_for_this_round_.first)) {
-          LOG(log_tr_) << "checkPbftBlockValid_ returned true";
+        auto block = pbft_chain_->getUnverifiedPbftBlock(soft_voted_block_for_this_round_.first);
+        if (block && pbft_chain_->checkPbftBlockValidation(*block)) {
           unverified_soft_vote_block_for_this_round_is_valid = true;
         } else {
+          if (!block) {
+            LOG(log_er_) << "Cannot find the unverified pbft block " << soft_voted_block_for_this_round_.first
+                         << " in round " << round << " step 3";
+          }
           syncPbftChainFromPeers_(invalid_soft_voted_block, soft_voted_block_for_this_round_.first);
         }
       }
@@ -1352,15 +1356,6 @@ std::pair<blk_hash_t, bool> PbftManager::identifyLeaderBlock_(std::vector<Vote> 
   return std::make_pair(leader.second, true);
 }
 
-bool PbftManager::checkPbftBlockValid_(blk_hash_t const &block_hash) const {
-  auto block = pbft_chain_->getUnverifiedPbftBlock(block_hash);
-  if (!block) {
-    LOG(log_er_) << "Cannot find the unverified pbft block, block hash " << block_hash;
-    return false;
-  }
-  return pbft_chain_->checkPbftBlockValidation(*block);
-}
-
 bool PbftManager::syncRequestedAlreadyThisStep_() const {
   return getPbftRound() == pbft_round_last_requested_sync_ && step_ == pbft_step_last_requested_sync_;
 }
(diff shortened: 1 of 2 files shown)


MatusKysel (approved):

src/consensus/pbft_manager.cpp
```diff
@@ -1455,20 +1455,29 @@ std::pair<vec_blk_t, bool> PbftManager::comparePbftBlockScheduleWithDAGblocks_(P
...
+    if (!pbft_block) {
+      LOG(log_nf_) << "Can not find the cert voted block hash " << cert_voted_block_hash
+                   << " in both pbft queue and DB";
+      return false;
+    }
+    // Read from DB pushing into unverified queue
+    pbft_chain_->pushUnverifiedPbftBlock(pbft_block);
+  }
+
+  if (!checkPbftBlockValid_(cert_voted_block_hash)) {
```
JakubFornadel, line 1470: Let's delete this function and use `pbft_chain_->checkPbftBlockValidation()` instead. The call it on 4 places, but on 3 places you basically already have the shared_ptr to block. On the 4th place you can call `block = pbft_chain_->getUnverifiedPbftBlock()` and then `pbft_chain_->checkPbftBlockValidation(block)`

src/consensus/pbft_manager.cpp
```diff
@@ -1455,20 +1455,29 @@ std::pair<vec_blk_t, bool> PbftManager::comparePbftBlockScheduleWithDAGblocks_(P
...
   if (!pbft_block) {
-    LOG(log_er_) << "Can not find the cert vote block hash " << cert_voted_block_hash << " in pbft queue";
+    pbft_block = db_->getPbftCertVotedBlock(cert_voted_block_hash);
+    if (!pbft_block) {
+      LOG(log_nf_) << "Can not find the cert voted block hash " << cert_voted_block_hash
+                   << " in both pbft queue and DB";
+      return false;
+    }
+    // Read from DB pushing into unverified queue
+    pbft_chain_->pushUnverifiedPbftBlock(pbft_block);
```
JakubFornadel, line 1467: If we are pushing block back to the unverified queue only in scenarios when the node got restarted so it does not have it memory, please comment it properly because I doubt that anyone would know why is this here except you...

Or put some comment above `if (!pbft_block)` condition if everything inside is related to the node restart scenario

src/consensus/pbft_manager.cpp
```diff
@@ -1455,20 +1455,29 @@ std::pair<vec_blk_t, bool> PbftManager::comparePbftBlockScheduleWithDAGblocks_(P
...
-    syncPbftChainFromPeers_(invalid_cert_voted_block, cert_voted_block_hash);
-    return false;
-  }
   auto pbft_block = pbft_chain_->getUnverifiedPbftBlock(cert_voted_block_hash);
   if (!pbft_block) {
-    LOG(log_er_) << "Can not find the cert vote block hash " << cert_voted_block_hash << " in pbft queue";
+    pbft_block = db_->getPbftCertVotedBlock(cert_voted_block_hash);
+    if (!pbft_block) {
+      LOG(log_nf_) << "Can not find the cert voted block hash " << cert_voted_block_hash
+                   << " in both pbft queue and DB";
```
JakubFornadel, line 1463: If I understood you correctly on call you said this can happen only after the node is rebooted(restarted) but I see it very often during testing and I am pretty sure the node is not restarted during those tests so there is also some other case I guess ?

gaoqi8649: getPbftCertVotedBlock when the node cert voted a PBFT block, getUnverifiedPbftBlock when the node receive the PBFT block from peers.
When node receive enough cert votes that doesn't mean the node also receive the proposed PBFT block. In real case, if node has received 2t+1 cert votes usually the node has received the proposed PBFT block. Because node send out proposed block in step 1 and send out cert votes in step 3, there are at lease 2 lambda time (2s) between it. So in real case, when see the logs usually after node reboot (that cause unverified queue empty).
In the unit tests, since we set a very short lambda time for speed up, 2 lambda time duration is very short that caused doesn't have a significant difference to send out proposed block and cert votes. So in unit tests, node may have received enough 2t+1 cert votes but have not received block yet often. I think this is the reason

src/consensus/pbft_manager.cpp
gaoqi8649: I prefer to keep the checkPbftBlockValid_ function. And in the other 3 places have changed to use pbft_chain_->checkPbftBlockValidation(block)

src/consensus/pbft_manager.cpp
gaoqi8649: yeah, add comments

src/consensus/pbft_manager.cpp
JakubFornadel: Separate function is created usually 1 of these reasons:

- some functionality is used repeatedly so you put it into the function -> not this case
- big chunks of code are logically grouped in functions so the code is cleaner -> not this case
- interface/data encapsulation (getters/setters/need to use locks, etc...) -> not this case

JakubFornadel: You are repeating this code everywhere in fact:
`auto pbft_block = pbft_chain_->getUnverifiedPbftBlock(previous_round_next_voted_value_);
  if (!pbft_block) {
    pbft_block = db_->getPbftCertVotedBlock(previous_round_next_voted_value_);
    if (pbft_block) {
      // PBFT unverified queue empty after node reboot, read from DB pushing back in unverified queue
      pbft_chain_->pushUnverifiedPbftBlock(pbft_block);
    }
  }`

so if you return bool flag you can reuse it and also print log errors based on it. 

And please use it everywhere if possible and delete checkPbftBlockValid_. Or if you want to keep it pls provide some valid reasoj for it

gaoqi8649: I agree with you. Fixed

JakubFornadel (approved):

PR #1118, opened 2021-08-30
Trx rlp cache

MatusKysel:
This PR enabled RLP cache for transactions. Transaction RLP serialization during syncing accounted for 28% of cpu usage and after this change just to 16%.
Before:
Screenshot_20210920_164554

After:
Screenshot_20210920_164702

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1845,7 +1845,10 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
 
 void PbftManager::syncBlockQueuePush(SyncBlock &&block, dev::p2p::NodeID const &node_id) {
   const auto period = block.pbft_blk->getPeriod();
-  if (period <= sync_period_) return;
+  if ((period != sync_period_ + 1) && sync_period_ != 0) {
+    LOG(log_er_) << "Trying to push block with " << period << " period, but current period is " << sync_period_;
+    return;
+  }
   sync_period_ = period;
   std::unique_lock lock(sync_queue_access_);
   sync_queue_.emplace_back(std::move(block), node_id);
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -104,7 +104,7 @@ void PbftManager::run() {
        period <= curr_period;  //
        ++period) {
     auto pbft_block = db_->getPbftBlock(period);
-    if (!pbft_block) {
+    if (!pbft_block.has_value()) {
       LOG(log_er_) << "DB corrupted - Cannot find PBFT block in period " << period << " in PBFT chain DB pbft_blocks.";
       assert(false);
     }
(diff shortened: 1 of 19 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -1442,8 +1442,6 @@ void PbftManager::pushSyncedPbftBlocksIntoChain_() {
         break;
       }
 
-      syncBlockQueuePop();
```
JakubFornadel, line 1445: This is not good as you just introduced the race condition by this change. There is relatively long time between `processSyncBlock()` and `pushPbftBlock_` when the the sync block is neither in sync_queue nor in database and there is some code that is checking current sync pbft period. This code wont work properly now... It was popped intentionally after  pushPbftBlock_

tests/final_chain_test.cpp
```diff
@@ -49,7 +49,7 @@ struct FinalChainTest : WithDataDir {
     db->saveDagBlock(dag_blk);
     PbftBlock pbft_block(blk_hash_t(), blk_hash_t(), blk_hash_t(), 1, addr_t(1), KeyPair::create().secret());
     std::vector<std::shared_ptr<Vote>> votes;
-    SyncBlock sync_block(pbft_block, votes);
+    SyncBlock sync_block(make_shared<PbftBlock>(std::move(pbft_block)), std::move(votes));
```
JakubFornadel, line 52: std::make_shared

src/final_chain/final_chain.cpp
```diff
@@ -92,21 +92,21 @@ class FinalChainImpl final : public FinalChain {
...
-      for (auto const& trx : period_transactions) {
+      DB::MultiGetQuery db_query(db_, sync_block.transactions.size());
+      for (auto const& trx : sync_block.transactions) {
         db_query.append(DB::Columns::final_chain_transaction_location_by_hash, trx.getHash());
       }
       auto trx_db_results = db_query.execute(false);
-      to_execute.reserve(period_transactions.size());
-      for (uint i = 0; i < period_transactions.size(); ++i) {
+      to_execute.reserve(sync_block.transactions.size());
+      for (uint i = 0; i < sync_block.transactions.size(); ++i) {
```
JakubFornadel, line 101: size_t, not uint

JakubFornadel (changes requested):
See my first comment, which describes a major problem in the changes

src/consensus/pbft_manager.hpp
```diff
@@ -237,7 +237,7 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
   std::condition_variable stop_cv_;
   std::mutex stop_mtx_;
 
-  std::queue<std::pair<SyncBlock, dev::p2p::NodeID>> sync_queue_;
+  std::deque<std::pair<SyncBlock, dev::p2p::NodeID>> sync_queue_;
```
JakubFornadel, line 240: why deque instead of queue ?

mfrankovi (approved):

src/consensus/pbft_manager.cpp
```diff
@@ -1855,15 +1843,18 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
   return std::optional<SyncBlock>(std::move(sync_block.first));
 }
 
-void PbftManager::syncBlockQueuePush(SyncBlock const &block, dev::p2p::NodeID const &node_id) {
+void PbftManager::syncBlockQueuePush(SyncBlock &&block, dev::p2p::NodeID const &node_id) {
+  const auto period = block.pbft_blk->getPeriod();
+  if (period <= sync_period_) return;
```
JakubFornadel, line 1848: - We should log error msg here
- I think the condition should not check <= but period should be excatly +1 ->  if (period != sync_period_ + 1)

src/consensus/sync_block.cpp
```diff
@@ -11,24 +11,22 @@ namespace taraxa {
 
 using namespace std;
 
-SyncBlock::SyncBlock(PbftBlock const& pbft_blk, std::vector<std::shared_ptr<Vote>> const& cert_votes)
-    : pbft_blk(new PbftBlock(pbft_blk)), cert_votes(cert_votes) {}
+SyncBlock::SyncBlock(std::shared_ptr<PbftBlock> pbft_blk, std::vector<std::shared_ptr<Vote>> cert_votes)
+    : pbft_blk(std::move(pbft_blk)), cert_votes(std::move(cert_votes)) {}
```
JakubFornadel, line 15: When I look at this, it seems that passing const ref to vector and then making a copy `cert_votes(cert_votes)` is better/cheaper than passing it as copy and then moving it ?

It is basically 1 reference + 1 copy ctor VS 1 copt ctr + 1 move ctor

JakubFornadel (approved):

src/consensus/sync_block.cpp
MatusKysel: it was making scene before but lot of changed since this PR

PR #1125, opened 2021-09-01
Syncing optimizations

mfrankovi:
Moving queueing syncing blocks from pbft to networking
Improvements to validation of syncing
Optimizing syncing to avoid data copy and calculating dag block order
Adding hash for dag order and unique transactions order in pbf block

### DIFF
--- a/src/consensus/pbft_chain.cpp
+++ b/src/consensus/pbft_chain.cpp
@@ -14,18 +14,22 @@ PbftBlock::PbftBlock(bytes const& b) : PbftBlock(dev::RLP(b)) {}
 PbftBlock::PbftBlock(dev::RLP const& r) {
   dev::RLP const rlp(r);
   if (!rlp.isList()) throw std::invalid_argument("PBFT RLP must be a list");
-  prev_block_hash_ = rlp[0].toHash<blk_hash_t>();
-  dag_block_hash_as_pivot_ = rlp[1].toHash<blk_hash_t>();
-  period_ = rlp[2].toInt<uint64_t>();
-  timestamp_ = rlp[3].toInt<uint64_t>();
-  signature_ = rlp[4].toHash<sig_t>();
+  auto it = rlp.begin();
+
+  prev_block_hash_ = (*it++).toHash<blk_hash_t>();
+  dag_block_hash_as_pivot_ = (*it++).toHash<blk_hash_t>();
+  order_hash_ = (*it++).toHash<blk_hash_t>();
+  period_ = (*it++).toInt<uint64_t>();
+  timestamp_ = (*it++).toInt<uint64_t>();
+  signature_ = (*it++).toHash<sig_t>();
   calculateHash_();
 }
 
-PbftBlock::PbftBlock(blk_hash_t const& prev_blk_hash, blk_hash_t const& dag_blk_hash_as_pivot, uint64_t period,
-                     addr_t const& beneficiary, secret_t const& sk)
+PbftBlock::PbftBlock(blk_hash_t const& prev_blk_hash, blk_hash_t const& dag_blk_hash_as_pivot,
+                     blk_hash_t const& order_hash, uint64_t period, addr_t const& beneficiary, secret_t const& sk)
     : prev_block_hash_(prev_blk_hash),
       dag_block_hash_as_pivot_(dag_blk_hash_as_pivot),
+      order_hash_(order_hash),
       period_(period),
       beneficiary_(beneficiary) {
   timestamp_ = dev::utcTime();
@@ -38,6 +42,7 @@ PbftBlock::PbftBlock(std::string const& str) {
   block_hash_ = blk_hash_t(doc["block_hash"].asString());
   prev_block_hash_ = blk_hash_t(doc["prev_block_hash"].asString());
   dag_block_hash_as_pivot_ = blk_hash_t(doc["dag_block_hash_as_pivot"].asString());
+  order_hash_ = blk_hash_t(doc["order_hash"].asString());
   period_ = doc["period"].asUInt64();
   timestamp_ = doc["timestamp"].asUInt64();
   signature_ = sig_t(doc["signature"].asString());
@@ -75,6 +80,7 @@ Json::Value PbftBlock::getJson() const {
   Json::Value json;
   json["prev_block_hash"] = prev_block_hash_.toString();
   json["dag_block_hash_as_pivot"] = dag_block_hash_as_pivot_.toString();
+  json["order_hash"] = order_hash_.toString();
   json["period"] = (Json::Value::UInt64)period_;
   json["timestamp"] = (Json::Value::UInt64)timestamp_;
   json["block_hash"] = block_hash_.toString();
@@ -85,9 +91,10 @@ Json::Value PbftBlock::getJson() const {
 
 // Using to setup PBFT block hash
 void PbftBlock::streamRLP(dev::RLPStream& strm, bool include_sig) const {
-  strm.appendList(include_sig ? 5 : 4);
+  strm.appendList(include_sig ? 6 : 5);
   strm << prev_block_hash_;
   strm << dag_block_hash_as_pivot_;
+  strm << order_hash_;
   strm << period_;
   strm << timestamp_;
   if (include_sig) {
--- a/src/consensus/pbft_chain.cpp
+++ b/src/consensus/pbft_chain.cpp
@@ -106,44 +106,6 @@ std::ostream& operator<<(std::ostream& strm, PbftBlock const& pbft_blk) {
   return strm;
 }
 
-PbftBlockCert::PbftBlockCert(PbftBlock const& pbft_blk, std::vector<Vote> const& cert_votes)
-    : pbft_blk(new PbftBlock(pbft_blk)), cert_votes(cert_votes) {}
-
-PbftBlockCert::PbftBlockCert(dev::RLP const& rlp) {
-  auto it = rlp.begin();
-  pbft_blk = make_shared<PbftBlock>(*it++);
-  for (auto const vote_rlp : *it++) {
-    cert_votes.emplace_back(vote_rlp);
-  }
-
-  for (auto const dag_block_rlp : *it++) {
-    DagBlock block(dag_block_rlp);
-    dag_blocks_per_level[block.getLevel()].emplace_back(block);
-  }
-
-  for (auto const trx_rlp : *it) {
-    auto trx = Transaction(trx_rlp);
-    transactions.emplace_back(trx);
-  }
-}
-
-PbftBlockCert::PbftBlockCert(bytes const& all_rlp) : PbftBlockCert(dev::RLP(all_rlp)) {}
-
-bytes PbftBlockCert::rlp() const {
-  RLPStream s(2);
-  s.appendRaw(pbft_blk->rlp(true));
-  s.appendList(cert_votes.size());
-  for (auto const& v : cert_votes) {
-    s.appendRaw(v.rlp(true));
-  }
-  return s.out();
-}
-
-std::ostream& operator<<(std::ostream& strm, PbftBlockCert const& b) {
-  strm << "[PbftBlockCert] : " << b.pbft_blk << " , num of votes " << b.cert_votes.size() << std::endl;
-  return strm;
-}
-
 PbftChain::PbftChain(blk_hash_t const& dag_genesis_hash, addr_t node_addr, std::shared_ptr<DbStorage> db)
     : head_hash_(blk_hash_t(0)),
       dag_genesis_hash_(dag_genesis_hash),
@@ -194,10 +156,6 @@ bool PbftChain::findUnverifiedPbftBlock(taraxa::blk_hash_t const& pbft_block_has
   return unverified_blocks_.find(pbft_block_hash) != unverified_blocks_.end();
 }
 
-bool PbftChain::findPbftBlockInSyncedSet(taraxa::blk_hash_t const& pbft_block_hash) const {
-  return pbft_synced_set_.find(pbft_block_hash) != pbft_synced_set_.end();
-}
-
 PbftBlock PbftChain::getPbftBlockInChain(const taraxa::blk_hash_t& pbft_block_hash) {
   auto pbft_block = db_->getPbftBlock(pbft_block_hash);
   if (pbft_block == nullptr) {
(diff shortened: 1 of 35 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -1271,9 +1271,51 @@ std::pair<blk_hash_t, bool> PbftManager::proposeMyPbftBlock_() {
 
   uint64_t propose_pbft_period = pbft_chain_->getPbftChainSize() + 1;
   addr_t beneficiary = node_addr_;
+
+  // get DAG block and transaction order
+  auto dag_block_order = dag_mgr_->getDagBlockOrder(dag_block_hash);
+  if (dag_block_order.second->empty()) {
+    LOG(log_er_) << "DAG anchor block hash " << dag_block_hash << " getDagBlockOrder failed in propose";
+    assert(false);
+    return std::make_pair(NULL_BLOCK_HASH, true);
```
JakubFornadel, line 1280: what is the logic behind return true on failures ? On line 1220 we return false for example...

src/consensus/pbft_manager.cpp
```diff
@@ -1590,50 +1614,70 @@ bool PbftManager::pushPbftBlock_(PbftBlockCert const &pbft_block_cert_votes, vec
...
+  if (!sync) {
+    std::unordered_set<trx_hash_t> trx_set;
+    std::vector<trx_hash_t> transactionsToQuery;
+    DbStorage::MultiGetQuery db_query(db_);
+    db_query.append(DbStorage::Columns::dag_blocks, dag_blocks_order);
+    auto dag_blocks_res = db_query.execute();
+
+    for (auto const &dag_blk_raw : dag_blocks_res) {
+      DagBlock dag_block(asBytes(dag_blk_raw));
+      for (auto const trx_hash : dag_block.getTrxs()) {
```
JakubFornadel, line 1665: maybe const auto& trx_hash ?

src/dag/dag.cpp
```diff
@@ -293,7 +295,8 @@ DagManager::DagManager(blk_hash_t const &genesis, addr_t node_addr, std::shared_
       db_(db),
       anchor_(genesis),
       period_(0),
-      genesis_(genesis) {
+      genesis_(genesis),
+      log_time_(log_time) {
```
JakubFornadel, line 299: what is this log_time good for ?

src/dag/dag.cpp
```diff
@@ -320,6 +323,8 @@ std::shared_ptr<DagManager> DagManager::getShared() {
 void DagManager::stop() {
   unique_lock lock(mutex_);
   trx_mgr_ = nullptr;
+  stopped_ = true;
+  block_worker_.join();
```
JakubFornadel, line 327: I think this should be: `if (block_worker_.joinable()) {  block_worker_.join(); }` + you should call stop() function in destructor as you are handling new thread and without explicitly calling `stop()` function (lets say in test) there would be a crash that might be hard to debug.

src/dag/dag.hpp
```diff
@@ -122,10 +122,12 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
 
   explicit DagManager(blk_hash_t const &genesis, addr_t node_addr, std::shared_ptr<TransactionManager> trx_mgr,
                       std::shared_ptr<PbftChain> pbft_chain, std::shared_ptr<DagBlockManager> dag_blk_mgr,
-                      std::shared_ptr<DbStorage> db);
+                      std::shared_ptr<DbStorage> db, logger::Logger log_time);
   virtual ~DagManager() = default;
   std::shared_ptr<DagManager> getShared();
+  void start();
   void stop();
+  void setNetwork(std::weak_ptr<Network> network) { network_ = move(network); }
```
JakubFornadel, line 130: let's use namespaces as it should be used so `move -> std::move`

src/dag/dag.hpp
```diff
@@ -194,14 +194,18 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
...
   blk_hash_t old_anchor_;  // anchor of the second to last period
   uint64_t period_;        // last period
   blk_hash_t genesis_;
   std::map<uint64_t, std::vector<blk_hash_t>> non_finalized_blks_;
-  std::map<uint64_t, std::vector<blk_hash_t>> finalized_blks_;
   DagFrontier frontier_;
+  std::atomic<bool> stopped_ = true;
+  std::thread block_worker_;
+
+  logger::Logger log_time_;
```
JakubFornadel, line 208: Cannot we use standard logger ?

src/network/network.hpp
```diff
@@ -53,6 +53,14 @@ class Network {
   bool pbft_syncing();
   uint64_t syncTimeSeconds() const;
 
+  uint64_t pbftSyncingPeriod() const;
+  void syncBlockQueuePop();
+  std::shared_ptr<std::pair<SyncBlock, dev::p2p::NodeID>> processSyncBlock();
```
JakubFornadel, line 58: Why does it return shared_ptr ? I dont see any copies that would justify using pointer. If it is here to check if there is some return value, let's use std::optional

src/network/taraxa_capability.hpp
```diff
@@ -146,6 +156,7 @@ struct TaraxaCapability : virtual CapabilityFace {
   util::ThreadPool syncing_tp_{1, false};
   SyncingState syncing_state_;
   PeersState peers_state_;
+  mutable std::shared_mutex sync_access_;
```
JakubFornadel, line 159: I would not know what is this mutex guarding without some checking. On the other hand if you group it with sync_queue_ (just move it's declaration next to it and maybe give some empty new line around) it would be clear straight away

src/network/taraxa_capability.cpp
```diff
@@ -1612,6 +1593,93 @@ void TaraxaCapability::broadcastPreviousRoundNextVotesBundle() {
...
+
+void TaraxaCapability::syncBlockQueuePop() {
+  unique_lock lock(sync_access_);
+  sync_queue_.pop();
+}
+
+void TaraxaCapability::handleMaliciousSyncBlock(NodeID const &id) {
+  syncing_state_.set_peer_malicious(id);
+  auto host = host_.lock();
+  if (host) host->disconnect(sync_queue_.front().second, p2p::UserReason);
```
JakubFornadel, line 1613: shouldn't you use `id` instead of `sync_queue_.front().second` ?

src/network/network.hpp
```diff
@@ -53,6 +53,14 @@ class Network {
...
+  void syncBlockQueuePush(SyncBlock const &block);
+  void clearSyncBlockQueue();
+  size_t syncBlockQueueSize() const;
+  void handleMaliciousSyncBlock(NodeID const &id);
```
JakubFornadel, line 62: So I went through the changes and I can see you moved `sync_queue_` from `pbft_manager` to `taraxa_capability` but I cant really see any added value in a way it is done (sure I might be wrong but this is how I understand the changes from quick review). Let me explain:

- I still think that the ultimate goal is to remove processing of sync blocks from pbft_manager and move it all somehow to tarcap. Then during syncing pbft_manager could be sort of "disabled"... 
- What we have here is that all the processing of sync blocks is still inside pbft_manager and as a result of moving `sync_queue_` into the tarcap is that you had to create 7 new methods inside network class as well as tarcap class, which is a lot of methods. 
- tarcap in fact does not really use `sync_queue_` except pushing blocks into it.  Only pbft_manager depends on it and use it internally. So if you leave the queue inside pbft_manager you can delete those 14 methods. I would keep maybe only adjusted `handleMaliciousSyncBlock()` function inside tarcap that would set malicious flag and disconnect the peer...

In summary, as we did not  remove also actual processing of sync blocks from pbft_manager, these changes seem like it complicated stuff rather then improved it ?

src/consensus/pbft_chain.cpp
```diff
@@ -14,18 +14,22 @@ PbftBlock::PbftBlock(bytes const& b) : PbftBlock(dev::RLP(b)) {}
 PbftBlock::PbftBlock(dev::RLP const& r) {
   dev::RLP const rlp(r);
   if (!rlp.isList()) throw std::invalid_argument("PBFT RLP must be a list");
-  prev_block_hash_ = rlp[0].toHash<blk_hash_t>();
-  dag_block_hash_as_pivot_ = rlp[1].toHash<blk_hash_t>();
-  period_ = rlp[2].toInt<uint64_t>();
-  timestamp_ = rlp[3].toInt<uint64_t>();
-  signature_ = rlp[4].toHash<sig_t>();
```
MatusKysel, line 21: btw no need to change that, if you accessing RLP elements in ascending order it is exactly same as iterator

src/consensus/pbft_manager.cpp
```diff
@@ -1271,9 +1271,51 @@ std::pair<blk_hash_t, bool> PbftManager::proposeMyPbftBlock_() {
...
+    auto dag_blk = dag_blk_mgr_->getDagBlock(blk_hash);
+    if (!dag_blk) {
+      LOG(log_er_) << "DAG anchor block hash " << dag_block_hash << " getDagBlock failed in propose for block "
+                   << blk_hash;
+      assert(false);
+      return std::make_pair(NULL_BLOCK_HASH, true);
+    }
+    order_stream << dag_blk->getHash();
+    std::vector<trx_hash_t> trx_hashes;
+    for (auto const &trx_hash : dag_blk->getTrxs()) {
```
MatusKysel, line 1295: why you looping this, when it returns vector of trx that you need

src/consensus/pbft_manager.cpp
```diff
@@ -1271,9 +1271,51 @@ std::pair<blk_hash_t, bool> PbftManager::proposeMyPbftBlock_() {
...
+        assert(false);
+      }
+    }
+  }
+  order_stream.appendList(non_executed_transactions.size());
+  for (auto const &trx_hash : non_executed_transactions) {
+    order_stream << trx_hash;
+  }
+
+  auto order_hash = dev::sha3(order_stream.out());
```
MatusKysel, line 1314: this whole section could be separate function

src/consensus/pbft_manager.cpp
```diff
@@ -1590,50 +1614,70 @@ bool PbftManager::pushPbftBlock_(PbftBlockCert const &pbft_block_cert_votes, vec
...
+  auto const &anchor_hash = sync_block.pbft_blk->getPivotDagBlockHash();
+  dag_mgr_->setDagBlockOrder(anchor_hash, pbft_period, dag_blocks_order, batch);
 
-  std::vector<Transaction> transactions;
-  transactions.reserve(transactions_res.size());
-  for (auto const &trx_raw : transactions_res) {
-    if (trx_raw.size() > 0) transactions.emplace_back(asBytes(trx_raw));
+  if (!sync) {
+    std::unordered_set<trx_hash_t> trx_set;
+    std::vector<trx_hash_t> transactionsToQuery;
```
MatusKysel, line 1658: naming

src/storage/db_storage.cpp
```diff
@@ -251,6 +251,30 @@ std::vector<std::shared_ptr<DagBlock>> DbStorage::getDagBlocksAtLevel(level_t le
...
+void DbStorage::updateDagBlockCounters(Batch& write_batch, vector<DagBlock> blks) {
+  // Lock is needed since we are editing some fields
+  lock_guard<mutex> u_lock(dag_blocks_mutex_);
+  for (auto const& blk : blks) {
+    auto level = blk.getLevel();
+    std::string blocks = getBlocksByLevel(level);
+    if (blocks == "") {
+      blocks = blk.getHash().toString();
+    } else {
+      blocks = blocks + "," + blk.getHash().toString();
```
MatusKysel, line 263: why we are storing it as string? isn't it better to have rlp array of block hashes or something?

src/consensus/pbft_chain.cpp
mfrankovi: Internally RLP does not have random access, so accessing by index actually internally makes it iterate multiple time

MatusKysel: yes, but it is not true if you doing it in ascending order, because it is doing same thing as you, https://github.com/Taraxa-project/taraxa-aleth/blob/taraxa_master/libdevcore/RLP.cpp#L66 you can see, it has internal iterator `m_lastEnd`, so no need to rewrite anything if your are accessing something like 1,2,3... or even 1,4,12 ...

src/consensus/pbft_manager.cpp
mfrankovi: I have removed the return after assert but I also do not understand the meaning of the flag in the return values of this method. Lets talk about this

src/dag/dag.cpp
mfrankovi: this log_time is some performance logger used across different classes. I am not sure if we actually use it for something. We need to check with everyone and remove all instances if it is not used

src/storage/db_storage.cpp
mfrankovi: Yes, this should be modified. I would do it later as a separate PR and change all the places we use this.

src/consensus/pbft_manager.cpp
```diff
@@ -1271,9 +1271,51 @@ std::pair<blk_hash_t, bool> PbftManager::proposeMyPbftBlock_() {
...
+    order_stream << dag_blk->getHash();
+    std::vector<trx_hash_t> trx_hashes;
+    for (auto const &trx_hash : dag_blk->getTrxs()) {
+      trx_hashes.emplace_back(trx_hash);
+    }
+    auto trx_statuses = db_->getTransactionStatus(trx_hashes);
+    for (uint32_t i = 0; i < trx_statuses.size(); i++) {
+      if (trx_statuses[i].state == TransactionStatusEnum::in_block) {
+        non_executed_transactions.emplace_back(trx_hashes[i]);
+      } else if (trx_statuses[i].state != TransactionStatusEnum::executed) {
```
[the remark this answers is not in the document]
mfrankovi: I am not sure I udnerstand this race condition. Lets talk about this

src/consensus/pbft_chain.cpp
mfrankovi: I see. Thanks

MatusKysel: you do not need to change it, just letting you know

src/network/network.hpp
mfrankovi: I have made some additional changes after your comment. There was some leftover processing in pbft manager that I also moved to tarcap. Network and tarcap have this issue where this maybe should be a single class since it is unlikely that we would have other capabilities. Currently only 4 methods are exposed in network that deal with the queue.

JakubFornadel: I dont see any changes that would move sync pbft blocks processing from pbft_manager to tarcap, so the comment from above still holds

src/network/sync_block.hpp
```diff
@@ -0,0 +1,32 @@
...
+struct Transaction;
+class DagBlock;
+
+class SyncBlock {
+ public:
+  SyncBlock(PbftBlock const& pbft_blk, std::vector<Vote> const& cert_votes);
+  SyncBlock(dev::RLP const& all_rlp);
+  SyncBlock(bytes const& all_rlp);
+
+  std::shared_ptr<PbftBlock> pbft_blk;
```
JakubFornadel, line 24: underscore at the end is used for private class members

Issue #1112, opened 2021-08-26
Enable LZ4 compression for p2p layer 


MatusKysel:
https://eips.ethereum.org/EIPS/eip-706

PR #1127, opened 2021-09-02
perf(net): enable compression of p2p communication

MatusKysel:
## Purpose

Enabled p2p LZ4 compression. This should resolve https://github.com/Taraxa-project/taraxa-node/issues/1112

## Notes
so now we will have 2 types of packet:

- 0 -> all packets below 500 bytes and what we are using now
- 1 -> compressed packet, that are using LZ4 compression


JakubFornadel:
So what are some stats ? How did you measure how much space it can save vs added computational overhead ?

MatusKysel:
> So what are some stats ? How did you measure how much space it can save vs added computational overhead ?

so I did not finished all scenarios, but syncing we are saving 35-80% of traffic while sacrificing ~1000 microseconds. Compressing  dag block or transaction is just inefficient,  that's why there is this threshold of 500 bytes.

JakubFornadel:
> > So what are some stats ? How did you measure how much space it can save vs added computational overhead ?
> 
> so I did not finished all scenarios, but syncing we are saving 35-80% of traffic while sacrificing ~1000 microseconds. Compressing dag block or transaction is just inefficient, that's why there is this threshold of 500 bytes.

Well that compression related to syncing packets is great :) I wonder if there is some way how to figure out what is the most effective threshold for deciding if packet should or should be compressed but I guess not...

JakubFornadel (approved):

kstdl:
What flow do we have on working with submodules? I can't find any info about it in doc directory
Because there is no PR in taraxa-aleth submodule and there is no way to comment it. Also this PR is points to a commit from `p2p-compress` branch, so if we will merge this we will have a link to commit from p2p-compress branch for taraxa-aleth submodule

kstdl (approved):

PR #1133, opened 2021-09-03
Fix issue with khan/pull-request-comment-trigger@master

agrebin:
## Purpose

This fixes an issue with the khan/pull-request-comment-trigger@master, which stopped working.
## For reviewers
The problem seems to be related to the node version installed on each ubuntu revision ubuntu-latest changes and so this may cause problems, found that ubuntu-16.04 works and fixing to this version.

## Version Notes ##
{ major | minor | patch }

### DIFF
--- a/.github/workflows/pr-actions.yml
+++ b/.github/workflows/pr-actions.yml
@@ -1,4 +1,4 @@
-name: prnet-actions
+name: prnet-actions 
 on:
   pull_request:
     types: [opened,reopened]
@@ -7,7 +7,7 @@ on:
 
 jobs:
   deploy:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-16.04
     steps:
       - uses: khan/pull-request-comment-trigger@master
         id: check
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-actions.yml
```diff
@@ -7,7 +7,7 @@ on:
 
 jobs:
   deploy:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-18.04
```
JakubFornadel, line 10: why not 20.04 ?

.github/workflows/pr-actions.yml
```diff
@@ -32,7 +32,7 @@ jobs:
           pull_request: '${{ steps.request.outputs.data }}'
         if: steps.check.outputs.triggered == 'true'
   redeploy:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-16.04
```
JakubFornadel, line 35: detto

PR #1134, opened 2021-09-06
chore(aleth): removed aleth submodule and moved relevant files to src/

MatusKysel:
## Purpose

This commit will remove aleth as submodule and just have all the files in main repo

Note: cpp-check now takes longer as it scans whole /src directory


JakubFornadel:
I totally agree that we moved it to the main repo as we are constantly editing it so  there is not much sense to have it as submodule. 1 note for discussion: maybe we could move it to some subfolder like `src/aleth` or something similar and create our internal library from it, so it would have separate cmake, etc... ? @kstdl is working on such modular approach already and this should be fairly easy to do it that way....

MatusKysel:
> I totally agree that we moved it to the main repo as we are constantly editing it so there is not much sense to have it as submodule. 1 note for discussion: maybe we could move it to some subfolder like `src/aleth` or something similar and create our internal library from it, so it would have separate cmake, etc... ? @kstdl is working on such modular approach already and this should be fairly easy to do it that way....

So should I moved it somewhere? or leave it for @kstdl to do it in his PR?

JakubFornadel:
> > I totally agree that we moved it to the main repo as we are constantly editing it so there is not much sense to have it as submodule. 1 note for discussion: maybe we could move it to some subfolder like `src/aleth` or something similar and create our internal library from it, so it would have separate cmake, etc... ? @kstdl is working on such modular approach already and this should be fairly easy to do it that way....
> 
> So should I moved it somewhere? or leave it for @kstdl to do it in his PR?

I dont know ping him and you guys decide :)

kstdl:
I think it should be in a separate PR. But I can make it after my current one to avoid some potential merge conflicts

MatusKysel:
@JakubFornadel and @kstdl I have moved it to src/aleth and created there cmakelist that handles aleth libraries

JakubFornadel (approved):

kstdl (approved):

PR #1144, opened 2021-09-09
refactor(pbft_manager): move queue with sync blocks into the pbft_mgr…

JakubFornadel:
## Purpose

So the only change here is that I moved(copy paste) sync_queue + methods related to processing data from this queue into the pbft_manager. Logic behind is this:

- the only thing that tarcap does is that it push data into this queue + check it's size. Similar to other queues, e.g. queues for tx_manager, dag_manager, etc...
- pbft_manager is working with data from this queue so I moved processing methods inside pbft_manager. In original PR they were inside tarcap and pbft_manager would simply call those functions from it's internal processing function.
- As a result 4 methods from intermediate network  "interface" could be deleted:
  ```
  uint64_t pbftSyncingPeriod() const;
  std::optional<SyncBlock> processSyncBlock();
  void syncBlockQueuePush(SyncBlock const &block);
  size_t syncBlockQueueSize() const;
  ```
I am totally for moving all syncing logic from pbft_manager somewhere else (even to tarcap), but I am for solution where processing of syncing data would not be triggered from pbft_manager so until we figure out how to get rid of the dependency of syncing data on pbft_manager, I think this is cleaner solution + similar to approach used in other managers and theirs queues...

To have a bigger picture about design, original PR must be checked. Check where is `processSyncBlock()` called

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1423,9 +1423,9 @@ void PbftManager::syncPbftChainFromPeers_(PbftSyncRequestReason reason, taraxa::
     return;
   }
   if (auto net = network_.lock()) {
-    if (net->syncBlockQueueSize()) {
+    if (syncBlockQueueSize()) {
       LOG(log_tr_) << "PBFT synced queue is still processing so skip syncing. Synced queue size "
-                   << net->syncBlockQueueSize();
+                   << syncBlockQueueSize();
 
       return;
     }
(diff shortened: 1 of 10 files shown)


MatusKysel (approved):

kstdl:
I don't think there is much difference in these two implementations. Since this data is pbft, it is slightly better to be in the pbft class.
But I also agree with its revision in future and I think that it is more global design problem with networking or/and some other modules

kstdl (approved):

JakubFornadel:
> I don't think there is much difference in these two implementations. Since this data is pbft, it is slightly better to be in the pbft class.
> But I also agree with its revision in future and I think that it is more global design problem with networking or/and some other modules

Yes you are right with both points

PR #1149, opened 2021-09-10
fix: update transaction status correctly

mfrankovi:
Any time transaction status is updated based on the current transaction status, get and update operation are now protected with a mutex inside of transaction manager to prevent a race condition where transactions status ends up in an incorrect state:
Example:
At the same time transaction arrives with gossip and with syncing in executed block:
1. Syncing thread asks for a transaction status and gets unknown
2. Gossip handling thread asks for a transaction status and gets unknown
3. Syncing thread updates the status to finalized
4. Gossip thread updates the status to in_queue

With the added mutex this scenario is not possible

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -23,14 +23,15 @@ PbftManager::PbftManager(PbftConfig const &conf, blk_hash_t const &genesis, addr
                          std::shared_ptr<DbStorage> db, std::shared_ptr<PbftChain> pbft_chain,
                          std::shared_ptr<VoteManager> vote_mgr,
                          std::shared_ptr<NextVotesForPreviousRound> next_votes_mgr, std::shared_ptr<DagManager> dag_mgr,
-                         std::shared_ptr<DagBlockManager> dag_blk_mgr, std::shared_ptr<FinalChain> final_chain,
-                         secret_t node_sk, vrf_sk_t vrf_sk)
+                         std::shared_ptr<DagBlockManager> dag_blk_mgr, std::shared_ptr<TransactionManager> trx_mgr,
+                         std::shared_ptr<FinalChain> final_chain, secret_t node_sk, vrf_sk_t vrf_sk)
     : db_(db),
       previous_round_next_votes_(next_votes_mgr),
       pbft_chain_(pbft_chain),
       vote_mgr_(vote_mgr),
       dag_mgr_(dag_mgr),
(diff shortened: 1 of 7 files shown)


src/storage/db_storage.cpp
```diff
@@ -338,14 +338,6 @@ void DbStorage::savePeriodData(const SyncBlock& sync_block, Batch& write_batch)
     // Remove dag blocks
     remove(write_batch, Columns::dag_blocks, toSlice(block.getHash()));
   }
-  uint64_t position = 0;
```
MatusKysel, line 341: this should not be moved, because this approach is more error prone

src/consensus/pbft_manager.cpp
```diff
@@ -1672,6 +1673,8 @@ bool PbftManager::pushPbftBlock_(SyncBlock &sync_block, vec_blk_t &dag_blocks_or
   // Commit DB
   db_->commitWriteBatch(batch);
```
MatusKysel, line 1674: I would just add the mutex here and but splitting this function is not good as we can forgot to remove transaction in the future

src/transaction_manager/transaction_manager.hpp
```diff
@@ -121,6 +124,8 @@ class TransactionManager : public std::enable_shared_from_this<TransactionManage
   addr_t node_addr_;
   std::shared_ptr<DagManager> dag_mgr_;
   logger::Logger log_time_;
+  // Guards updating transaction status based on retrieved status
+  std::shared_mutex transaction_status_mutex_;
```
MatusKysel, line 128: no need for shared_mux as we always locking for writing

MatusKysel (changes requested):
I do not like it, because if we are using mutex, we could have trx status as in memory container and not use DB for that. The only think that should be saved in DB is final position inside the block, where we know, that was finalized.

src/storage/db_storage.cpp
mfrankovi: > I do not like it, because if we are using mutex, we could have trx status as in memory container and not use DB for that. The only think that should be saved in DB is final position inside the block, where we know, that was finalized.

I do not see how this could be done without a mutex. I mean maybe the mutex is not in the right class, maybe it should be implemented in DbStorage but if we have threads running at the same time that in parallel want to update the status of transaction based on the current status, I do not see a different way to do this

justinsnapp (approved):

PR #1162, opened 2021-09-14
chore: remove some redudant copies

MatusKysel:


JakubFornadel (changes requested):
Lets wait for my branch to merged, you are touching a lot of stuff that I already refactored

src/network/taraxa_capability.cpp
```diff
@@ -576,7 +576,7 @@ void TaraxaCapability::interpretCapabilityPacketImpl(NodeID const &_nodeID, unsi
         std::vector<Vote> send_next_votes_bundle;
         for (auto const &v : next_votes_bundle) {
           if (!peer->isVoteKnown(v.getHash())) {
-            send_next_votes_bundle.emplace_back(v);
+            send_next_votes_bundle.push_back(std::move(v));
```
gaoqi8649, line 579: I think emplace_back(v) is some kind same with push_back(std::move(v)). What's different with them?

MatusKysel: so emplace should be used when you want to create object in place and insert it to the container. Here we want to move that object so we need to do `push_back(std::move(v))` or `emplace_back(std::move(v))` this should be same, but I prefer to use emplace only when we creating specific object

gaoqi8649: Thanks!

src/network/tarcap/packets_handlers/vote_packets_handler.cpp
```diff
@@ -182,7 +182,7 @@ inline void VotePacketsHandler::processPbftNextVotesPacket(const dev::RLP &packe
       std::vector<Vote> send_next_votes_bundle;
       for (auto const &v : next_votes) {
         if (!peer_to_share_to.second->isVoteKnown(v.getHash())) {
-          send_next_votes_bundle.emplace_back(v);
+          send_next_votes_bundle.push_back(std::move(v));
```
JakubFornadel, line 185: btw Vote class does not support move semantics. This is true for almost all(if not literally all) of our basic classes like Transaction, DagBlock, etc... So we are in fact doing copies almost everywhere. We should refactor those classes to actually gain something by using std::move

JakubFornadel (approved):

PR #1171, opened 2021-09-21
fix: dag order

mfrankovi:
When a pbft block with incorrect order hash was pushed into a chain the push would fail but at the same time dag blocks within the pbft block were removed from the non-finalized set with method setDagBlockOrder, breaking the DAG and network got stuck. The check for incorrect order hash is now moved before calling setDagBlockOrder

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1614,42 +1614,6 @@ bool PbftManager::pushPbftBlock_(SyncBlock &sync_block, vec_blk_t &dag_blocks_or
   auto const &cert_votes = sync_block.cert_votes;
   auto pbft_period = sync_block.pbft_blk->getPeriod();
 
-  auto batch = db_->createWriteBatch();
-  dag_blk_mgr_->processSyncedBlock(batch, sync_block);
-
-  LOG(log_nf_) << "Storing cert votes of pbft blk " << pbft_block_hash;
-  LOG(log_dg_) << "Stored following cert votes:\n" << cert_votes;
-  // Update PBFT chain head block
-  db_->addPbftHeadToBatch(pbft_chain_->getHeadHash(), pbft_chain_->getJsonStrForBlock(pbft_block_hash), batch);
-
-  if (sync) {
-    dag_blocks_order.reserve(sync_block.dag_blocks.size());
-    std::transform(sync_block.dag_blocks.begin(), sync_block.dag_blocks.end(), std::back_inserter(dag_blocks_order),
-                   [](const DagBlock &dag_block) { return dag_block.getHash(); });
-    // Update counts correctly
-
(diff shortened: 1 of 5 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -1696,10 +1663,53 @@ bool PbftManager::pushPbftBlock_(SyncBlock &sync_block, vec_blk_t &dag_blocks_or
     if (calculated_order_hash != sync_block.pbft_blk->getOrderHash()) {
       LOG(log_er_) << "Order hash incorrect. Pbft order hash: " << sync_block.pbft_blk->getOrderHash()
                    << " . Calculated hash:" << calculated_order_hash;
+      std::vector<trx_hash_t> trx_order;
+      std::vector<blk_hash_t> blk_order;
```
MatusKysel, line 1667: reserve size

src/consensus/pbft_manager.cpp
```diff
@@ -1929,6 +1939,16 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
     LOG(log_er_) << "Order hash incorrect in sync block " << pbft_block_hash << " expected: " << order_hash
                  << " received " << sync_block.first.pbft_blk->getOrderHash() << " from "
                  << sync_block.second.abridged() << ", stop syncing.";
+    std::vector<trx_hash_t> trx_order;
+    std::vector<blk_hash_t> blk_order;
```
MatusKysel, line 1943: same here, I know it is just code for logging, but still

MatusKysel (approved):

gaoqi8649 (approved):

JakubFornadel (approved):

src/consensus/pbft_manager.cpp
```diff
@@ -1696,10 +1663,53 @@ bool PbftManager::pushPbftBlock_(SyncBlock &sync_block, vec_blk_t &dag_blocks_or
...
+  if (sync) {
+    dag_blocks_order.reserve(sync_block.dag_blocks.size());
+    std::transform(sync_block.dag_blocks.begin(), sync_block.dag_blocks.end(), std::back_inserter(dag_blocks_order),
+                   [](const DagBlock &dag_block) { return dag_block.getHash(); });
+    // Update counts correctly
+
+    // Non-finalized block should be empty when syncing, maybe we should clear it if we are deep out of sync to improve
+    // performance
+    auto non_finalized_blocks = dag_mgr_->getNonFinalizedBlocks();
+    unordered_set<blk_hash_t> non_finalized_blocks_set;
```
JakubFornadel, line 1694: std::unordered_set. It is considered bad practice to have `using namespace std` and skip using `std::`. I am planning to remove this from our code so you can save me some time and write new code with std::  here and in other places

PR #1172, opened 2021-09-21
Dag blocks processing refactor

JakubFornadel:
## Version Notes ##
{ major | minor | patch }


src/util/util.hpp
```diff
@@ -231,14 +229,29 @@ class ExpirationCache {
...
+      if (cache_.count(key)) {
+        return false;
+      }
+    }
 
-    if (cache_.find(key) != cache_.end()) {
+    boost::unique_lock lock(mtx_);
+
+    // There must be double check if key is not already in cache due to possible race condition
+    if (!cache_.insert(key).second) {
```
MatusKysel, line 251: you can get rid of this with upgradable lock

src/util/util.hpp
```diff
@@ -287,11 +300,30 @@ class ExpirationCacheMap {
...
+      boost::shared_lock lock(mtx_);
+      if (cache_.count(key)) {
+        return false;
+      }
+    }
+
+    boost::unique_lock lock(mtx_);
+
+    // There must be double check if key is not already in cache due to possible race condition
+    if (!cache_.emplace(key, value).second) {
```
MatusKysel, line 323: same here

src/util/util.hpp
JakubFornadel: nope, I did check boost documentation properly and there is a huge difference between upgradable + upgrade_to_unique lock vs sharedlock + unique lock. ONLY 1 SINGLE UPGRADABLE LOCK AT THE SAME TIME CAN BE ACQUIRED, so if this is the method that is called very often, it behaves almost as unique_lock. And because this is called very often, it would block other threads even in case some element is already in cache. In shared_lock + unique_lock if the element is already in cache it will not block other threads because  you can acquire as many shared_locks as you wish at the same time

MatusKysel: but with this there is edge scenario when you pass first part of scope and then other thread calls count() and return size-1 as you did not do insert meanwhile

JakubFornadel: I dont understand what scenario are you talking about ? but I dont think there is any edge case, insertion is guarded by unique_lock. The first part with shared_lock is purely as performance optimization, it basically does not even need to be there and it would behave in the same way but worse performance

MatusKysel (approved):

src/network/tarcap/packets_handlers/common/packet_handler.cpp
```diff
@@ -64,7 +64,7 @@ bool PacketHandler::sealAndSend(const dev::p2p::NodeID& nodeID, SubprotocolPacke
     peer = peers_state_->getPendingPeer(nodeID);
 
     if (!peer) {
-      LOG(log_wr_) << "sealAndSend failed to find peer";
+      LOG(log_er_) << "sealAndSend failed to find peer";
```
MatusKysel, line 67: why error? peer could disconnect meanwhile, so it is not an error for this node but for his more likely

mfrankovi: I will approve the PR but besides here I also see a couple of other places where errors are logged for a case when peer might have just disconnected

mfrankovi (approved):

src/network/tarcap/packets_handlers/common/packet_handler.cpp
JakubFornadel: I want to see if it happens a lot, if it is warning we will never know as we wont have enabled channels for every packet handler

src/network/tarcap/threadpool/packets_blocking_mask.cpp
```diff
@@ -1,90 +1,178 @@
...
-  // blocked_packets_peers_time_ contains peers_time blocks that are not compatible with hard blocks, we should
-  // not try to create hard block (and later unblock) when there is ongoing peers_time block
-  assert(blocked_packets_peers_time_.find(packet_type) == blocked_packets_peers_time_.end());
+void PacketsBlockingMask::markPacketAsHardBlocked(const PacketData& blocking_packet,
+                                                  SubprotocolPacketType packet_type_to_block) {
+  auto& packet_hard_block = hard_blocked_packet_types_[packet_type_to_block];
+
+  // If this assert is triggered, it means we are trying to insert blocking packet id that is already inserted, which
+  // should never happen as packets id's are supposed to be unique
+  assert(packet_hard_block.insert(blocking_packet.id_).second);
```
MatusKysel, line 13: This will be removed by release build, huge bug

src/network/tarcap/packets_handlers/common/packet_handler.cpp
JakubFornadel: @mfrankovi If those errors should not be errors then do not approve such PR and write comments. If you both think this should be warnng I will change it

JakubFornadel: changed to wr

src/network/tarcap/threadpool/packets_blocking_mask.cpp
```diff
@@ -1,90 +1,178 @@
...
-  assert(blocked_packets_types_mask_ & packet_type);
-  blocked_packets_types_mask_ ^= packet_type;
+void PacketsBlockingMask::markPacketAsPeerOrderBlocked(const PacketData& blocking_packet,
+                                                       SubprotocolPacketType packet_type_to_block) {
+  auto& packet_peer_order_blocks = peer_order_blocked_packet_types_[packet_type_to_block];
+  auto& peer_order_blocks = packet_peer_order_blocks[blocking_packet.from_node_id_];
+
+  // If this assert is triggered, it means we are trying to insert blocking packet id that is already inserted, which
+  // should never happen as packets id's are supposed to be unique
+  assert(peer_order_blocks.insert(blocking_packet.id_).second);
```
MatusKysel, line 41: Same here

src/network/tarcap/threadpool/packets_blocking_mask.cpp
JakubFornadel: kind of agree, but do you realize that we have tens of asserts throughout the whole codebase with basically the same bug as here ? So all of them are basically huge bugs and we should not use asserts at all

JakubFornadel: Didn't get the comment at first, you are right !

MatusKysel (approved):

PR #1175, opened 2021-09-22
chore(final_chain): fix cyclic dependency issues

MatusKysel:


src/network/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -41,7 +40,11 @@ void StatusPacketHandler::process(const dev::RLP& packet_rlp, const PacketData&
       if (!pending_peer) {
         LOG(log_er_) << "Peer " << packet_data.from_node_id_.abridged()
                      << " missing in both peers and pending peers map - will be disconnected.";
-        host->disconnect(packet_data.from_node_id_, p2p::UserReason);
+        if (auto host_shr = host.lock(); host_shr) {
+          host_shr->disconnect(packet_data.from_node_id_, p2p::UserReason);
+        } else {
+          LOG(log_er_) << "Invalid host in StatusPacketHandler";
```
JakubFornadel, line 46: every packet handler has its own channel so there is not really any need to specify "StatusPacketHandler".

src/network/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -66,21 +69,33 @@ void StatusPacketHandler::process(const dev::RLP& packet_rlp, const PacketData&
                    << getFormattedVersion({node_major_version, node_minor_version, node_patch_version})
                    << ", our node version" << TARAXA_VERSION << ", host " << packet_data.from_node_id_.abridged()
                    << " will be disconnected";
-      host->disconnect(packet_data.from_node_id_, dev::p2p::UserReason);
+      if (auto host_shr = host.lock(); host_shr) {
+        host_shr->disconnect(packet_data.from_node_id_, p2p::UserReason);
+      } else {
+        LOG(log_er_) << "Invalid host in StatusPacketHandler";
+      }
```
JakubFornadel, line 76: hmm weak_ptr<Host> is part of peers_state class, which is available for every packet handler, let's remove `host` argument from all `process` functions. and use host from `peers_state`

JakubFornadel:
I still dont understand how exactly are those segfaults happening but if it is related to the circular dependency that is somehow related to tha shared_ptr<Host>, this seems like it does not really fix it, it only reduces the time during which we hold shared_ptr<Host> inside packet handlers so it should not happen that often but the bug is still there isnt it ?

src/final_chain/final_chain.cpp
```diff
@@ -37,7 +37,7 @@ class FinalChainImpl final : public FinalChain {
 
  public:
   FinalChainImpl(shared_ptr<DB> const& db, Config const& config, addr_t const& node_addr)
-      : db_(db),
+      : db_(std::move(db)),
```
JakubFornadel, line 40: std move from reference to the shared_ptr ? Does not seem ok

MatusKysel:
> I still dont understand how exactly are those segfaults happening but if it is related to the circular dependency that is somehow related to the shared_ptr< Host >, this seems like it does not really fix it, it only reduces the time during which we hold shared_ptr< Host > inside packet handlers so it should not happen that often but the bug is still there isnt it ?

Yes, we have bad design so we end up with cyclic dependencies and this is only easy way to mitigate this "bad" behavior in our tests. The best would be to change our architecture so this kind of things never happened

JakubFornadel (approved):

JakubFornadel:
Merging this through admin rights. We were talking about this on call that we need it asap so people should do review. Nobody did....

PR #1176, opened 2021-09-23
fix: dag order

mfrankovi:
Fix for dag order hash verification. Similar fix is already applied to testnet branch. This is the same fix but with moving the verification inside existing comparePbftBlockScheduleWithDAGblocks_ method

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1202,6 +1202,9 @@ std::pair<blk_hash_t, bool> PbftManager::proposeMyPbftBlock_() {
     }
   }
 
+  LOG(log_nf_) << "DAG order for proposed block" << dag_block_order.second;
+  LOG(log_nf_) << "Transaction order for proposed block" << non_executed_transactions;
+
   auto order_hash = calculateOrderHash(dag_block_order.second, non_executed_transactions);
 
   // generate generate pbft block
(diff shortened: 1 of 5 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -1367,18 +1370,76 @@ bool PbftManager::comparePbftBlockScheduleWithDAGblocks_(blk_hash_t const &pbft_
   if (!pbft_block) {
     return false;
   }
-
-  return comparePbftBlockScheduleWithDAGblocks_(*pbft_block).second;
+  return comparePbftBlockScheduleWithDAGblocks_(pbft_block).has_value();
```
MatusKysel, line 1373: std::move on block

src/consensus/pbft_manager.cpp
```diff
@@ -1367,18 +1370,76 @@ bool PbftManager::comparePbftBlockScheduleWithDAGblocks_(blk_hash_t const &pbft_
...
+    return std::move(dag_blocks_order);
+  }
+  auto const &anchor_hash = pbft_block->getPivotDagBlockHash();
   auto dag_blocks_order = dag_mgr_->getDagBlockOrder(anchor_hash).second;
   if (!dag_blocks_order.empty()) {
-    return std::make_pair(std::move(dag_blocks_order), true);
+    std::unordered_set<trx_hash_t> trx_set;
+    std::vector<trx_hash_t> transactions_to_query;
+    DbStorage::MultiGetQuery db_query(db_);
+    db_query.append(DbStorage::Columns::dag_blocks, dag_blocks_order);
```
MatusKysel, line 1391: isn't is better to get them from dag_blk_mgr_->getDagBlock? I think we can be 99,9% sure those block are still cached at this time

src/consensus/pbft_manager.cpp
```diff
@@ -1367,18 +1370,76 @@ bool PbftManager::comparePbftBlockScheduleWithDAGblocks_(blk_hash_t const &pbft_
...
+    }
+
+    auto calculated_order_hash = calculateOrderHash(dag_blocks_order, non_executed_transactions);
+    if (calculated_order_hash != pbft_block->getOrderHash()) {
+      LOG(log_er_) << "Order hash incorrect. Pbft order hash: " << pbft_block->getOrderHash()
+                   << " . Calculated hash:" << calculated_order_hash << ". Dag order: " << dag_blocks_order
+                   << ". Trx order: " << non_executed_transactions;
+      return {};
+    }
+    cert_sync_block_.pbft_blk = pbft_block;
```
MatusKysel, line 1437: std::move

src/consensus/pbft_manager.cpp
mfrankovi: yes, and I added the change for the other two comments

gaoqi8649 (approved):

justinsnapp (approved):

PR #1177, opened 2021-09-23
fix: finalchain destructions

mfrankovi:
Check that final chain thread is stopped before destroying objects

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -85,6 +85,7 @@ void PbftManager::stop() {
     stop_cv_.notify_all();
   }
   daemon_->join();
+  final_chain_->stop();
 
   LOG(log_dg_) << "PBFT daemon terminated ...";
 }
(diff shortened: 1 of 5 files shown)


MatusKysel (approved):

JakubFornadel (commented):
Nice. I think you can even delete these timeouts, which are not necessary anymore:

https://github.com/Taraxa-project/taraxa-node/blob/develop/tests/network_test.cpp#L195
https://github.com/Taraxa-project/taraxa-node/blob/develop/tests/full_node_test.cpp#L1140

JakubFornadel (changes requested):
See my comment

mfrankovi:
> Nice. I think you can even delete these timeouts, which are not necessary anymore:
> 
> https://github.com/Taraxa-project/taraxa-node/blob/develop/tests/network_test.cpp#L195
> https://github.com/Taraxa-project/taraxa-node/blob/develop/tests/full_node_test.cpp#L1140

Deleted the timeouts and tests seem to pass fine

gaoqi8649 (approved):

JakubFornadel (approved):

JakubFornadel (approved):

Issue #1038, opened 2021-08-02
Remove NewBlockHashPacket/GetNewBlockPacket packets 


JakubFornadel:
## Is your feature request related to a problem? Please describe.

I think that in general we could get rid of whole "NewBlockHashPacket" packet. When I checked it is only send inside "onNewBlockVerified" to "in some way randomized" set of peers that do not have this block already(or at least we did not mark it as known fro them) and I am not sure what is the purpose of it ? 

If such peer received only block hash but does not have it locally it will instantly request our node to send him full block. If we dont send "NewBlockHashPacket"(why not send full block straight away ?), we could get rid also of "GetNewBlockPacket", which  is sent only if we received "NewBlockHashPacket", but we have not received full block yet. 

Seems to be like a lot of unnecessary code and complexity.... @justinsnapp @mfrankovi you might know ?

MatusKysel:
I think this should be closed as it is saving quite a lot of network bandwidth

PR #1178, opened 2021-09-23
chore: remove dag block hash packets

mfrankovi:
Removing functionality of sending only dag block hashes as part of gossiping as it makes it very difficult to synchronize dag blocks order. This functionality does not make sense in an environment where a lot of dag blocks are proposed in parallel in short amount of time.

### DIFF
--- a/src/network/tarcap/packets_handlers/dag_packets_handler.cpp
+++ b/src/network/tarcap/packets_handlers/dag_packets_handler.cpp
@@ -29,15 +29,6 @@ thread_local mt19937_64 DagPacketsHandler::urng_{std::mt19937_64(std::random_dev
 void DagPacketsHandler::process(const dev::RLP &packet_rlp, const PacketData &packet_data,
 
                                 const std::shared_ptr<TaraxaPeer> &peer) {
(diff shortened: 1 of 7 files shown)


MatusKysel:
resolves https://github.com/Taraxa-project/taraxa-node/issues/1038

src/network/tarcap/packets_handlers/dag_packets_handler.cpp
```diff
@@ -34,10 +31,6 @@ void DagPacketsHandler::process(const dev::RLP &packet_rlp, const PacketData &pa
                                 const std::shared_ptr<TaraxaPeer> &peer) {
   if (packet_data.type_ == PriorityQueuePacketType::kPqNewBlockPacket) {
     processNewBlockPacket(packet_rlp, packet_data, peer);
```
MatusKysel, line 33: no need to call this function, probably just copy everything here as it is for other handlers

mfrankovi: Done. Also removed some additional unused code related to the change.

MatusKysel (approved):

JakubFornadel (approved):

PR #1180, opened 2021-09-24
Improve dag status

MatusKysel:
## Purpose

This PR improves DAG block status handling and completely removes redundant DB data. It is based on this PR https://github.com/Taraxa-project/taraxa-node/pull/1118 this should be merged before!!!

"dag_block_period": 38520098,
    "dag_blocks": 2880577,
    "dag_blocks_index": 31480472,
  REMOVED  ----> "dag_blocks_state": 2028541,   <---- REMOVED
  REMOVED  ----> "dag_finalized_blocks": 27889837,   <---- REMOVED
    "default": 0,
    "dpos_proposal_period_levels_status": 3301,
    "final_chain_blk_by_number": 7247123,
    "final_chain_blk_hash_by_number": 745559,
    "final_chain_blk_number_by_hash": 744199,
    "final_chain_log_blooms_index": 4164726,
    "final_chain_meta": 3253,
    "final_chain_receipt_by_trx_hash": 229678726,
    "final_chain_replay_protection": 0,
    "final_chain_transaction_count_by_blk_number": 391413,
    "final_chain_transaction_hashes_by_blk_number": 140407817,
    "final_chain_transaction_location_by_hash": 206590064,
    "next_votes": 104382,
    "pbft_block_period": 744178,
    "pbft_cert_voted_block": 2955383,
    "pbft_head": 4585,
    "pbft_mgr_round_step": 3321,
    "pbft_mgr_status": 3405,
    "pbft_mgr_voted_value": 3705,
    "pbft_round_2t_plus_1": 418042,
    "period_data": 1503914655,
    "proposal_period_levels_map": 439397,
    "soft_votes": 80270,
    "status": 3555,
    "transactions": 3325643,
    "trx_status": 211065444,
    "unverified_votes": 15561489,
    "verified_votes": 8568994

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -113,8 +113,7 @@ void PbftManager::run() {
                    << pbft_block->getPeriod() << " in block data than in block order db: " << period;
       assert(false);
     }
-    finalize_(*pbft_block, db_->getFinalizedDagBlockHashesByAnchor(pbft_block->getPivotDagBlockHash()),
-              period == curr_period);
+    finalize_(*pbft_block, db_->getFinalizedDagBlockHashesByPeriod(period), period == curr_period);
   }
 
   // Initialize PBFT status
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1380,7 +1380,7 @@ std::optional<vec_blk_t> PbftManager::comparePbftBlockScheduleWithDAGblocks_(std
     dag_blocks_order.reserve(cert_sync_block_.dag_blocks.size());
     std::transform(cert_sync_block_.dag_blocks.begin(), cert_sync_block_.dag_blocks.end(),
                    std::back_inserter(dag_blocks_order), [](const DagBlock &dag_block) { return dag_block.getHash(); });
-    return std::move(dag_blocks_order);
+    return {std::move(dag_blocks_order)};
   }
   auto const &anchor_hash = pbft_block->getPivotDagBlockHash();
   auto dag_blocks_order = dag_mgr_->getDagBlockOrder(anchor_hash).second;
@@ -1435,7 +1435,7 @@ std::optional<vec_blk_t> PbftManager::comparePbftBlockScheduleWithDAGblocks_(std
     }
     cert_sync_block_.pbft_blk = std::move(pbft_block);
 
-    return std::move(dag_blocks_order);
+    return {std::move(dag_blocks_order)};
   }
   syncPbftChainFromPeers_(missing_dag_blk, anchor_hash);
   return {};
@@ -1580,7 +1580,7 @@ bool PbftManager::pushPbftBlock_(SyncBlock &sync_block, vec_blk_t &dag_blocks_or
 
   // Set DAG blocks period
   auto const &anchor_hash = sync_block.pbft_blk->getPivotDagBlockHash();
-  dag_mgr_->setDagBlockOrder(anchor_hash, pbft_period, dag_blocks_order, batch);
+  dag_mgr_->setDagBlockOrder(anchor_hash, pbft_period, dag_blocks_order);
 
   db_->savePeriodData(sync_block, batch);
 
--- a/src/dag/dag.cpp
+++ b/src/dag/dag.cpp
@@ -510,7 +510,6 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
                                   DbStorage::Batch &write_batch) {
   uLock lock(mutex_);
   LOG(log_dg_) << "setDagBlockOrder called with anchor " << new_anchor << " and period " << period;
-  db_->putFinalizedDagBlockHashesByAnchor(write_batch, new_anchor, dag_order);
   if (period != period_ + 1) {
     LOG(log_er_) << " Inserting period (" << period << ") anchor " << new_anchor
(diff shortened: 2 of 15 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -1487,8 +1487,6 @@ void PbftManager::pushSyncedPbftBlocksIntoChain_() {
         break;
       }
 
-      syncBlockQueuePop();
```
gaoqi8649, line 1490: I remember Jakub mentioned that we cannot remove this.
For example, sync_queue_ has only one PBFT block period 11. If we pop the block from sync_queue_ immediately. And has not push into PBFT chain. That will cause sync the block period 11 again from peers and push into sync_queue_.

src/consensus/pbft_manager.cpp
```diff
@@ -1435,14 +1434,14 @@ std::optional<vec_blk_t> PbftManager::comparePbftBlockScheduleWithDAGblocks_(std
...
-    return std::move(dag_blocks_order);
+    return {std::move(dag_blocks_order)};
   }
   syncPbftChainFromPeers_(missing_dag_blk, anchor_hash);
   return {};
 }
 
 bool PbftManager::pushCertVotedPbftBlockIntoChain_(taraxa::blk_hash_t const &cert_voted_block_hash,
-                                                   std::vector<std::shared_ptr<Vote>> const &cert_votes_for_round) {
+                                                   std::vector<std::shared_ptr<Vote>> &&cert_votes_for_round) {
```
gaoqi8649, line 1444: I don't understand here. &&cert_votes_for_round is RValue, and std::move is for LValue. But you use "std::move(cert_votes_for_round)" is std::move for a RValue here. Could you explain that? Thanks

src/consensus/pbft_manager.hpp
```diff
@@ -238,7 +238,8 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
   std::condition_variable stop_cv_;
   std::mutex stop_mtx_;
 
-  std::queue<std::pair<SyncBlock, dev::p2p::NodeID>> sync_queue_;
+  std::deque<std::pair<SyncBlock, dev::p2p::NodeID>> sync_queue_;
+  std::atomic<uint64_t> sync_period_{0};
```
gaoqi8649, line 242: The last synced period is sync_queue_.back().first.pbft_blk->getPeriod(), don't need extra sync_period_

src/consensus/pbft_manager.cpp
MatusKysel: those changes are part of https://github.com/Taraxa-project/taraxa-node/pull/1118

src/consensus/pbft_manager.cpp
MatusKysel: https://stackoverflow.com/questions/31213539/c-stdmove-with-rvalue-reference-is-not-moving-contents

src/dag/dag.cpp
```diff
@@ -443,9 +442,9 @@ void DagManager::addToDag(blk_hash_t const &hash, blk_hash_t const &pivot, std::
 
 bool DagManager::getLatestPivotAndTips(blk_hash_t &pivot, std::vector<blk_hash_t> &tips) const {
   // make sure the state of dag is the same when collection pivot and tips
-  sharedLock lock(mutex_);
   pivot.clear();
   tips.clear();
+  sharedLock lock(mutex_);
```
JakubFornadel, line 447: - It seems like this lock is protecting data that are read inside getFrontier() function -> shouldn't be the lock there ? Or is this some sort of optimization that we dont want to use locks for every call to getFrontier ? 
- as sharedLock is alias to type, it should be with capital first letter

src/dag/dag.cpp
```diff
@@ -443,9 +442,9 @@ void DagManager::addToDag(blk_hash_t const &hash, blk_hash_t const &pivot, std::
 
 bool DagManager::getLatestPivotAndTips(blk_hash_t &pivot, std::vector<blk_hash_t> &tips) const {
```
JakubFornadel, line 443: These ref arguments dont make much sense because getFrontier() is returning copies anyway. I think it should return something like:
`std::optional<std::pair<blk_hash_t, std::vector<blk_hash_t>>>`
?

src/dag/dag.cpp
MatusKysel: getFrontier is used also in different function, that is under uniqe lock, so I did not changed how we are locking data. 
and I will rename it for dag.cpp

src/dag/dag.cpp
MatusKysel: good point I will change this

src/dag/dag.cpp
```diff
@@ -579,26 +562,8 @@ void DagManager::recoverDag() {
...
-      addDagBlock(*blk, true, false);
-    }
-  }
-  for (auto &it : non_finalized) {
-    for (auto &blk_hash : it.second) {
-      auto blk = dag_blk_mgr_->getDagBlock(blk_hash);
-      addDagBlock(*blk, false, false);
-    }
+  for (const auto &blk : db_->getNonfinalizedDagBlocks()) {
+    addDagBlock(std::move(blk), false);
```
JakubFornadel, line 566: what is this move good for when addDagBlock has const ref as argument ?

JakubFornadel (approved):
Seems ok to me. I think @mfrankovi might also look at this PR as he knows this part of code much better

mfrankovi (approved):

PR #1185, opened 2021-09-27
fix(pbft): cannot get unverified PBFT block in queue because has been added into chain already

gaoqi8649:


src/consensus/pbft_manager.cpp
```diff
@@ -895,12 +895,9 @@ void PbftManager::certifyBlock_() {
       }
 
       if (executed_soft_voted_block_for_this_round || unverified_soft_vote_block_for_this_round_is_valid) {
-        // generate cert vote
-
         // comparePbftBlockScheduleWithDAGblocks_ has checked the cert voted block exist
-
         last_cert_voted_value_ = soft_voted_block_for_this_round_.first;
-        auto cert_voted_block = pbft_chain_->getUnverifiedPbftBlock(last_cert_voted_value_);
+        auto cert_voted_block = getUnfinalizedBlock_(last_cert_voted_value_);
```
justinsnapp, line 900: So this works because we check both unfinalized block queue and the db.   My concern here is that a node can actually vote to certify a block that is in its chain already?  Or do we check that elsehwere?

gaoqi8649: The race condition like, the node has get 2t+1 cert votes before it cert vote the value.
Even like that, I still think the node should continue to cert vote on the value, because that will make other nodes who have not get 2t+1 cert votes that make consensus faster

justinsnapp: ok

PR #1189, opened 2021-09-28
fix(pbft): fix last period

gaoqi8649:
Matus add a new variable sync_period_. But sync_period_ only update when push syncing PBFT blocks. When get enough cert votes to push proposed gossiping PBFT blocks, the sync_period_ doesn't update. 
For example: node syncing period 1 to 10 and catched up. Then push proposed gossiping PBFT blocks period from 11 - 15 by getting enough cert votes. Then, if want to push syncing PBFT block period 16, that will fail because sync_period_ is still 10. 
I remove the sync_period_. Since I think that's better using the last block in synced queue/PBFT chain to instruct the last period

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1763,7 +1763,7 @@ bool PbftManager::is_syncing_() {
 }
 
 std::optional<SyncBlock> PbftManager::processSyncBlock() {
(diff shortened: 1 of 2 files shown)


JakubFornadel (changes requested):
This is rather revert than a fix. Proper fix is implemented here 
https://github.com/Taraxa-project/taraxa-node/pull/1191

PR #1190, opened 2021-09-28
chore(syncing): fix condition for pushing syncing blocks

MatusKysel:
## Purpose

This should fix latest issues on testnet described by https://github.com/Taraxa-project/taraxa-node/pull/1189


MatusKysel:
@mfrankovi  fix is better

PR #1191, opened 2021-09-28
chore(syncing): fix condition for pushing syncing blocks

MatusKysel:
## Purpose

This should fix latest issues on testnet described by #1189

### DIFF
--- a/src/consensus/pbft_manager.cpp
+++ b/src/consensus/pbft_manager.cpp
@@ -1762,14 +1762,11 @@ bool PbftManager::is_syncing_() {
 }
 
 uint64_t PbftManager::pbftSyncingPeriod() const {
-  return std::max(sync_period_.load(), pbft_chain_->getPbftChainSize());
+  return std::max(sync_queue_.getPeriod(), pbft_chain_->getPbftChainSize());
 }
 
 std::optional<SyncBlock> PbftManager::processSyncBlock() {
-  std::unique_lock lock(sync_queue_access_);
-  auto sync_block = std::move(sync_queue_.front());
-  sync_queue_.pop_front();
-  lock.unlock();
+  auto sync_block = sync_queue_.pop();
   auto pbft_block_hash = sync_block.first.pbft_blk->getBlockHash();
   LOG(log_nf_) << "Pop pbft block " << pbft_block_hash << " from synced queue";
 
@@ -1782,7 +1779,7 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
     LOG(log_er_) << "Invalid PBFT block " << pbft_block_hash
                  << "; prevHash: " << sync_block.first.pbft_blk->getPrevBlockHash() << " from peer "
                  << sync_block.second.abridged() << " received, stop syncing.";
-    clearSyncBlockQueue();
+    sync_queue_.clear();
     // Handle malicious peer on network level
     net->handleMaliciousSyncPeer(sync_block.second);
     return nullopt;
@@ -1793,7 +1790,7 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
     if (vote->getBlockHash() != pbft_block_hash) {
       LOG(log_er_) << "Invalid cert votes block hash " << vote->getBlockHash() << " instead of " << pbft_block_hash
                    << " from peer " << sync_block.second.abridged() << " received, stop syncing.";
-      clearSyncBlockQueue();
+      sync_queue_.clear();
       net->handleMaliciousSyncPeer(sync_block.second);
       return nullopt;
     }
@@ -1814,7 +1811,7 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
     LOG(log_er_) << "Order hash incorrect in sync block " << pbft_block_hash << " expected: " << order_hash
                  << " received " << sync_block.first.pbft_blk->getOrderHash() << "; Dag order: " << blk_order
                  << "; Trx order: " << trx_order << "; from " << sync_block.second.abridged() << ", stop syncing.";
-    clearSyncBlockQueue();
+    sync_queue_.clear();
     net->handleMaliciousSyncPeer(sync_block.second);
     return nullopt;
   }
@@ -1832,7 +1829,7 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
     LOG(log_er_) << "Synced PBFT block " << pbft_block_hash
                  << " doesn't have enough valid cert votes. Clear synced PBFT blocks! DPOS total votes count: "
                  << getDposTotalVotesCount();
-    clearSyncBlockQueue();
+    sync_queue_.clear();
     net->handleMaliciousSyncPeer(sync_block.second);
     return nullopt;
   }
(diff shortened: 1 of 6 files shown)


src/consensus/pbft_manager.cpp
```diff
@@ -1852,7 +1852,7 @@ void PbftManager::syncBlockQueuePush(SyncBlock &&block, dev::p2p::NodeID const &
 }
 
 void PbftManager::clearSyncBlockQueue() {
-  sync_period_ = 0;
+  sync_period_ = pbft_chain_->getPbftChainSize();
```
mfrankovi, line 1855: I believe there can be a race condition between syncBlockQueuePush and clearSyncBlockQueue. Scenario:
Queue already contains block period 7,8 and 9.
At the same time one thread enters syncBlockQueuePush and another clearSyncBlockQueue
First thread successfully skips if (period != pbftSyncingPeriod() + 1) 
After that second thread sets sync_peiod to 0 and clears the queue
First thread then pushes belock period 10 in the queue which is invalid since periods 7,8 and 9 were deleted
Solution would be putting everything in these two methods under the same lock

JakubFornadel, line 1855: This seems unnecessary as `pbft_chain_->getPbftChainSize()` would be returned from `pbftSyncingPeriod()` in `syncBlockQueuePush` anyway...

JakubFornadel: @mfrankovi yop I think you are right.

JakubFornadel (changes requested):
So I went through the code once again and I dont really understand what is this atomoc `sync_period_` good for ? As Miho commented, it introduced race condition and to fix it, it must be under the same unique_lock as sync_queue. In those 1 or 2 other places why not just use sync_queue.back()->getPeriod() instead of sync_period ? I think sync_period does not have any added value and complicates stuff

mfrankovi:
> So I went through the code once again and I dont really understand what is this atomoc `sync_period_` good for ? As Miho commented, it introduced race condition and to fix it, it must be under the same unique_lock as sync_queue. In those 1 or 2 other places why not just use sync_queue.back()->getPeriod() instead of sync_period ? I think sync_period does not have any added value and complicates stuff

The problem is also the time when we pop the block off the queue. If you use sync_queue.back()->getPeriod() info but the block is already popped from the queue but not pushed to pbft chain, you don't have good enough info. I think if we are to remove this sync_period_ and use sync_queue.back()->getPeriod() we need to make sure that the block is popped off the queue only after it has been pushed to pbft chain

src/consensus/sync_queue.cpp
```diff
@@ -0,0 +1,44 @@
...
+  }
+  sync_period_ = period;
+  sync_queue_.emplace_back(std::move(block), node_id);
+  return true;
+}
+
+std::pair<SyncBlock, dev::p2p::NodeID> SyncBlockQueue::pop() {
+  std::unique_lock lock(sync_queue_access_);
+  auto sync_block = std::move(sync_queue_.front());
+  sync_queue_.pop_front();
```
JakubFornadel, line 40: I think you should check here if sync_queue_ is empty and if so, set sync_period_to 0

src/consensus/sync_queue.cpp
```diff
@@ -0,0 +1,44 @@
...
+void SyncBlockQueue::clear() {
+  std::unique_lock lock(sync_queue_access_);
+  sync_period_ = 0;
+  sync_queue_.clear();
+}
+
+bool SyncBlockQueue::push(SyncBlock &&block, dev::p2p::NodeID const &node_id, uint64_t max_pbft_size) {
+  const auto period = block.pbft_blk->getPeriod();
+  std::unique_lock lock(sync_queue_access_);
+  if (period != std::max(sync_period_, max_pbft_size) + 1) {
```
JakubFornadel, line 29: not sure if it ican actually happen but according to this code, it might be possible that  `sync_period_ != 0 (this means sync_queue_ is not empty)` and `max_pbft_size > sync_period_.` In such case it is pointless to push into queue on top of the data that is there because such data are not valid. In case this is true, we could clear sync_queue

src/consensus/sync_queue.hpp
```diff
@@ -0,0 +1,28 @@
...
+
+  bool push(SyncBlock &&block, dev::p2p::NodeID const &node_id, uint64_t max_pbft_size);
+  std::pair<SyncBlock, dev::p2p::NodeID> pop();
+  void clear();
+  size_t size() const;
+  uint64_t getPeriod() const;
+
+ private:
+  std::deque<std::pair<SyncBlock, dev::p2p::NodeID>> sync_queue_;
+  // We need this variable as for small amount of time block is not part of queue but still beeing processed
```
JakubFornadel, line 23: I would also mention that sync_period_ should be always equal to the sync_queue_.back()->getPeriod(). You forgot about it in pop() function for example

src/consensus/sync_queue.cpp
JakubFornadel: hmm or maybe not, we cant do it because we need to keep that value for the time between popping from queue and pushing into the chain right ?

MatusKysel: nope, because of processing of this block can take a while, we can to set it to 0

MatusKysel: That's the reason why we have the sync_period_ there

src/consensus/sync_queue.cpp
MatusKysel: because of previous discussion this needs to be like this

src/consensus/sync_queue.hpp
```diff
@@ -0,0 +1,28 @@
...
+  SyncBlockQueue() = default;
+
+  bool push(SyncBlock &&block, dev::p2p::NodeID const &node_id, uint64_t max_pbft_size);
+  std::pair<SyncBlock, dev::p2p::NodeID> pop();
+  void clear();
+  size_t size() const;
+  uint64_t getPeriod() const;
+
+ private:
+  std::deque<std::pair<SyncBlock, dev::p2p::NodeID>> sync_queue_;
```
JakubFornadel, line 22: no need to call this sync_queue_ when class name is SyncBlockQueue

src/consensus/sync_queue.hpp
```diff
@@ -0,0 +1,28 @@
...
+  // We need this variable as for small amount of time block is not part of queue but still beeing processed
+  uint64_t sync_period_{0};
+  mutable std::shared_mutex sync_queue_access_;
```
JakubFornadel, line 25: same here for the naming

src/consensus/sync_queue.cpp
JakubFornadel: no this is different. If `max_pbft_size > sync_period_` it means sync_queue has outdated data which will be rejected anyway

MatusKysel: nope, let's say you sync up to period 5, then you will receive "normally" blocks up to period 10 and then sync again. `sync_period_` will not be cleared but `max_pbft_size` will be higher. That's what happened on testnet

src/consensus/sync_queue.hpp
JakubFornadel, line 23: We need this variable for a small amount of time when block is not part of queue but still being processed

src/consensus/sync_queue.cpp
JakubFornadel: ok update, this is what I meant: 

```
if (max_pbft_size > sync_period_ && !sync_queue.empty()) {
 sync_queue.clear();
}
```

JakubFornadel (approved):

mfrankovi (approved):

PR #1193, opened 2021-09-28
chore: include threadID into the log format + use short version of No…

JakubFornadel:
## Purpose

- Add threadID into the log format as due to new tarcap design, it can happen for example that multiple packets of the same type are processed at the same time and logs would mix up, which would make debugging very difficult. ThreadID should enable filtering if needed

- use short node id instead of full version

New format:

```
"%ThreadID% %ShortNodeId% %Channel% [%TimeStamp%] %SeverityStr%: %Message%"
0x00007f4652ffd700 de2b1203… TARCAP_TP [2021-09-28 16:10:46.099704] DEBUG: Worker (0) started
```

### DIFF
--- a/src/cli/default_config.hpp
+++ b/src/cli/default_config.hpp
@@ -48,14 +48,14 @@ const char *default_json = R"foo({
         "outputs": [
           {
             "type": "console",
-            "format": "%TimeStamp% %Channel% %SeverityStr% %Message%"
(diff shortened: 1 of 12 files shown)


JakubFornadel:
@rattrap Do we need to adjust also Kibana as log format changed ?

MatusKysel (approved):

mfrankovi (approved):

PR #1202, opened 2021-09-29
Check Synced Cert Votes

gaoqi8649:
Since node only save 2t+1 cert votes for each finalized PBFT block. When node receive a synced PBFT block from peers, if the number of cert votes are not 2t+1, or any cert votes failed checking conditions, the whole cert votes bundle should be invalid and failed immediately.

### DIFF
--- a/libraries/storage/include/storage/storage.hpp
+++ b/libraries/storage/include/storage/storage.hpp
@@ -178,7 +178,7 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
   std::set<blk_hash_t> getBlocksByLevel(level_t level);
   std::vector<std::shared_ptr<DagBlock>> getDagBlocksAtLevel(level_t level, int number_of_levels);
   void updateDagBlockCounters(Batch& write_batch, std::vector<DagBlock> blks);
-  std::map<int, std::vector<DagBlock>> getNonfinalizedDagBlocks();
+  std::map<level_t, std::vector<DagBlock>> getNonfinalizedDagBlocks();
   void removeNonfinalizedDagBlock(blk_hash_t const& hash);
 
   // Transaction
--- a/libraries/storage/src/storage.cpp
+++ b/libraries/storage/src/storage.cpp
@@ -250,8 +250,8 @@ std::vector<std::shared_ptr<DagBlock>> DbStorage::getDagBlocksAtLevel(level_t le
   return res;
 }
 
-std::map<int, std::vector<DagBlock>> DbStorage::getNonfinalizedDagBlocks() {
-  std::map<int, std::vector<DagBlock>> res;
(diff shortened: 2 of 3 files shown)


src/consensus/vote.cpp
```diff
@@ -543,51 +543,48 @@ bool VoteManager::voteValidation(std::shared_ptr<Vote>& vote, size_t dpos_total_
 bool VoteManager::pbftBlockHasEnoughValidCertVotes(SyncBlock& pbft_block_and_votes, size_t dpos_total_votes_count,
                                                    size_t sortition_threshold, size_t pbft_2t_plus_1) const {
   if (pbft_block_and_votes.cert_votes.empty()) {
-    LOG(log_er_) << "No any cert votes! The synced PBFT block comes from a "
-                    "malicious player.";
+    LOG(log_er_) << "No any cert votes! The synced PBFT block comes from a malicious player.";
+    return false;
+  }
+  if (pbft_block_and_votes.cert_votes.size() != pbft_2t_plus_1) {
```
MatusKysel, line 549: so if we have more votes it is not ok?

src/consensus/vote.cpp
```diff
@@ -543,51 +543,48 @@ bool VoteManager::voteValidation(std::shared_ptr<Vote>& vote, size_t dpos_total_
 bool VoteManager::pbftBlockHasEnoughValidCertVotes(SyncBlock& pbft_block_and_votes, size_t dpos_total_votes_count,
                                                    size_t sortition_threshold, size_t pbft_2t_plus_1) const {
   if (pbft_block_and_votes.cert_votes.empty()) {
-    LOG(log_er_) << "No any cert votes! The synced PBFT block comes from a "
-                    "malicious player.";
+    LOG(log_er_) << "No any cert votes! The synced PBFT block comes from a malicious player.";
```
justinsnapp, line 546: Should be "No cert votes provided.  Synced PBFT block comes from a malicious player."

src/consensus/vote.cpp
justinsnapp: @gaoqi8649  the log is inaccurate because you say doesnt have enough cert votes but it could in fact have too many.

@MatusKysel thats a good question.   Sending excessive cert votes while syncing i think is indeed a bad behavior.   Seems not worth it to verify them in hopes we just got too many....

gaoqi8649: @MatusKysel Since for good nodes, they will only save 2t+1 cert votes in DB. So for any good nodes, that's not possible sending out over 2t+1 cert votes.

gaoqi8649: @justinsnapp fixed log error, thanks

src/consensus/vote.cpp
gaoqi8649: Fixed log message, thanks

MatusKysel (approved):

JakubFornadel (approved):

PR #1204, opened 2021-09-30
fix: get dag order with correct period

mfrankovi:
getDagBlockOrder only works correctly if it is called for the current period. New parameter was added to specify for which pbft period we request the order. If the period is not the expected one, error is logged and empty response is returned.

In general we should not be calling getDagBlockOrder with incorrect period. Additional changes in pbft_manager should be done to ensure this

### DIFF
--- a/libraries/core_libs/consensus/src/dag/dag.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag.cpp
@@ -475,19 +475,19 @@ void DagManager::getGhostPath(std::vector<blk_hash_t> &ghost) const {
 }
 
 // return {period, block order}, for pbft-pivot-blk proposing
-std::pair<uint64_t, std::vector<blk_hash_t>> DagManager::getDagBlockOrder(blk_hash_t const &anchor, uint64_t period) {
+std::vector<blk_hash_t> DagManager::getDagBlockOrder(blk_hash_t const &anchor, uint64_t period) {
   SharedLock lock(mutex_);
   std::vector<blk_hash_t> blk_orders;
 
   if (period != period_ + 1) {
     LOG(log_er_) << "getDagBlockOrder called with period " << period << ". Expected period " << period_ + 1
                  << std::endl;
-    return {0, {}};
+    return {};
   }
 
   if (anchor_ == anchor) {
     LOG(log_wr_) << "Query period from " << anchor_ << " to " << anchor << " not ok " << std::endl;
-    return {0, {}};
+    return {};
   }
 
   auto new_period = period_ + 1;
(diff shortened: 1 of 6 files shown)


libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -475,12 +475,16 @@ void DagManager::getGhostPath(std::vector<blk_hash_t> &ghost) const {
 }
 
 // return {period, block order}, for pbft-pivot-blk proposing
-std::pair<uint64_t, std::vector<blk_hash_t>> DagManager::getDagBlockOrder(blk_hash_t const &anchor) {
+std::pair<uint64_t, std::vector<blk_hash_t>> DagManager::getDagBlockOrder(blk_hash_t const &anchor, uint64_t period) {
```
MatusKysel, line 478: I know it is not part of your changes, but do we need to return period? I have checked the code and it seems to me, that is used just for test and never for actual node. Maybe we could remove it and make it simpler

MatusKysel: I mean period in return std::pair to be clear

mfrankovi: Done

MatusKysel (approved):

justinsnapp (approved):

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -475,29 +475,33 @@ void DagManager::getGhostPath(std::vector<blk_hash_t> &ghost) const {
...
 // return {period, block order}, for pbft-pivot-blk proposing
-std::pair<uint64_t, std::vector<blk_hash_t>> DagManager::getDagBlockOrder(blk_hash_t const &anchor) {
+std::vector<blk_hash_t> DagManager::getDagBlockOrder(blk_hash_t const &anchor, uint64_t period) {
   SharedLock lock(mutex_);
-  // TODO: need to check if the anchor already processed
-  // if the period already processed
   std::vector<blk_hash_t> blk_orders;
 
+  if (period != period_ + 1) {
+    LOG(log_er_) << "getDagBlockOrder called with period " << period << ". Expected period " << period_ + 1
```
gaoqi8649, line 483: I think change it to log_wr_?

gaoqi8649: Like in the case happened in Testnet. There are at least have one node get enough cert votes and pushed period 5104 into PBFT chain. Then the node will be able to proposed block period at 5105. But for some other nodes who have not had period 5104, will fail at here. So those nodes will not cert vote for the block period 5105 until they push period 5104 into chain. That's not error and I think that will happen often

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -475,29 +475,33 @@ void DagManager::getGhostPath(std::vector<blk_hash_t> &ghost) const {
 }
 
 // return {period, block order}, for pbft-pivot-blk proposing
```
JakubFornadel, line 477: lets update also comment, which is outdated

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -475,29 +475,33 @@ void DagManager::getGhostPath(std::vector<blk_hash_t> &ghost) const {
...
-std::pair<uint64_t, std::vector<blk_hash_t>> DagManager::getDagBlockOrder(blk_hash_t const &anchor) {
+std::vector<blk_hash_t> DagManager::getDagBlockOrder(blk_hash_t const &anchor, uint64_t period) {
   SharedLock lock(mutex_);
-  // TODO: need to check if the anchor already processed
-  // if the period already processed
   std::vector<blk_hash_t> blk_orders;
 
+  if (period != period_ + 1) {
+    LOG(log_er_) << "getDagBlockOrder called with period " << period << ". Expected period " << period_ + 1
+                 << std::endl;
```
JakubFornadel, line 484: no need to use std::endl; for `LOG()`

JakubFornadel (approved):

PR #1206, opened 2021-09-30
test(threadpool): added test for dag block blocking dependencies

JakubFornadel:
#### 6 unit tests for tarcap threadpool
Tests check:
- proper working of all blocking dependencies
- proper concurrent processing
- threads borrowing
- threads starvation in "lower" priority queues

#### Threadpool fix
Thank to threads starvation test I found bug in priority queue pop() function, which would always return only packets from High priority queue if there were any. Such situation should never happen as that would block processing of packets from other queues - there are max threads limits for each priority queue. This limit was not properly checked

#### Votes packet handler split into 3 separate handler:
- PbftVote packet handler
- VotesSync packet handler
- GetVotesSync packet handler
This enables us to debug it more effectively as those handlers have now their own loggers, functionality is separated etc...

#### SyncingHandler as inheritance instead of compostion
SyncingHandler is used in multiple packet handlers that somehow need to trigger syncing functionality. It was used as their class member. Now it is their base class thanks to which we will see now which handler trigger syncing + some redundant class members could be deleted.

# For reviewers
There are no functional/logic/flow changes so no need to concentrate on that. The tests might be worth looking at if anyone interested, otherwise it is just naming/design refactor

[This PR](https://github.com/Taraxa-project/taraxa-testnet/pull/70) in taraxa-tesnet should be merged after this PR is merged.

### DIFF
--- a/libraries/core_libs/network/include/network/tarcap/packet_types.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packet_types.hpp
@@ -12,7 +12,7 @@ namespace taraxa::network::tarcap {
 enum SubprotocolPacketType : uint32_t {
   // Consensus packets with high processing priority
   HighPriorityPackets = 0,
-  PbftVotePacket,
+  VotePacket,
   GetVotesSyncPacket,
   VotesSyncPacket,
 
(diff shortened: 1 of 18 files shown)


libraries/core_libs/network/include/network/tarcap/packet_types.hpp
```diff
@@ -56,10 +56,10 @@ inline std::string convertPacketTypeToString(SubprotocolPacketType packet_type)
       return "TestPacket";
     case PbftVotePacket:
```
MatusKysel, line 57: maybe we should also rename this to just `Votepacket`

PR #1211, opened 2021-10-04
temp logs for missing trx debug

MatusKysel:
## Version Notes ##
{ major | minor | patch }


libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
```diff
@@ -28,6 +28,10 @@ inline void TransactionPacketHandler::process(const PacketData &packet_data, con
   for (size_t tx_idx = 0; tx_idx < transaction_count; tx_idx++) {
     const auto &transaction = transactions.emplace_back(packet_data.rlp_[tx_idx].data().toBytes());
 
+    LOG(log_er_) << "Received Transaction " << transaction.getHash().abridged() << " peer id "
```
JakubFornadel, line 31: this is ideal, move this log after for loop, inside the loop just collect hashes of txs. If you receive 50 txs you will print 50 times packet received time, node id, etc... which is useless

PR #1212, opened 2021-10-04
fix: missing transaction

mfrankovi:
If block is missing a tip or pivot the transactions included in the block are now saved and not dropped

### DIFF
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
@@ -39,12 +39,6 @@ void DagBlockPacketHandler::process(const PacketData &packet_data, const std::sh
     new_transactions.push_back(std::move(transaction));
   }
 
-  // Ignore new block packets when pbft syncing
-  if (syncing_state_->is_pbft_syncing()) {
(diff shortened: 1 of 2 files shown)


JakubFornadel (approved):

kstdl (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
```diff
@@ -53,6 +53,7 @@ void DagBlockPacketHandler::process(const PacketData &packet_data, const std::sh
     }
 
     if (auto status = syncing_handler_->checkDagBlockValidation(block); !status.first) {
+      trx_mgr_->insertBroadcastedTransactions(new_transactions);
```
MatusKysel, line 56: but then we do not need to send all trxs on `MissingHashes` in `GetDagSyncPacketsHandler` and also we should think about malicious peer and somehow later remove those trxs

JakubFornadel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
JakubFornadel: @MatusKysel  Lets put this into the issue

MatusKysel (approved):

PR #1213, opened 2021-10-04
build(conan): fix conan cache invalidation caused by different build type detection

kstdl:
## Purpose

Dependencies from the conan was rebuilt on every build because of different build type detection by conan-cmake. Fixed by setting build type to the conan profile

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -454,22 +454,26 @@ jobs:
             echo ${GCLOUD_SERVICE_KEY} | docker login -u _json_key --password-stdin https://gcr.io
 
       - run:
-         name: Build docker image
+         name: Build docker builder image
          command: |
-                docker build -t ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} .
+                docker build -t ${IMAGE}-builder:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} --target builder .
 
       - run:
          name: Build ctest image
          command: |
-               docker build -t ${IMAGE}-ctest:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} \
-                 --target build .              
-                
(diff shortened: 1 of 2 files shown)


agrebin (approved):

.circleci/config.yml
```diff
@@ -416,6 +416,16 @@ jobs:
...
       - checkout
+      - run:
+         name: Login into gcr
+         command: |
+            echo ${GCLOUD_SERVICE_KEY} | docker login -u _json_key --password-stdin https://gcr.io
+
+      - run:
+         name: Build docker builder image
+         command: |
+                docker build -t test --target builder .
```
sfrique, line 427: Are we sure we want this to run before the clean? 
To be honest, this cache should be automatically used, I don't see why we need to explicitly build this target.

Also, this uses a bad tag `test`

kstdl: As I saw docker removes build intermediate containers. So I thought builder image was just removed during full build. 
I will move it a little bit down after clean and will fix the tag. I changed it just for local test but forgot to change it back

.circleci/config.yml
```diff
@@ -427,6 +427,17 @@ jobs:
...
+
+      - run:
+         name: Login into gcr
+         command: |
+            echo ${GCLOUD_SERVICE_KEY} | docker login -u _json_key --password-stdin https://gcr.io
+
+      - run:
+         name: Build docker builder image
+         command: |
+                docker build -t ${IMAGE}:${DOCKER_BRANCH_TAG}-${CIRCLE_BUILD_NUM} --target builder .
```
sfrique, line 439: Have to check if it works, but the `DOCKER_BRANCH_TAG`  is only generated a step later, right?

kstdl: Missed that. Thought it is a default circleci variable 
And it isnt working, because build failed

JakubFornadel (approved):

PR #1217, opened 2021-10-05
fix: missing transaction

mfrankovi:
Even if block is missing a tip or pivot, save the transactions included in the block

### DIFF
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
@@ -47,6 +47,8 @@ void DagBlockPacketHandler::process(const PacketData &packet_data, const std::sh
     }
 
     if (auto status = syncing_handler_->checkDagBlockValidation(block); !status.first) {
+      trx_mgr_->insertBroadcastedTransactions(new_transactions);
+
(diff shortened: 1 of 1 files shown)


libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
```diff
@@ -47,6 +47,8 @@ void DagBlockPacketHandler::process(const PacketData &packet_data, const std::sh
     }
 
     if (auto status = syncing_handler_->checkDagBlockValidation(block); !status.first) {
+      trx_mgr_->insertBroadcastedTransactions(new_transactions);
```
MatusKysel, line 50: maybe this should after `if (syncing_state_->is_pbft_syncing())` as we are still syncing we do not need those trxs

mfrankovi: I was also having doubts about that. If nodes are behaving honestly they should not be gossiping new blocks. This actually happens only when the node which sent this block thinks we are synced and this only happens when we are behind only by one block or so. So I think it is better to save transaction in this case

MatusKysel (approved):

kstdl (approved):

PR #1223, opened 2021-10-06
fix: missing transaction

mfrankovi:
Transactions sent with the new dag block were inserted with a delay since they were in the dag blocks unverified queue. This could cause missing transaction error for another block that arrives and needs the transaction. The fix now inserts transactions in the trx manager as soon as we process the packet

### DIFF
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
@@ -28,16 +28,20 @@ void DagBlockPacketHandler::process(const PacketData &packet_data, const std::sh
   DagBlock block(packet_data.rlp_[0].data().toBytes());
   blk_hash_t const hash = block.getHash();
   const auto transactions_count = packet_data.rlp_.itemCount() - 1;
-  LOG(log_dg_) << "Received DagBlockPacket " << hash.abridged() << " with " << transactions_count << " txs";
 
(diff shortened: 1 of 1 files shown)


MatusKysel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
```diff
@@ -46,9 +50,8 @@ void DagBlockPacketHandler::process(const PacketData &packet_data, const std::sh
       return;
     }
 
+    trx_mgr_->insertBroadcastedTransactions(new_transactions);
     if (auto status = syncing_handler_->checkDagBlockValidation(block); !status.first) {
-      trx_mgr_->insertBroadcastedTransactions(new_transactions);
```
JakubFornadel, line 50: So it might be little bit too "hot" hotfix :) I checked and with this approach we would process txs twice or at least try to process it twice. I would keep insertBroadcastedTransactions here and also put it maybe inside onNewBlockReceived -> insertBroadcastedBlockWithTransactions. Then we dont need and probably should not to save txs into the dag unverified queue.

JakubFornadel: another point is that I think we might get into this error also inside dag_sync_packet_handler -> process() function where we are doing basically the same stuff as here but we dont insert txs in case of failed dag block validation. Not sure if that is valid scenario though...

Issue #1216, opened 2021-10-05
Controlling the rate of DAG block production to maintain a desired forking rate / DAG efficiency


justinsnapp:
## Is your feature request related to a problem? Please describe.

As more nodes get delegate to we have an increasing DAG block production rate.   In general the number of nodees that can be delegate to is variable, starting with a small number of genesis and increasing to perhaps some cap.  Particularly in the testnet, there is no real cap on the number of delegated to nodes at present.

The two competing goals are to have (1) rapid block production (as many proposers as possible able to generate DAG blocks as quickly as possible) WHILE (2) maintaining a finite amount of forking and maximizing DAG efficiency.

At high DAG block generation rate we see the same transactions packed into multiple blocks and lots of blocks at the same level.

The relationship between DAG generation parameters and block efficiency is not easily calculated as it can depend on the connectivity and network diameter of the proposing nodes.   In the limit of small network diameter and good coordination then high block generation rate with high DAG efficiency is of course possible. (At the cost of centralization.)

## Describe the solution you'd like.

We have lots of "knobs" we can adjust to use feedback to achieve a balance.  This is illustrated in the graph showing the relevant DAG proposal parameters for VRF and VDF based proposal...

```
        VDF
          |
          |
          |
          |
          |                       ________________________
vdf stale T  `  `   `  `  `  `  ` |
          |                       |
          |                       |
          |                       |
          |                       |
          |                       |
          |                       |
          |                       |
vdf max   T   `  `   `  `   `  ` /.
          |                     /
          |                    /  .
          |                   /
          |                  /    . 
          |                 /      
          |                /      .
          |               /        
          |              /        .
          |             /
          |            /          .
vdf min   T  `  `  `  |
          |           |           .
          |           |
          |           |           .
          |           |
          |           |           .
          |           |
          |           |           .
          |           |
           -----------+-----------+-------------------------> VRF Value
                   vrf_omit     vrf_stale

```

### Two feedback approaches

There are multiple ways to determine an error signal for the feedback loop.    Two likely approaches are to:

1. Define a target DAG efficiency.   Efficiency for any given PBFT block being defined as sum (unique transactions) over all DAG blocks in a PBFT block divided by the sum (transactions in a DAG block) over all DAG blocks in a PBFT block. 

2. Define a target forking rate.   A proxy for forking rate can be defined as total number of DAG blocks / length of the anchor chain in a PBFT block.

[...]

PR #1233, opened 2021-10-18
Adjust vrf params

kstdl:
## Purpose

Controlling the rate of DAG block production to maintain a desired DAG efficiency. It is done by VRF params manipulation which adjusts DAG block producers count.

## Related Github issues

https://github.com/Taraxa-project/taraxa-node/issues/1216

## Tests

Implementing the test suite that will check the whole adjusting algorithm

### DIFF
--- a/libraries/aleth/libdevcore/CommonJS.h
+++ b/libraries/aleth/libdevcore/CommonJS.h
@@ -3,6 +3,8 @@
 // Licensed under the GNU General Public License, Version 3.
 #pragma once
 
+#include <json/json.h>
+
 #include <string>
 
 #include "CommonData.h"
@@ -91,5 +93,13 @@ inline u256 jsToU256(std::string const& _s) { return jsToInt<32>(_s); }
 inline uint64_t jsToInt(std::string const& _s) { return std::stoull(_s, 0, 16); }
 
 inline std::string jsToDecimal(std::string const& _s) { return toString(jsToU256(_s)); }
+
+inline uint64_t getUInt(const Json::Value& v) {
+  if (v.isString()) {
+    return dev::jsToInt(v.asString());
+  }
+
+  return v.asUInt64();
+}
 
 }  // namespace dev
--- a/libraries/cli/include/cli/devnet_config.hpp
+++ b/libraries/cli/include/cli/devnet_config.hpp
@@ -142,7 +142,7 @@ const char *devnet_json = R"foo({
     "sortition": {
       "changes_count_for_average": "0x5",
       "max_interval_correction": "0x3E8",
-      "target_dag_efficiency": "0x1388",
+      "dag_efficiency_targets": ["0x12C0", "0x1450"],
       "computation_interval": "0x32",
       "vrf": {
         "threshold_upper": "0xbffd",
--- a/libraries/cli/include/cli/devnet_config.hpp
+++ b/libraries/cli/include/cli/devnet_config.hpp
@@ -139,13 +139,17 @@ const char *devnet_json = R"foo({
     "replay_protection_service": {
       "range": "0xa"
     },
-    "vdf": {
-      "difficulty_max": "0x12",
-      "difficulty_min": "0x10",
-      "threshold_selection": "0xbffd",
-      "threshold_vdf_omit": "0x6bf7",
-      "difficulty_stale": "0x13",
-      "lambda_bound": "0x64"
+    "sortition": {
+      "vrf": {
+        "threshold_upper": "0xbffd",
+        "threshold_lower": "0x6bf7"
+      },
+      "vdf": {
+        "difficulty_max": "0x12",
+        "difficulty_min": "0x10",
+        "difficulty_stale": "0x13",
+        "lambda_bound": "0x64"
+      }
     }
   }
 }
--- a/libraries/cli/include/cli/devnet_config.hpp
+++ b/libraries/cli/include/cli/devnet_config.hpp
@@ -140,6 +140,10 @@ const char *devnet_json = R"foo({
       "range": "0xa"
     },
     "sortition": {
+      "changes_count_for_average": "0x5",
+      "max_interval_correction": "0x3E8",
+      "target_dag_efficiency": "0x1388",
+      "computation_interval": "0x32",
       "vrf": {
         "threshold_upper": "0xbffd",
         "threshold_lower": "0x6bf7"
--- a/libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
+++ b/libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
@@ -13,10 +13,10 @@ struct SortitionParamsChange {
   uint16_t actual_correction_per_percent = 1;
 
   SortitionParamsChange() = default;
-  SortitionParamsChange(uint64_t period, uint16_t efficiency, VrfParams const& vrf);
-  SortitionParamsChange(uint64_t period, uint16_t efficiency, VrfParams const& vrf,
-                        SortitionParamsChange const& previous);
-  static SortitionParamsChange from_rlp(dev::RLP const& rlp);
+  SortitionParamsChange(uint64_t period, uint16_t efficiency, const VrfParams& vrf);
+  SortitionParamsChange(uint64_t period, uint16_t efficiency, const VrfParams& vrf,
+                        const SortitionParamsChange& previous);
+  static SortitionParamsChange from_rlp(const dev::RLP& rlp);
   bytes rlp() const;
 
   /**
@@ -25,26 +25,22 @@ struct SortitionParamsChange {
    * Current efficiency is 72%. Based on past calculations, we changed the parameter to 100.
    * The interval efficiency after these changes was 74%. It turns out that it has changed by 2%, but not the 3% as we
    * need Therefore, the required changePerPercent is equal `100 / 2 = 50`, and not `100 / 3 = 33.(3)` as it was before
+   *
+   * @param previous previous params change data
+   * @returns params correction per percent
    */
-  uint16_t correctionPerPercent(SortitionParamsChange const& previous) const;
+  uint16_t correctionPerPercent(const SortitionParamsChange& previous) const;
 };
 
 class SortitionParamsManager {
-  SortitionConfig config_;
-  std::shared_ptr<DbStorage> db_;
-  std::vector<uint16_t> dag_efficiencies_;
-  std::deque<SortitionParamsChange> params_changes_;
-  std::optional<SortitionParamsChange> calculateChange(const uint64_t period);
-  int32_t getChange(const uint64_t period, const uint16_t efficiency) const;
-  void cleanup();
-
  public:
-  SortitionParamsManager(SortitionConfig sort_conf, std::shared_ptr<DbStorage> db);
+  SortitionParamsManager(const addr_t& node_addr, SortitionConfig sort_conf, std::shared_ptr<DbStorage> db);
   SortitionParams getSortitionParams(std::optional<uint64_t> for_period = {}) const;
   uint64_t currentProposalPeriod() const;
   /**
    * Calculates the average correction from previous changes we store in memory.
    * We can regulate corrections count by removing oldest elements of  params_changes_ collection
+   * @returns average of corrections per percent for the latest `config_.changes_count_for_average` elements
    */
   uint16_t averageCorrectionPerPercent() const;
 
@@ -52,6 +48,8 @@ class SortitionParamsManager {
    * Calculating DAG efficiency in passed SyncBlock(PBFT)
    * DAG efficiency is count of unique transactions over all DAG blocks in a PBFT block divided by the count over all
    * DAG blocks in a PBFT block.
+   * @param block SyncBlock to calculate its efficiency
+   * @returns DAG efficiency
    */
   uint16_t calculateDagEfficiency(const SyncBlock& block) const;
 
@@ -59,13 +57,25 @@ class SortitionParamsManager {
    * Called from PbftManager.pushPbftBlock_()
    * Calculates and stores needed data on every block.
    * Every `computation_interval` blocks from config launches `recalculate` based on previous data
+   * @param block SyncBlock to process
+   * @param batch DB batch in which all changes  will be added
    */
   void pbftBlockPushed(const SyncBlock& block, DbStorage::Batch& batch);
 
   /**
-   * Calculate average DAG efficiency from dag_efficiencies_
+   * Calculate average DAG efficiency from dag_efficiencies_. Used at the end of interval.
+   * @returns average DAG efficiency in current interval
    */
   uint16_t averageDagEfficiency();
+
+ private:
+  SortitionConfig config_;
+  std::shared_ptr<DbStorage> db_;
+  std::vector<uint16_t> dag_efficiencies_;
+  std::deque<SortitionParamsChange> params_changes_;
+  std::optional<SortitionParamsChange> calculateChange(uint64_t period);
+  int32_t getChange(uint64_t period, uint16_t efficiency) const;
+  void cleanup(uint64_t current_period);
 
   LOG_OBJECTS_DEFINE
 };
--- a/libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
+++ b/libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
@@ -1,0 +1,70 @@
+#pragma once
+
+#include "pbft/sync_block.hpp"
+#include "storage/storage.hpp"
+#include "vdf/config.hpp"
+
+namespace taraxa {
+
+struct SortitionParamsChange {
+  VrfParams vrf_params_;
+  uint16_t interval_efficiency_ = 0;
+  uint16_t actual_correction_per_percent_ = 1;
+
+  SortitionParamsChange() = default;
+  SortitionParamsChange(uint16_t efficiency, VrfParams const& vrf);
+  SortitionParamsChange(uint16_t efficiency, VrfParams const& vrf, SortitionParamsChange const& previous);
+  static SortitionParamsChange from_rlp(dev::RLP const& rlp);
+  bytes rlp() const;
+
+  /**
+   * Calculates needed correction of VRF params per DAG efficiency percent
+   * This calculates parameter change which is necessary to change the efficiency by 1 percent. For example:
+   * Current efficiency is 72%. Based on past calculations, we changed the parameter to 100.
+   * The interval efficiency after these changes was 74%. It turns out that it has changed by 2%, but not the 3% as we
+   * need Therefore, the required changePerPercent is equal `100 / 2 = 50`, and not `100 / 3 = 33.(3)` as it was before
+   */
+  uint16_t correctionPerPercent(SortitionParamsChange const& previous) const;
+};
+
+class SortitionParamsManager {
+  SortitionConfig config_;
+  std::shared_ptr<DbStorage> db_;
+  std::vector<uint16_t> dag_efficiencies_;
+  std::deque<SortitionParamsChange> params_changes_;
+  void recalculate(const uint64_t period);
+  int32_t getChange(const uint64_t period, const uint16_t efficiency) const;
+
+ public:
+  SortitionParamsManager(SortitionConfig sort_conf, std::shared_ptr<DbStorage> db);
+  const SortitionParams& getSortitionParams() const { return config_; }
+
+  /**
+   * Calculates the average correction from previous changes we store in memory.
+   * We can regulate corrections count by removing oldest elements of  params_changes_ collection
+   */
+  uint16_t averageCorrectionPerPercent() const;
+
+  /**
+   * Calculating DAG efficiency in passed SyncBlock(PBFT)
+   * DAG efficiency is count of unique transactions over all DAG blocks in a PBFT block divided by the count over all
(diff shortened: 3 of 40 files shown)


libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
```diff
@@ -0,0 +1,73 @@
...
+   * This calculates parameter change which is necessary to change the efficiency by 1 percent. For example:
+   * Current efficiency is 72%. Based on past calculations, we changed the parameter to 100.
+   * The interval efficiency after these changes was 74%. It turns out that it has changed by 2%, but not the 3% as we
+   * need Therefore, the required changePerPercent is equal `100 / 2 = 50`, and not `100 / 3 = 33.(3)` as it was before
+   */
+  uint16_t correctionPerPercent(SortitionParamsChange const& previous) const;
+};
+
+class SortitionParamsManager {
+  SortitionConfig config_;
```
JakubFornadel, line 33: we are trying to follow google c++ style. Class should start with public section, followed by protected an private:
https://google.github.io/styleguide/cppguide.html#Declaration_Order

libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
```diff
@@ -0,0 +1,73 @@
...
+  std::optional<SortitionParamsChange> calculateChange(const uint64_t period);
+  int32_t getChange(const uint64_t period, const uint16_t efficiency) const;
+  void cleanup();
+
+ public:
+  SortitionParamsManager(addr_t node_addr, SortitionConfig sort_conf, std::shared_ptr<DbStorage> db);
+  SortitionParams getSortitionParams(std::optional<uint64_t> for_period = {}) const;
+  uint64_t currentProposalPeriod() const;
+  /**
+   * Calculates the average correction from previous changes we store in memory.
```
JakubFornadel, line 46: We did not really have any discussion about it on meeting but I think we all could start using doxygen for comments and maybe some day even have some basic auto-generated doc

libraries/core_libs/consensus/include/pbft/block_proposer.hpp
```diff
@@ -41,21 +41,21 @@ class ProposeModelFace {
 
 class SortitionPropose : public ProposeModelFace {
  public:
-  SortitionPropose(VdfConfig const& vdf_config, addr_t node_addr, std::shared_ptr<DagManager> dag_mgr,
+  SortitionPropose(addr_t node_addr, std::shared_ptr<DagManager> dag_mgr, std::shared_ptr<DagBlockManager> dag_blk_mgr,
                    std::shared_ptr<TransactionManager> trx_mgr)
-      : vdf_config_(vdf_config), dag_mgr_(dag_mgr), trx_mgr_(trx_mgr) {
+      : dag_mgr_(dag_mgr), dag_blk_mgr_(std::move(dag_blk_mgr)), trx_mgr_(trx_mgr) {
```
JakubFornadel, line 46: std::move(dag_mgr) & std::move(trx_mgr)

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -0,0 +1,176 @@
...
+  p.period = rlp[2].toInt<uint64_t>();
+  p.interval_efficiency = rlp[3].toInt<uint16_t>();
+  p.actual_correction_per_percent = rlp[4].toInt<uint16_t>();
+
+  return p;
+}
+
+SortitionParamsManager::SortitionParamsManager(addr_t node_addr, SortitionConfig sort_conf,
+                                               std::shared_ptr<DbStorage> db)
+    : config_(sort_conf), db_(db) {
```
JakubFornadel, line 60: std::move(db)

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -0,0 +1,176 @@
...
+  // restore VRF params from last change
+  if (!params_changes_.empty()) {
+    config_.vrf = params_changes_.back().vrf_params;
+  }
+
+  dag_efficiencies_.reserve(config_.computation_interval);
+  dag_efficiencies_ = db_->getLastIntervalEfficiencies(config_.computation_interval);
+}
+uint64_t SortitionParamsManager::currentProposalPeriod() const {
+  if (params_changes_.empty()) return 0;
```
JakubFornadel, line 73: I guess we should decide if we want to use parentheses for 1-line if's or not

JakubFornadel:
Overall nice PR with good code structure and proper comments. Well done :)

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
kstdl: I am ok with this parentheses. But I think for this kind of simple condition checks(just one condition inside and simple return or value setting) we could have it in one line because it is pretty readable. 
But from the other side this rule seems to be very complicated

kstdl: But I think we should just use rules from the styleguide about it and do not create a new exceptions.
Because if we will introduce some exceptions to it - it will be a new style and we should kind support it like create a doc with a description and ideally change clang-format rules and I dont think it is worth it, etc

JakubFornadel: https://google.github.io/styleguide/cppguide.html#Conditionals

- For historical reasons, we allow one exception to the above rules: if an if statement has no else or else if clauses, then the curly braces for the controlled statement or the line breaks inside the curly braces may be omitted

Well official google style guide allows not to use braces for historical reasons... Seems to me that there was a lot of legacy code with such `if's` so they created this exception. Anyway "always use braces" is just bulletproof rule so I personally would vote for it :)

libraries/cli/include/cli/devnet_config.hpp
```diff
@@ -59,7 +59,7 @@ const char *devnet_json = R"foo({
         "verbosity": "ERROR",
         "channels": [
           {
-            "name": "SUMMARY",
+            "name": "DEBUG",
```
gaoqi8649, line 62: mistake here, change name to DEBUG

JakubFornadel (approved):

mfrankovi (approved):

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -0,0 +1,176 @@
...
+  for (auto prev = params_changes_.rbegin(), it = params_changes_.rbegin(); it != params_changes_.rend(); prev = it++) {
+    auto upper_bound = prev->period;
+    if (it->period == upper_bound) {
+      upper_bound = it->period + config_.computation_interval;
+    }
+    if (period >= it->period && period < upper_bound) {
+      p.vrf = it->vrf_params;
+    }
+  }
+  return std::move(p);
```
MatusKysel, line 92: no need to move on return

tests/sortition_test.cpp
```diff
@@ -0,0 +1,399 @@
...
+  for (size_t i = 0; i < count; ++i) {
+    res.push_back(dev::h256(i));
+  }
+  return res;
+}
+
+SyncBlock createBlock(uint64_t period, uint16_t efficiency, size_t dag_blocks_count = 5) {
+  // produced transactions count should be equal to or multiple of this value to produce block with accurate enough
+  // efficiency
+  const size_t TRX_COUNT = 100 * ONE_PERCENT;
```
MatusKysel, line 23: I think for both `TRX_COUNT` and `ONE_PERCENT` we should use google-style -> `kTrxCount` and `kOnePercent`

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -0,0 +1,176 @@
...
+  p.period = rlp[2].toInt<uint64_t>();
+  p.interval_efficiency = rlp[3].toInt<uint16_t>();
+  p.actual_correction_per_percent = rlp[4].toInt<uint16_t>();
+
+  return p;
+}
+
+SortitionParamsManager::SortitionParamsManager(const addr_t& node_addr, SortitionConfig sort_conf,
+                                               std::shared_ptr<DbStorage> db)
+    : config_(sort_conf), db_(std::move(db)) {
```
MatusKysel, line 60: config_(std::move(sort_conf))

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -375,6 +376,63 @@ void DbStorage::saveDagBlock(DagBlock const& blk, Batch* write_batch_p) {
...
+  insert(batch, Columns::pbft_block_dag_efficiency, toSlice(period), toSlice(efficiency));
+}
+
+std::vector<uint16_t> DbStorage::getLastIntervalEfficiencies(uint16_t computation_interval) {
+  std::vector<uint16_t> efficiencies;
+
+  auto it =
+      std::unique_ptr<rocksdb::Iterator>(db_->NewIterator(read_options_, handle(Columns::pbft_block_dag_efficiency)));
+  it->SeekToLast();
+  const auto last_period = FromSlice<uint64_t>(it->key().data());
```
MatusKysel, line 390: from my experience ordering is not guaranteed by rocksDB, so this it not good.

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -375,6 +376,63 @@ void DbStorage::saveDagBlock(DagBlock const& blk, Batch* write_batch_p) {
...
+    // order doesn't matter
+    efficiencies.push_back(FromSlice<uint16_t>(it->value()));
+    it->Prev();
+  }
+
+  return efficiencies;
+}
+
+void DbStorage::cleanupDagEfficiencies(DbStorage::Batch& batch) {
+  forEach(Columns::pbft_block_dag_efficiency,
```
MatusKysel, line 404: check `DeleteRange`

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -375,6 +376,63 @@ void DbStorage::saveDagBlock(DagBlock const& blk, Batch* write_batch_p) {
...
+void DbStorage::saveSortitionParamsChange(uint64_t period, SortitionParamsChange params, Batch& batch) {
+  insert(batch, Columns::sortition_params_change, toSlice(period), toSlice(params.rlp()));
+}
+
+std::deque<SortitionParamsChange> DbStorage::getLastSortitionParams(size_t count) {
+  std::deque<SortitionParamsChange> changes;
+
+  auto it =
+      std::unique_ptr<rocksdb::Iterator>(db_->NewIterator(read_options_, handle(Columns::sortition_params_change)));
+  for (it->SeekToLast(); it->Valid() && changes.size() < count; it->Prev()) {
```
MatusKysel, line 418: same as before, I think ordering is not guaranteed

libraries/types/dag_block/src/dag_block.cpp
```diff
@@ -36,25 +36,22 @@ DagBlock::DagBlock(string const &json)
...
       }()) {}
 
-DagBlock::DagBlock(Json::Value const &doc) {
-  if (auto const &v = doc["level"]; v.isString()) {
-    level_ = dev::jsToInt(v.asString());
-  } else {
-    // this was inconsistent with getJson()
-    // fixme: eliminate this branch
-    level_ = v.asUInt64();
+uint64_t getUInt(auto const &v) {
```
MatusKysel, line 39: let's move this to `CommonJS.h` or maybe there is better file for that

libraries/core_libs/storage/src/storage.cpp
kstdl: https://github.com/facebook/rocksdb/wiki/RocksDB-Overview#3-high-level-architecture
So if I am understanding it correctly that means that we have values in column sorted by key

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
kstdl: I forgot why I did this, but during the build I got an error. So it is needed I guess 
```
taraxa-node/libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp:92:10: error: local variable 'p' will be copied despite being returned by name [-Werror,-Wreturn-std-move]
  return p;
         ^
taraxa-node/libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp:92:10: note: call 'std::move' explicitly to avoid copying
  return p;
         ^
         std::move(p)
```

MatusKysel: hmm, I think you've got this warning because p is reference.If it is so, then there is a bug as you are changing config_ basically

MatusKysel: can you try to replace `auto` for `SortitionParams`

libraries/aleth/libdevcore/CommonJS.h
```diff
@@ -92,4 +92,12 @@ inline uint64_t jsToInt(std::string const& _s) { return std::stoull(_s, 0, 16);
 
 inline std::string jsToDecimal(std::string const& _s) { return toString(jsToU256(_s)); }
 
+inline uint64_t getUInt(auto const& v) {
```
JakubFornadel, line 95: why not simply const std::string& if you dont call with with any other argument type ?

libraries/types/dag_block/src/dag_block.cpp
```diff
@@ -37,24 +37,13 @@ DagBlock::DagBlock(string const &json)
...
 
 DagBlock::DagBlock(Json::Value const &doc) {
-  if (auto const &v = doc["level"]; v.isString()) {
-    level_ = dev::jsToInt(v.asString());
-  } else {
-    // this was inconsistent with getJson()
-    // fixme: eliminate this branch
-    level_ = v.asUInt64();
-  }
+  level_ = dev::getUInt(doc["level"]);
```
JakubFornadel, line 40: not sure where is this used but I am not a big fan of this silent return of some default value as we have no chance to notice that. Shouldnt all these fields be mandatory ?

libraries/types/dag_block/src/dag_block.cpp
```diff
@@ -37,24 +37,13 @@ DagBlock::DagBlock(string const &json)
...
   sig_ = sig_t(doc["sig"].asString());
   pivot_ = blk_hash_t(doc["pivot"].asString());
-  if (auto const &v = doc["timestamp"]; v.isString()) {
-    timestamp_ = dev::jsToInt(v.asString());
-  } else {
-    // this was inconsistent with getJson()
-    // fixme: eliminate this branch
-    timestamp_ = v.asUInt64();
-  }
+  timestamp_ = dev::getUInt(doc["timestamp"]);
```
JakubFornadel, line 46: same here

libraries/vdf/include/vdf/config.hpp
```diff
@@ -2,49 +2,68 @@
...
   uint16_t lambda_bound = 1500;  // lambda upper bound, should be constant
 };
-Json::Value enc_json(VdfConfig const& obj);
-void dec_json(Json::Value const& json, VdfConfig& obj);
+
+struct SortitionParams {
+  SortitionParams() = default;
+  SortitionParams(SortitionParams&&) = default;
+  SortitionParams(SortitionParams const& config) = default;
+  SortitionParams(uint16_t const selection, uint16_t const threshold_lower, uint16_t const min, uint16_t const max,
```
JakubFornadel, line 27: it does not make sense to have const qualifier before uint16_t

libraries/vdf/include/vdf/config.hpp
```diff
@@ -2,49 +2,68 @@
...
   uint16_t difficulty_stale = 0;
   uint16_t lambda_bound = 1500;  // lambda upper bound, should be constant
 };
-Json::Value enc_json(VdfConfig const& obj);
-void dec_json(Json::Value const& json, VdfConfig& obj);
+
+struct SortitionParams {
+  SortitionParams() = default;
+  SortitionParams(SortitionParams&&) = default;
+  SortitionParams(SortitionParams const& config) = default;
```
JakubFornadel, line 26: const Type&

libraries/vdf/include/vdf/config.hpp
```diff
@@ -2,49 +2,68 @@
...
+  SortitionParams(uint16_t const selection, uint16_t const threshold_lower, uint16_t const min, uint16_t const max,
+                  uint16_t const stale, uint16_t const lambda_max_bound)
+      : vrf{selection, threshold_lower}, vdf{min, max, stale, lambda_max_bound} {}
+  SortitionParams(VrfParams vrf, VdfParams vdf) : vrf{vrf}, vdf{vdf} {}
```
JakubFornadel, line 30: const VrfParams&, const VdfParams& ?

libraries/aleth/libdevcore/CommonJS.h
kstdl: this function taking `Json::Value` argument, not `std::string`

kstdl: Specifying it explicitly is better, but we will need to link jsoncpp library to libdevcore.

libraries/types/dag_block/src/dag_block.cpp
kstdl: What do you mean by `silent return of some default value`? Where do you can see it?

JakubFornadel: when there is no "level" in doc

libraries/aleth/libdevcore/CommonJS.h
JakubFornadel: if those fields that you are parsing by this are mandatory (which probably should be), you would not even need this function - problem solved

libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
```diff
@@ -0,0 +1,83 @@
...
+namespace taraxa {
+
+struct SortitionParamsChange {
+  uint64_t period = 0;
+  VrfParams vrf_params;
+  uint16_t interval_efficiency = 0;
+  uint16_t actual_correction_per_percent = 1;
+
+  SortitionParamsChange() = default;
+  SortitionParamsChange(uint64_t period, uint16_t efficiency, VrfParams const& vrf);
```
JakubFornadel, line 16: please refactor new classes and refactored functions according to const Type& styling

libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
```diff
@@ -0,0 +1,83 @@
...
+   */
+  uint16_t averageDagEfficiency();
+
+ private:
+  SortitionConfig config_;
+  std::shared_ptr<DbStorage> db_;
+  std::vector<uint16_t> dag_efficiencies_;
+  std::deque<SortitionParamsChange> params_changes_;
+  std::optional<SortitionParamsChange> calculateChange(const uint64_t period);
+  int32_t getChange(const uint64_t period, const uint16_t efficiency) const;
```
JakubFornadel, line 77: const uint64_t...

libraries/aleth/libdevcore/CommonJS.h
kstdl: Maybe you didnt get what this function do. It is not about mandatory field or not. It is about the field type. If we passing int as string (for example `"52"`, but not just `52`), we should parse it different ways 
This code was in several places, so I just moved it to the function

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
JakubFornadel: agree that std::move on return here is not ok

kstdl: Yeah it was a reference. But I didnt understand now why. I probably need to brush up on how the type inference for auto works

libraries/types/dag_block/src/dag_block.cpp
kstdl: If we want to change the logic here and  check if we have all fields in json then we should create a separate task for it and change it in all places

MatusKysel (approved):

libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
kstdl: It is also should be a separate PR or at least a commit if you want it for some reason it to be done in this PR. It is easier(or just not very harder) to change it in whole project then separate just my changes

kstdl: I did it just for new files and in some places in refactored code but maybe not in the all

gaoqi8649 (approved):

libraries/aleth/libdevcore/CommonJS.h
JakubFornadel: oh yeah I did not read the code properly and mistakenly translated `return v.asUInt64();` as returning some default uint64 value, my bad

libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
JakubFornadel: we agreed that we are going to follow our style for all new or refactored code :)

Issue #1249, opened 2021-10-28
Vote processing while syncing


MatusKysel:
### Description
While syncing we still do vote processing, I do not understand why this is necessary, but even if it is somehow, we do not clean those votes, until we are 100%. So my local node had 2M votes in memory, then it verified them and just after that it was cleared. This should not be happening.

### Solutions

- Stop processing votes while syncing
- clean-up unverified votes even after sync pbft block is push to DB

MatusKysel:
https://github.com/Taraxa-project/taraxa-node/pull/1250#issuecomment-955072798

MatusKysel:
closed by https://github.com/Taraxa-project/taraxa-node/pull/1255

PR #1250, opened 2021-10-28
chore(votes): clean all vote while syncing

MatusKysel:
## Purpose

fixes: https://github.com/Taraxa-project/taraxa-node/issues/1249


gaoqi8649:
> @gaoqi8649 so what do you suggest ? How we should deal with that? because when I have tried to sync up with testnet I ended up with 2M unverified votes that where eating my memory

This is only happen on bootstrapping nodes. Node has so many unverified votes and cannot cleanup because syncing from beginning that takes a long time. After we introduce public DB for players download data, the node will only need to sync like within 24 hours newest data and not syncing from beginning. The bootstrapping nodes syncing will be very fast, and this will not happen.

PR #1255, opened 2021-11-03
chore(syncing): introduce deep syncing

MatusKysel:
## Purpose

Introduce deep syncing method, to skip specific packet processing. Please especially check this `filterSyncIrrelevantPackets`

### DIFF
--- a/libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
@@ -6,15 +6,12 @@
 #include "libp2p/Common.h"
 
 namespace taraxa::network::tarcap {
-
-class PeersState;
-
 /**
  * @brief SyncingState contains common members and functions related to syncing that are shared among multiple classes
  */
 class SyncingState {
  public:
-  SyncingState(std::shared_ptr<PeersState> peers_state);
+  SyncingState(uint16_t deep_syncing_threshold);
 
   /**
    * @brief Set pbft syncing
@@ -22,7 +19,8 @@ class SyncingState {
    * @param syncing
    * @param peer_id used in case syncing flag == true to set which peer is the node syncing with
    */
-  void set_pbft_syncing(bool syncing, const std::optional<dev::p2p::NodeID> &peer_id = {});
+  void set_pbft_syncing(bool syncing, uint32_t period_difference = 0,
+                        const std::optional<dev::p2p::NodeID> &peer_id = {});
 
   /**
    * @brief Set dag syncing
@@ -45,6 +43,7 @@ class SyncingState {
   bool is_actively_syncing() const;
 
   bool is_syncing() const;
+  bool is_deep_pbft_syncing() const;
   bool is_pbft_syncing() const;
   bool is_dag_syncing() const;
 
@@ -61,12 +60,13 @@ class SyncingState {
   void set_peer(const dev::p2p::NodeID &peer_id);
 
  private:
-  std::shared_ptr<PeersState> peers_state_{nullptr};
-
+  std::atomic<bool> deep_pbft_syncing_{false};
   std::atomic<bool> pbft_syncing_{false};
   std::atomic<bool> dag_syncing_{false};
 
   ExpirationCache<dev::p2p::NodeID> malicious_peers_;
+
+  const uint16_t kDeepSyncingThreshold;
 
   // Number of seconds needed for ongoing syncing to be declared as inactive
   static constexpr std::chrono::seconds SYNCING_INACTIVITY_THRESHOLD{60};
--- a/libraries/core_libs/network/include/network/tarcap/taraxa_capability.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/taraxa_capability.hpp
@@ -109,6 +109,8 @@ class TaraxaCapability : public dev::p2p::CapabilityFace {
                               const std::shared_ptr<DagBlockManager> &dag_blk_mgr,
                               const std::shared_ptr<TransactionManager> &trx_mgr, addr_t const &node_addr);
 
+  bool filterSyncIrrelevantPackets(SubprotocolPacketType packet_type) const;
+
  public:
   // TODO: Remove in future when tests are refactored
   // Test state
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
@@ -12,7 +12,7 @@ ExtVotesPacketHandler::ExtVotesPacketHandler(std::shared_ptr<PeersState> peers_s
 void ExtVotesPacketHandler::onNewPbftVote(std::shared_ptr<Vote> const &vote) {
   std::vector<dev::p2p::NodeID> peers_to_send;
   for (auto const &peer : peers_state_->getAllPeers()) {
-    if (!peer.second->isVoteKnown(vote->getHash())) {
+    if (!peer.second->isVoteKnown(vote->getHash()) && !peer.second->syncing_) {
       peers_to_send.push_back(peer.first);
     }
   }
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
@@ -173,7 +173,7 @@ bool StatusPacketHandler::sendStatus(const dev::p2p::NodeID& node_id, bool initi
     } else {
       success = sealAndSend(
           node_id, StatusPacket,
-          std::move(dev::RLPStream(5) << dag_max_level << pbft_chain_size << syncing_state_->is_pbft_syncing()
+          std::move(dev::RLPStream(5) << dag_max_level << pbft_chain_size << syncing_state_->is_deep_pbft_syncing()
                                       << pbft_round << pbft_previous_round_next_votes_size));
     }
   }
--- a/libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
+++ b/libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
@@ -37,7 +37,7 @@ TaraxaCapability::TaraxaCapability(std::weak_ptr<dev::p2p::Host> host, const dev
                                    std::shared_ptr<TransactionManager> trx_mgr, addr_t const &node_addr)
     : test_state_(std::make_shared<TestState>()),
       peers_state_(nullptr),
-      syncing_state_(std::make_shared<SyncingState>(peers_state_)),
+      syncing_state_(std::make_shared<SyncingState>(conf.deep_syncing_threshold)),
       node_stats_(nullptr),
       packets_handlers_(std::make_shared<PacketsHandler>()),
       thread_pool_(std::make_shared<TarcapThreadPool>(conf.network_packets_processing_threads, node_addr)),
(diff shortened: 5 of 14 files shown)


MatusKysel:
this should solve https://github.com/Taraxa-project/taraxa-node/issues/1249

libraries/core_libs/network/include/network/tarcap/taraxa_capability.hpp
```diff
@@ -109,6 +109,8 @@ class TaraxaCapability : public dev::p2p::CapabilityFace {
                               const std::shared_ptr<DagBlockManager> &dag_blk_mgr,
                               const std::shared_ptr<TransactionManager> &trx_mgr, addr_t const &node_addr);
 
+  inline bool filterSyncIrrelevantPackets(SubprotocolPacketType packet_type);
```
JakubFornadel, line 112: Why would you create inline class method ? I saw it in multiple PR's of yours... Inline can improve speed/size/etc... but it can also make it much worse. Compiler can actually make any function inline so why not let the decision to the compiler optimization ?

libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
```diff
@@ -61,13 +60,14 @@ class SyncingState {
...
  private:
-  std::shared_ptr<PeersState> peers_state_{nullptr};
-
+  std::atomic<bool> deep_pbft_syncing_{false};
   std::atomic<bool> pbft_syncing_{false};
   std::atomic<bool> dag_syncing_{false};
 
   ExpirationCache<dev::p2p::NodeID> malicious_peers_;
 
+  uint16_t deep_syncing_threshold_;
```
JakubFornadel, line 69: this should be constant

libraries/core_libs/network/include/network/tarcap/taraxa_capability.hpp
MatusKysel: inline is just suggestion for compiler https://en.cppreference.com/w/cpp/language/inline 
```
Since this meaning of the keyword inline is non-binding, compilers are free to use inline substitution for any function that's not marked inline, and are free to generate function calls to any function marked inline. Those optimization choices do not change the rules regarding multiple definitions and shared statics listed above. 
```

libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
```diff
@@ -61,13 +60,14 @@ class SyncingState {
...
  private:
-  std::shared_ptr<PeersState> peers_state_{nullptr};
-
+  std::atomic<bool> deep_pbft_syncing_{false};
   std::atomic<bool> pbft_syncing_{false};
   std::atomic<bool> dag_syncing_{false};
 
   ExpirationCache<dev::p2p::NodeID> malicious_peers_;
 
+  const uint16_t deep_syncing_threshold_;
```
JakubFornadel, line 69: naming... https://google.github.io/styleguide/cppguide.html#Constant_Names

MatusKysel: does not apply here as it is class member first and this part ` whose value is fixed for the duration of the program` does not apply here as this variable is always different per each instance

libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
```diff
@@ -321,13 +321,38 @@ void TaraxaCapability::interpretCapabilityPacket(std::weak_ptr<dev::p2p::Session
...
+inline bool TaraxaCapability::filterSyncIrrelevantPackets(SubprotocolPacketType packet_type) {
+  switch (packet_type) {
+    case DagBlockPacket:
+    case GetDagSyncPacket:
+    case DagSyncPacket:
+    case TransactionPacket:
+    case VotePacket:
+    case GetVotesSyncPacket:
+    case VotesSyncPacket:
+    case PbftBlockPacket:
```
JakubFornadel, line 343: all these case are basically transformed into the `if else` conditions that we dont need to check at all. I would call this method:

```
TaraxaCapability::isPbftSyncingRelevatPacket(SubprotocolPacketType packet_type) {
  switch (packet_type) {
     case TestPacket:
     case StatusPacket:
     case GetPbftSyncPacket:
     case PbftSyncPacket:
     case SyncedPacket:
       return true;
  }

  return false
}
```

MatusKysel: but the we will not catch bug if someone implements new type and do not set if it should be filtered out or not. And also compiler will do something like that

MatusKysel: also it is less clear what packet we are filtering or not

libraries/core_libs/network/include/network/tarcap/taraxa_capability.hpp
MatusKysel: because it is just more complicated if and it is in function just to have more clear code and call this function is much more expensive then doing that switch

libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
JakubFornadel: quite the opposite I think. If I create some new packet type which is related to deep syncing and wont add it to this switch, I am pretty sure that some syncing tests would fail so this is covered and if that new type is not related to deep syncing I should not be forced to bother with this function at all because it should check only few selected types that are related to deep syncing, nothing else... And the other argument that compiler will do something like that, well I dont see how it actually could remove valid if else condition. This function will be called basically with all the possible packet types so no way compiler will optimize it. Logically we want to check only those few selected types, nothing else so we can get rid of 8 useless if else conditions + 8 lines of code

libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
MatusKysel: if so, we where wrong https://stackoverflow.com/questions/67655378/what-is-the-proper-case-of-constant-class-member-fields-according-to-the-google this should not apply

libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
MatusKysel: yeah, you are probably right

libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
JakubFornadel: based on single stackoverflow answer with single upvote ? :) I guess I can live with it, but if I forget about all the rules I think it is isier to read if I know straight away which class member is constant and which is not

JakubFornadel (approved):

libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
MatusKysel: I am taking it back, you are right, it should be kDeepp......

libraries/core_libs/network/include/network/tarcap/taraxa_capability.hpp
JakubFornadel: I pretty much agree with [this answer](https://stackoverflow.com/questions/1759300/when-should-i-write-the-keyword-inline-for-a-function-method) (1k+ upvotes), which says:

It is said that inline hints to the compiler that you think the function should be inlined. That may have been true in 1998, but a decade later the compiler needs no such hints. Not to mention humans are usually wrong when it comes to optimizing code, so most compilers flat out ignore the 'hint'.

- When should I write the keyword 'inline' for a function/method in C++?

Only when you want the function to be defined in a header. More exactly only when the function's definition can show up in multiple translation units. It's a good idea to define small (as in one liner) functions in the header file as it gives the compiler more information to work with while optimizing your code. It also increases compilation time.

- When should I not write the keyword 'inline' for a function/method in C++?

Don't add inline just because you think your code will run faster if the compiler inlines it.

- When will the compiler not know when to make a function/method 'inline'?

Generally, the compiler will be able to do this better than you. However, the compiler doesn't have the option to inline code if it doesn't have the function definition. In maximally optimized code usually all private methods are inlined whether you ask for it or not.

but I am willing to change my opinion if you provde some better source of explanation and arguments :)

JakubFornadel: This [microsoft doc](https://docs.microsoft.com/en-us/cpp/cpp/inline-functions-cpp?view=msvc-160) on the other hand claims it is worth it. Anyway I dont know that much about it, so lets use it if you think it is worth it too :) But when googling it, I only found a form where you use inline keyword only in cpp file, not inside header declaration...

kstdl (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -12,7 +12,7 @@ ExtVotesPacketHandler::ExtVotesPacketHandler(std::shared_ptr<PeersState> peers_s
 void ExtVotesPacketHandler::onNewPbftVote(std::shared_ptr<Vote> const &vote) {
   std::vector<dev::p2p::NodeID> peers_to_send;
   for (auto const &peer : peers_state_->getAllPeers()) {
-    if (!peer.second->isVoteKnown(vote->getHash())) {
+    if (!peer.second->isVoteKnown(vote->getHash()) && !peer.second->syncing_) {
```
gaoqi8649, line 15: I think the peer.second->syncing_ is both for request blocks from peers and sending out blocks to peers.
For example, if a bootstrapping node is syncing from CN0, then other consensus nodes will not send votes to CN0. That's not right. In that case, we don't want sending votes to the bootstrapping node, but we still want sending votes to CN0

libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
```diff
@@ -321,13 +321,29 @@ void TaraxaCapability::interpretCapabilityPacket(std::weak_ptr<dev::p2p::Session
...
   thread_pool_->push(PacketData(packet_type, std::move(node_id), _r.data().toBytes()));
 }
 
+inline bool TaraxaCapability::filterSyncIrrelevantPackets(SubprotocolPacketType packet_type) const {
+  switch (packet_type) {
+    case TestPacket:
+    case StatusPacket:
+    case GetPbftSyncPacket:
+    case PbftSyncPacket:
+    case SyncedPacket:
```
gaoqi8649, line 340: I don't understand why include "TestPacket" and "SatusPacket". During syncing I think still need to update peers status
"GetPbftSyncPacket" is receiving request from peers. Does that mean if node is sending blocks to a peer, the node will not accept other nodes syncing request?
"PbftSyncPacket" is receiving PBFT blocks from peer. node is syncing from peer, but ignore PBFT blocks. Looks like wrong here.
If I am wrong please correct me here. I think the "SyncedPacket" is sender send to receiver the last packet. The receiver node is syncing, but ignore the last received packet. Looks like wrong.
Maybe I am wrong and I misunderstanding here

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -173,7 +173,7 @@ bool StatusPacketHandler::sendStatus(const dev::p2p::NodeID& node_id, bool initi
     } else {
       success = sealAndSend(
           node_id, StatusPacket,
-          std::move(dev::RLPStream(5) << dag_max_level << pbft_chain_size << syncing_state_->is_pbft_syncing()
+          std::move(dev::RLPStream(5) << dag_max_level << pbft_chain_size << syncing_state_->is_deep_pbft_syncing()
```
MatusKysel, line 176: @gaoqi8649 here it is were we are sending what should be in `syncing_` variable

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
MatusKysel: I do not understand your example, why other node would not send votes to CN0? CN0 not syncing anymore it will receive votes as usual.   `syncing_` variable is set base on your status packet, where node that is sending that packet specifies if it is syncing or not.

libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
MatusKysel: those are packets that we are accepting

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
gaoqi8649: If node1 is syncing from CN0, the "syncing_" for node1 is true. How about the "syncing_" for CN0? True or false?

MatusKysel: it is false, if CN0 is up to date

mfrankovi (approved):

gaoqi8649 (approved):

PR #1257, opened 2021-11-04
chore(vote_mngr): do not save all votes to DB

MatusKysel:
## Purpose

This PR should optimize what votes are we saving to DB, so now on it is only our votes and the rest we will receive after the node restarts

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -505,9 +505,15 @@ void PbftManager::initialState_() {
 
   auto soft_voted_block_hash = db_->getPbftMgrVotedValue(PbftMgrVotedValue::SoftVotedBlockHashInRound);
   auto soft_voted_block = db_->getPbftMgrStatus(PbftMgrStatus::SoftVotedBlockInRound);
+
   if (soft_voted_block_hash) {
     // From DB
     soft_voted_block_for_this_round_ = std::make_pair(*soft_voted_block_hash, soft_voted_block);
+    for (const auto &vote : db_->getSoftVotes(round)) {
+      if (vote->getBlockHash() == *soft_voted_block_hash) {
+        vote_mgr_->addVerifiedVote(vote);
+      }
+    }
   } else {
     // Default value
     soft_voted_block_for_this_round_ = std::make_pair(NULL_BLOCK_HASH, soft_voted_block);
--- a/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
+++ b/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
@@ -13,7 +13,8 @@ namespace taraxa {
 VoteManager::VoteManager(addr_t node_addr, std::shared_ptr<DbStorage> db, std::shared_ptr<FinalChain> final_chain,
                          std::shared_ptr<PbftChain> pbft_chain,
                          std::shared_ptr<NextVotesForPreviousRound> next_votes_mgr)
-    : db_(std::move(db)),
+    : node_addr_(std::move(node_addr)),
+      db_(std::move(db)),
       pbft_chain_(std::move(pbft_chain)),
       final_chain_(std::move(final_chain)),
       previous_round_next_votes_(std::move(next_votes_mgr)) {
@@ -30,29 +31,12 @@ VoteManager::~VoteManager() { daemon_->join(); }
 void VoteManager::setNetwork(std::weak_ptr<Network> network) { network_ = std::move(network); }
 
 void VoteManager::retreieveVotes_() {
-  LOG(log_si_) << "Retrieve unverified votes from DB";
-  auto unverified_votes = db_->getUnverifiedVotes();
-  for (auto const& v : unverified_votes) {
-    auto pbft_round = v->getRound();
-    auto hash = v->getHash();
-    {
-      UniqueLock lock(unverified_votes_access_);
-      if (unverified_votes_.count(pbft_round)) {
-        unverified_votes_[pbft_round][hash] = v;
-      } else {
-        std::unordered_map<vote_hash_t, std::shared_ptr<Vote>> votes{std::make_pair(hash, v)};
-        unverified_votes_[pbft_round] = std::move(votes);
-      }
-    }
-    LOG(log_dg_) << "Retrieved unverified vote " << *v;
-  }
-
   LOG(log_si_) << "Retrieve verified votes from DB";
   auto verified_votes = db_->getVerifiedVotes();
+  const auto pbft_step = db_->getPbftMgrField(PbftMgrRoundStep::PbftStep);
   for (auto const& v : verified_votes) {
     // Rebroadcast our own next votes in case we were partitioned...
-    if (v->getVoterAddr() == node_addr_ && v->getStep() >= FIRST_FINISH_STEP &&
-        db_->getPbftMgrField(PbftMgrRoundStep::PbftStep) > EXTENDED_PARTITION_STEPS) {
+    if (v->getStep() >= FIRST_FINISH_STEP && pbft_step > EXTENDED_PARTITION_STEPS) {
       std::vector<std::shared_ptr<Vote>> votes = {v};
       if (auto net = network_.lock()) {
         net->onNewPbftVotes(std::move(votes));
@@ -80,9 +64,6 @@ bool VoteManager::addUnverifiedVote(std::shared_ptr<Vote> const& vote) {
       std::unordered_map<vote_hash_t, std::shared_ptr<Vote>> votes{std::make_pair(hash, vote)};
       unverified_votes_[pbft_round] = std::move(votes);
     }
-
-    // Save vote also in db
-    db_->saveUnverifiedVote(vote);
   }
   LOG(log_nf_) << "Add unverified vote " << vote->getHash().abridged();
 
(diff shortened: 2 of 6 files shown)


libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -416,33 +365,22 @@ void VoteManager::cleanupVotes(uint64_t pbft_round) {
...
+  auto batch = db_->createWriteBatch();
   {
     UniqueLock lock(verified_votes_access_);
     auto it = verified_votes_.begin();
     while (it != verified_votes_.end() && it->first < pbft_round) {
       for (auto const& step : it->second) {
         for (auto const& voted_value : step.second) {
           for (auto const& v : voted_value.second) {
-            remove_verified_votes_hash.emplace_back(v.first);
+            if (v.second->getVoterAddr() == node_addr_) db_->removeVerifiedVoteToBatch(v.first, batch);
```
JakubFornadel, line 377: missing parenthesis for if

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -503,9 +503,15 @@ void PbftManager::initialState_() {
 
   auto soft_voted_block_hash = db_->getPbftMgrVotedValue(PbftMgrVotedValue::SoftVotedBlockHashInRound);
   auto soft_voted_block = db_->getPbftMgrStatus(PbftMgrStatus::SoftVotedBlockInRound);
+
   if (soft_voted_block_hash) {
     // From DB
     soft_voted_block_for_this_round_ = std::make_pair(*soft_voted_block_hash, soft_voted_block);
+    for (const auto &vote : db_->getSoftVotes(round)) {
+      if (vote->getBlockHash() == *soft_voted_block_hash) {
+        vote_mgr_->addVerifiedVote(vote);
```
JakubFornadel, line 512: so as your are basically only removing functionality that saved votes into the db, how is possible that this is needed and wasnt before ?

MatusKysel: yeah, so this is there in case of crash after soft votes are saved, before we had them in DB as all votes where saved, so no need to recover them.

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -30,29 +31,12 @@ VoteManager::~VoteManager() { daemon_->join(); }
...
-  }
-
   LOG(log_si_) << "Retrieve verified votes from DB";
   auto verified_votes = db_->getVerifiedVotes();
+  const auto pbft_step = db_->getPbftMgrField(PbftMgrRoundStep::PbftStep);
   for (auto const& v : verified_votes) {
     // Rebroadcast our own next votes in case we were partitioned...
-    if (v->getVoterAddr() == node_addr_ && v->getStep() >= FIRST_FINISH_STEP &&
-        db_->getPbftMgrField(PbftMgrRoundStep::PbftStep) > EXTENDED_PARTITION_STEPS) {
+    if (v->getStep() >= FIRST_FINISH_STEP && pbft_step > EXTENDED_PARTITION_STEPS) {
```
gaoqi8649, line 39: Here only rebroadcast the node own next votes, not rebroadcast other peers votes. So need "v->getVoterAddr() == node_addr_"

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -30,29 +31,12 @@ VoteManager::~VoteManager() { daemon_->join(); }
...
-      UniqueLock lock(unverified_votes_access_);
-      if (unverified_votes_.count(pbft_round)) {
-        unverified_votes_[pbft_round][hash] = v;
-      } else {
-        std::unordered_map<vote_hash_t, std::shared_ptr<Vote>> votes{std::make_pair(hash, v)};
-        unverified_votes_[pbft_round] = std::move(votes);
-      }
-    }
-    LOG(log_dg_) << "Retrieved unverified vote " << *v;
-  }
```
gaoqi8649, line 48: If don't retrieve those unverified votes, will need wait peers sending again?

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -292,9 +269,6 @@ void VoteManager::verifyVotes(uint64_t pbft_round, size_t sortition_threshold, u
                               std::function<size_t(addr_t const&)> const& dpos_eligible_vote_count) {
   // Cleanup votes for previous rounds
   cleanupVotes(pbft_round);
-
-  std::vector<std::shared_ptr<Vote>> future_unverifiable_votes;
```
gaoqi8649, line 296: The future_unverifiable_votes is for protect from malicious players. Malicious players could send huge votes in future round, like in round 1000000000000.

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
gaoqi8649: Why only remove node own votes and not remove other peers votes?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -503,9 +503,15 @@ void PbftManager::initialState_() {
...
+      }
+    }
```
gaoqi8649, line 514: If soft_voted_block_hash has value, the current round soft votes saved in soft_votes column in DB. Don't need add back to verified votes queue

gaoqi8649:
Justin and I discussed that if need save unverified votes in DB. 
For example, there are 3 consensus nodes CN0, CN1 and CN2. CN0 crash at step 100, CN1 crash at step 10000, CN2 crash at step 50000. After reboot the 3 nodes, they will lose all unverified votes, and until CN0 and CN1 get to step 50000, the 3 nodes will be able to make consensus.

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
MatusKysel: as I stated in `Purpose` of this PR, we are saving in DB only our own votes

MatusKysel: so there are no other votes to remove, that's why there is that condition

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
MatusKysel: we are saving in DB only our own votes, so no need to check anymore here

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
MatusKysel: yes, unverified votes are note saved as they could be irrelevant after restart e.g. network progress could be few pbft block meanwhile

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
MatusKysel: I think we need to, because votes are saved in DB but never retrieved.  Before we have also kept them in verified_votes column in DB, but now we only keep there our own votes. But maybe I do not understand whole picture and we are not using soft votes after we soft-voted block, then I think we should not save them to DB at all. What do you think @gaoqi8649 ?

MatusKysel:
> Justin and I discussed that if need save unverified votes in DB. For example, there are 3 consensus nodes CN0, CN1 and CN2. CN0 crash at step 100, CN1 crash at step 10000, CN2 crash at step 50000. After reboot the 3 nodes, they will lose all unverified votes, and until CN0 and CN1 get to step 50000, the 3 nodes will be able to make consensus.

So what will we lose? CN1 & CN2 -> 100 votes of CN0 and CN1 will lose 10000 votes of CN2 and same thing CN2, will lose 10000 votes of CN1. They still need to get to step 50000 as they were offline, so no big deal. Anyway you would still send those votes even with the old approach, because after restart we resend all our votes to the rest of the network.

JakubFornadel (approved):

kstdl (approved):

gaoqi8649:
> > Justin and I discussed that if need save unverified votes in DB. For example, there are 3 consensus nodes CN0, CN1 and CN2. CN0 crash at step 100, CN1 crash at step 10000, CN2 crash at step 50000. After reboot the 3 nodes, they will lose all unverified votes, and until CN0 and CN1 get to step 50000, the 3 nodes will be able to make consensus.
> 
> So what will we lose? CN1 & CN2 -> 100 votes of CN0 and CN1 will lose 10000 votes of CN2 and same thing CN2, will lose 10000 votes of CN1. They still need to get to step 50000 as they were offline, so no big deal. Anyway you would still send those votes even with the old approach, because after restart we resend all our votes to the rest of the network.

I mean I can think a race condition, if PBFT manager thread crash(or any issue cause vote manger cannot verify votes), the unverified votes will not be able to verify and they will stay in unverified queue. Then all 3 nodes will run out of memory. When crash CN0 is at step 100, CN1 is at step 10000 and CN2 is at step 50000. After reboot and vote manger is working again:

In old approach: All nodes will retrieve the votes from unverified queue. They don't need wait, and could make consensus by the unverified votes in DB. (Next votes start at step 4, so all 3 nodes has enough votes for consensus)

In new approach:
CN0 vote start at step 100
CN1 vote start at step 10000
CN2 vote start at step 50000
Until CN0 and CN1 reach step 50000, they could make consensus.

gaoqi8649 (approved):

PR #1258, opened 2021-11-05
chore(db): new option to remove old columns from DB

MatusKysel:
## Purpose

Added new options `--rebuild-db-columns`, that will remove old columns from DB and node can continue working even if DB scheme changes.

IMPORTANT:  `COLUMN(default_column);` is introduced again (it was removed, maybe even by me) but it needs to be there. This is default column for rocks DB, that can not be change. Right now we have there all period data and that's not good. So this is probably BREAKING CHANGE

### DIFF
--- a/libraries/core_libs/storage/src/storage.cpp
+++ b/libraries/core_libs/storage/src/storage.cpp
@@ -16,7 +16,7 @@ static constexpr uint16_t DAG_BLOCKS_POS_IN_PERIOD_DATA = 2;
 static constexpr uint16_t TRANSACTIONS_POS_IN_PERIOD_DATA = 3;
 
 DbStorage::DbStorage(fs::path const& path, uint32_t db_snapshot_each_n_pbft_block, uint32_t db_max_snapshots,
-                     uint32_t db_revert_to_period, addr_t node_addr, bool rebuild)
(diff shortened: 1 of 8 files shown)


libraries/core_libs/storage/src/storage.cpp
```diff
@@ -77,6 +81,32 @@ DbStorage::DbStorage(fs::path const& path, uint32_t db_snapshot_each_n_pbft_bloc
...
+  rocksdb::DB* db;
+  std::vector<std::string> column_families;
+  rocksdb::DB::ListColumnFamilies(options, db_path_.string(), &column_families);
+
+  std::vector<rocksdb::ColumnFamilyDescriptor> descriptors(column_families.size());
+  std::vector<rocksdb::ColumnFamilyHandle*> handles(column_families.size());
+  std::transform(column_families.begin(), column_families.end(), std::back_inserter(descriptors), [](const auto& name) {
+    return rocksdb::ColumnFamilyDescriptor(name, rocksdb::ColumnFamilyOptions());
+  });
+  checkStatus(rocksdb::DB::Open(options, db_path_.string(), descriptors, &handles, &db));
```
JakubFornadel, line 94: you can use smart pointer and get raw pointer by get() here

JakubFornadel (approved):

gaoqi8649 (approved):

PR #1259, opened 2021-11-06
chore(pbft): cleanup previous round votes when move to a new round

gaoqi8649:
## Version Notes ##
{ major | minor | patch }


JakubFornadel (approved):

MatusKysel (dismissed):
this is absolutely unnecessary as `resetRound_` round is called by `stateOperations_` where we are calling the same function inside `verifyVotes`

JakubFornadel:
> this is absolutely unnecessary as `resetRound_` round is called by `stateOperations_` where we are calling the same function inside `verifyVotes`

Nope, it is actually necessary. He is calling it to avoid copying votes for current round from verified queue back to unverified queue, which happens during `removeVerifiedVotes()`. It is very hard to read and understand the flow of whole pbft_manager... There are pretty big design issues thre

MatusKysel (approved):

PR #1262, opened 2021-11-09
chore(db): max_open_files setting for DB, to limit memory usage

MatusKysel:
## Purpose

This new setting should limit rocksDB memory usage, I think it is caused by `ulimit -n` that is set on our nodes to extreme number `1048576`. With this setting we can limit number of open files ~= RAM consumption. I am not 100% sure, but I would like to deploy it to our testnet and let's see if this can help

### DIFF
--- a/libraries/core_libs/storage/src/storage.cpp
+++ b/libraries/core_libs/storage/src/storage.cpp
@@ -15,8 +15,9 @@ static constexpr uint16_t CERT_VOTES_POS_IN_PERIOD_DATA = 1;
 static constexpr uint16_t DAG_BLOCKS_POS_IN_PERIOD_DATA = 2;
 static constexpr uint16_t TRANSACTIONS_POS_IN_PERIOD_DATA = 3;
 
-DbStorage::DbStorage(fs::path const& path, uint32_t db_snapshot_each_n_pbft_block, uint32_t db_max_snapshots,
-                     uint32_t db_revert_to_period, addr_t node_addr, bool rebuild, bool rebuild_columns)
+DbStorage::DbStorage(fs::path const& path, uint32_t db_snapshot_each_n_pbft_block, uint32_t max_open_files,
+                     uint32_t db_max_snapshots, uint32_t db_revert_to_period, addr_t node_addr, bool rebuild,
+                     bool rebuild_columns)
     : path_(path),
       handles_(Columns::all.size()),
       db_snapshot_each_n_pbft_block_(db_snapshot_each_n_pbft_block),
@@ -42,6 +43,10 @@ DbStorage::DbStorage(fs::path const& path, uint32_t db_snapshot_each_n_pbft_bloc
   options.create_missing_column_families = true;
   options.create_if_missing = true;
   options.compression = rocksdb::CompressionType::kLZ4Compression;
+  // This options is related to memory consuption
+  // https://github.com/facebook/rocksdb/issues/3216#issuecomment-817358217
+  // aleth default 256 (state_db is using another 128)
+  options.max_open_files = (max_open_files) ? max_open_files : 256;
   std::vector<rocksdb::ColumnFamilyDescriptor> descriptors;
   std::transform(Columns::all.begin(), Columns::all.end(), std::back_inserter(descriptors), [](const Column& col) {
     return rocksdb::ColumnFamilyDescriptor(col.name(), rocksdb::ColumnFamilyOptions());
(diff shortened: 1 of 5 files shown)


libraries/core_libs/storage/src/storage.cpp
```diff
@@ -90,7 +97,10 @@ void DbStorage::rebuildColumns(const rocksdb::Options& options) {
   std::transform(column_families.begin(), column_families.end(), std::back_inserter(descriptors), [](const auto& name) {
     return rocksdb::ColumnFamilyDescriptor(name, rocksdb::ColumnFamilyOptions());
   });
-  checkStatus(rocksdb::DB::Open(options, db_path_.string(), descriptors, &handles, &db));
+  auto db_ptr = static_cast<rocksdb::DB*>(nullptr);
```
JakubFornadel, line 100: why not simply `rocksdb::DB* db_ptr = nullptr` ?

JakubFornadel (approved):

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -59,8 +64,10 @@ DbStorage::DbStorage(fs::path const& path, uint32_t db_snapshot_each_n_pbft_bloc
   if (db_revert_to_period) {
     recoverToPeriod(db_revert_to_period);
   }
-
-  checkStatus(rocksdb::DB::Open(options, db_path_.string(), descriptors, &handles_, &db_));
+  rocksdb::DB* db = nullptr;
+  checkStatus(rocksdb::DB::Open(options, db_path_.string(), descriptors, &handles_, &db));
+  assert(db);
+  db_.reset(db);
```
gaoqi8649, line 70: why not use "rocksdb::DB* db_" directly? What is the benefits using "std::unique_ptr<rocksdb::DB> db_"? Thanks

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -81,7 +88,7 @@ DbStorage::DbStorage(fs::path const& path, uint32_t db_snapshot_each_n_pbft_bloc
 }
 
 void DbStorage::rebuildColumns(const rocksdb::Options& options) {
-  rocksdb::DB* db;
+  std::unique_ptr<rocksdb::DB> db;
```
gaoqi8649, line 91: Same here. Why change to "std::unique_ptr<rocksdb::DB> db"?

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -90,7 +97,10 @@ void DbStorage::rebuildColumns(const rocksdb::Options& options) {
   std::transform(column_families.begin(), column_families.end(), std::back_inserter(descriptors), [](const auto& name) {
     return rocksdb::ColumnFamilyDescriptor(name, rocksdb::ColumnFamilyOptions());
   });
-  checkStatus(rocksdb::DB::Open(options, db_path_.string(), descriptors, &handles, &db));
+  rocksdb::DB* db_ptr = nullptr;
+  checkStatus(rocksdb::DB::Open(options, db_path_.string(), descriptors, &handles, &db_ptr));
```
gaoqi8649, line 101: I think db_ptr should point to heap? Does it should delete db_ptr later?

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -59,8 +64,10 @@ DbStorage::DbStorage(fs::path const& path, uint32_t db_snapshot_each_n_pbft_bloc
   if (db_revert_to_period) {
     recoverToPeriod(db_revert_to_period);
   }
-
-  checkStatus(rocksdb::DB::Open(options, db_path_.string(), descriptors, &handles_, &db_));
+  rocksdb::DB* db = nullptr;
```
gaoqi8649, line 67: I think db should point to heap. Does it should delete db later?

libraries/core_libs/storage/src/storage.cpp
MatusKysel: so we do not need to take care of deleting it, in case some code change

libraries/core_libs/storage/src/storage.cpp
MatusKysel: same as before, so in case code change, unique_ptr will delete object

libraries/core_libs/storage/src/storage.cpp
MatusKysel: db_ptr points to nothing, then `rocksdb::DB::Open` will assign it object and destructor of unique_ptr will delete it when we return from function

libraries/core_libs/storage/src/storage.cpp
MatusKysel: I am not deleting anything I am just creating empty pointer to db

JakubFornadel (approved):

gaoqi8649 (approved):

PR #1266, opened 2021-11-10
chore(syncing): change condition to restart syncing

MatusKysel:
## Purpose

Start syncing even if we are just one block behind

### DIFF
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/common/ext_syncing_packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/common/ext_syncing_packet_handler.hpp
@@ -36,7 +36,7 @@ class ExtSyncingPacketHandler : public PacketHandler {
    * @param shared_state
    * @param force
    */
-  void restartSyncingPbft(bool force);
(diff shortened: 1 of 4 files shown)


gaoqi8649 (changes requested):
The design because we hit PBFT syncing blocks too often issue in previous time.
@justinsnapp and I discussed when trigger PBFT syncing dependent on peers status packet:
1. Only trigger restartSyncingPbft when node behind at least 2 blocks. When node just behind one block, will trigger PBFT syncing in PBFT manager.
2. When node behind more than 6 blocks, will trigger syncing in force

MatusKysel:
> The design because we hit PBFT syncing blocks too often issue in previous time. @justinsnapp and I discussed when trigger PBFT syncing dependent on peers status packet:
> 
>     1. Only trigger restartSyncingPbft when node behind at least 2 blocks. When node just behind one block, will trigger PBFT syncing in PBFT manager.
> 
>     2. When node behind more than 6 blocks, will trigger syncing in force

force syncing does not work like that, it should not be used there. I have change condition back

JakubFornadel (approved):

gaoqi8649 (approved):

PR #1275, opened 2021-11-11
feat(votes): introduce binominal distribution for votes weight valida…

MatusKysel:
…tion

Introducing binominal distribution to calculate weight of specific vote based on his stake. This work is based on alogrand implementation.

paper: https://people.csail.mit.edu/nickolai/papers/gilad-algorand-eprint.pdf 
TLDR (step 4): https://medium.com/@jsign.uy/the-intuition-behind-algorand-cryptographic-sortition-526e76e87e97 
code description: https://forum.algorand.org/t/where-in-go-algorand-code-is-leader-selected/1001/5


justinsnapp:
I am reading through the Algorand paper and this does almost exactly what I wanted to do.   I think in my original idea we would have computed a different proof for each value of j and then only had to check the specific proof, but that has two problems:  (1) the sender has to generate a lot of proofs and (2) the sender could withold a higher valued j / proof and send it later.     Whereas this approach has a single proof, but the receiver then has to loop through the computation of binomial distributions.  But this is cheaper to compute, and also has the big advantage that the value of j is uniquely determined and a higher j couldn't be withheld and later revealed by the sender.

justinsnapp (approved):

libraries/types/vote/include/vote/vote.hpp
```diff
@@ -67,6 +67,8 @@ struct VrfPbftSortition : public vrf_wrapper::VrfSortitionBase {
   }
   static inline uint512_t max512bits = std::numeric_limits<uint512_t>::max();
   bool canSpeak(size_t threshold, size_t valid_players) const;
+  uint64_t binominal_cdf(uint64_t stake, double threshold, double dpos_total_votes_count) const;
+  static uint64_t binominal_cdf(uint64_t stake, double threshold, double dpos_total_votes_count, uint512_t& output);
```
JakubFornadel, line 71: you forgot to change original formatting to camel case :)

JakubFornadel (approved):

gaoqi8649 (commented):
This is cool!
My only concern is that is in single calculation. In your unit test, you test in 1000 times, and the total result within 2% difference. How about the single calculation? If one single time generate more than 2 times of 2t+1 votes (that’s 2 * (2/3 total + 1) = 4/3 total + 2), that may break consensus. 
@justinsnapp @MatusKysel  How does the binomial distribution protect that will not generate more than (4/3 total + 2) in each single calculation? Thanks!

gaoqi8649:
> This is cool! My only concern is that is in single calculation. In your unit test, you test in 1000 times, and the total result within 2% difference. How about the single calculation? If one single time generate more than 2 times of 2t+1 votes (that’s 2 * (2/3 total + 1) = 4/3 total + 2), that may break consensus. @justinsnapp @MatusKysel How does the binomial distribution protect that will not generate more than (4/3 total + 2) in each single calculation? Thanks!

I think it more. Even in the current approach that cannot guarantee will never generate number of 2 times 2t+1 votes, since each sortition is an individual event. Let's just keep in mind, when number of delegation nodes increase, we may hit the issue.
The binomial distribution for weighted vote is good. So cool!

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1047,39 +1047,27 @@ void PbftManager::secondFinish_() {
...
-    if (step == 1 && weighted_index > 0) {
-      break;
-    }
-    if (shouldSpeak(vote_type, round, step_, weighted_index)) {
-      auto vote = generateVote(blockhash, vote_type, round, step, weighted_index);
-      votes.emplace_back(vote);
-      db_->saveVerifiedVote(vote);
-      vote_mgr_->addVerifiedVote(vote);
-    }
+  if (!weighted_votes_count_ || !getDposTotalVotesCount()) {
```
gaoqi8649, line 1058: I think don't need "!getDposTotalVotesCount()", that should never happen

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1112,8 +1100,10 @@ blk_hash_t PbftManager::calculateOrderHash(std::vector<DagBlock> const &dag_bloc
 
 std::pair<blk_hash_t, bool> PbftManager::proposeMyPbftBlock_() {
   auto round = getPbftRound();
-  // In propose block, only use first weighted index 0 for VRF sortition
-  if (weighted_votes_count_ == 0 || !shouldSpeak(propose_vote_type, round, step_, 0)) {
+  // In propose block, only use one vote for VRF sortition
+  VrfPbftSortition vrf_sortition(vrf_sk_, {propose_vote_type, propose_vote_type, step_});
+  if (weighted_votes_count_ == 0 ||
+      !vrf_sortition.getBinominalDistribution(1, sortition_threshold_, getDposTotalVotesCount())) {
```
gaoqi8649, line 1106: Here is wrong. In generateVote you don't only use stake 1 to generate proposal vote. Here need match. Either:
1. For any proposal_vote_type, only use stake 1 to generate vote
2. Or change here, using real stake to generate block

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
```diff
@@ -54,10 +54,10 @@ class NextVotesForPreviousRound {
 
   bool enough_votes_for_null_block_hash_;
   blk_hash_t voted_value_;  // For value is not null block hash
-  size_t next_votes_size_;
   // <voted PBFT block hash, next votes list that have exactly 2t+1 votes voted at the PBFT block hash>
   // only save votes == 2t+1 voted at same value in map and set
   std::unordered_map<blk_hash_t, std::vector<std::shared_ptr<Vote>>> next_votes_;
+  std::unordered_map<blk_hash_t, uint64_t> next_votes_weight_;
```
gaoqi8649, line 60: Since in current design, we save exactly 2t+1 next votes for previous round. In PBFT next voting state, node sometimes will votes on 2 values, NULL HASH and a voted value.
So in syncing next votes for previous round, we will always check if there are 2t+1 votes or 2 * (2t+1) votes.
The next_votes_weight_ will break that rule. Otherwise will cause lots of useless next votes syncing

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -311,11 +311,11 @@ size_t PbftManager::dposEligibleVoteCount_(addr_t const &addr) {
...
 }
 
-bool PbftManager::shouldSpeak(PbftVoteTypes type, uint64_t round, size_t step, size_t weighted_index) {
+bool PbftManager::shouldSpeak(PbftVoteTypes type, uint64_t round, size_t step) {
   // compute sortition
-  VrfPbftMsg msg(type, round, step, weighted_index);
+  VrfPbftMsg msg(type, round, step);
   VrfPbftSortition vrf_sortition(vrf_sk_, msg);
-  if (!vrf_sortition.canSpeak(sortition_threshold_, getDposTotalVotesCount())) {
+  if (!vrf_sortition.getBinominalDistribution(weighted_votes_count_, sortition_threshold_, getDposTotalVotesCount())) {
```
gaoqi8649, line 318: In my understand, getBinominalDistribution is generate how many weight the vote will have. 
I think should be:
auto weight = vrf_sortition.getBinominalDistribution(weighted_votes_count_, sortition_threshold_, getDposTotalVotesCount());
VrfPbftMsg msg(type, round, step, weight);
VrfPbftSortition vrf_sortition(vrf_sk_, msg);
if (!vrf_sortition.canSpeak(sortition_threshold_, getDposTotalVotesCount())) {

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -286,27 +289,12 @@ void VoteManager::verifyVotes(uint64_t pbft_round, size_t sortition_threshold, u
...
     bool vote_is_valid = true;
-
-    addr_t voter_account_address = v->getVoterAddr();
-    // Check if the voter account is valid, malicious vote
-    auto vote_weighted_index = v->getWeightedIndex();
-    auto dpos_votes_count = dpos_eligible_vote_count(voter_account_address);
-    if (vote_weighted_index > 0 && v->getStep() == 1) {
-      LOG(log_dg_) << "Account " << voter_account_address
-                   << " attempted to vote with weighted index > 0 in propose step. Vote: " << v;
-      vote_is_valid = false;
```
gaoqi8649, line 299: The part is for handle proposal vote type. Don't forget add the situation

libraries/types/vote/src/vote.cpp
```diff
@@ -61,14 +70,20 @@ Vote::Vote(secret_t const& node_sk, VrfPbftSortition const& vrf_sortition, blk_h
   vote_hash_ = sha3(true);
 }
 
-void Vote::validate(size_t dpos_total_votes_count, size_t sortition_threshold) const {
+void Vote::validate(uint64_t stake, double sortition_threshold, double dpos_total_votes_count) const {
+  if (!stake || !dpos_total_votes_count) {
```
gaoqi8649, line 74: I think don't need !dpos_total_votes_count, should never happen

libraries/types/vote/src/vote.cpp
```diff
@@ -61,14 +70,20 @@ Vote::Vote(secret_t const& node_sk, VrfPbftSortition const& vrf_sortition, blk_h
...
+  }
+
   if (!verifyVrfSortition()) {
     std::stringstream err;
     err << "Invalid vrf proof. " << *this;
     throw std::logic_error(err.str());
   }
 
-  if (!verifyCanSpeak(sortition_threshold, dpos_total_votes_count)) {
+  if (!calculateWeight(stake, sortition_threshold, dpos_total_votes_count)) {
```
gaoqi8649, line 86: I think here should be calculateWeight compare with vote weight:
if (calculateWeight(stake, sortition_threshold, dpos_total_votes_count) != getWeight())

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
gaoqi8649, line 318: should be getBinominalDistribution(weighted_votes_count_, getDposTotalVotesCount(), sortition_threshold_)?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
gaoqi8649: And same here, should switch sortition_threshold_ and getDposTotalVotesCount()?

libraries/types/vote/include/vote/vote.hpp
```diff
@@ -65,8 +61,13 @@ struct VrfPbftSortition : public vrf_wrapper::VrfSortitionBase {
   bool operator==(VrfPbftSortition const& other) const {
     return pk == other.pk && pbft_msg == other.pbft_msg && proof == other.proof && output == other.output;
   }
+
   static inline uint512_t max512bits = std::numeric_limits<uint512_t>::max();
-  bool canSpeak(size_t threshold, size_t valid_players) const;
+  static inline auto kMax512bFP = max512bits.convert_to<boost::multiprecision::mpfr_float>();
+  static uint64_t getBinominalDistribution(uint64_t stake, double threshold, double dpos_total_votes_count,
+                                           const uint512_t& output);
+  uint64_t getBinominalDistribution(uint64_t stake, double threshold, double dpos_total_votes_count) const;
```
gaoqi8649, line 69: Mismatch with implementation:
uint64_t VrfPbftSortition::getBinominalDistribution(uint64_t stake, double dpos_total_votes_count, double threshold, const uint512_t& output)

gaoqi8649 (commented):
The current design, for all delegation nodes who stay in 100% synced state, their block production rate is the same. So basic for those nodes, the block produced rate is “1 / number of delegation nodes (100% synced)“. we only use weight index 0 to do VRF sortition for proposal vote and block, no matter how many staking in the delegation nodes. This is because our 6 consensus nodes have majority staking. Don't want our CNs produce much more blocks than players.
After the change, what is the vote and block proposal rate?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
MatusKysel: no

libraries/types/vote/src/vote.cpp
MatusKysel: we are calculating it there so how you want to compare it with

libraries/types/vote/include/vote/vote.hpp
```diff
@@ -137,6 +143,7 @@ class Vote {
   VrfPbftSortition vrf_sortition_;
   mutable public_t cached_voter_;
   mutable addr_t cached_voter_addr_;
+  mutable std::optional<uint64_t> weight_;
```
JakubFornadel, line 146: does it need to be mutable ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1047,39 +1041,28 @@ void PbftManager::secondFinish_() {
...
+  if (!weighted_votes_count_) {
+    return 0;
   }
-
-  auto size = votes.size();
-
-  // pbft votes broadcast
-  if (size) {
+  auto vote = generateVote(blockhash, vote_type, round, step);
+  if (vote->calculateWeight(weighted_votes_count_, getDposTotalVotesCount(), sortition_threshold_)) {
```
JakubFornadel, line 1056: could be:
```
if (const auto weight = vote->calculateWeight(weighted_votes_count_, getDposTotalVotesCount(), sortition_threshold); weight) {

...

return weight.valuer();
}
```

JakubFornadel (approved):

libraries/types/vote/src/vote.cpp
```diff
@@ -82,15 +97,26 @@ void Vote::validate(size_t dpos_total_votes_count, size_t sortition_threshold) c
   }
 }
 
-bytes Vote::rlp(bool inc_sig) const {
+bytes Vote::rlp(bool inc_sig, bool inc_weight) const {
```
kstdl, line 100: do we really need inc_weight option? Maybe I am missing something, but it is set true everywhere

MatusKysel: I think we need it for next/soft votes, where we save them to DB and later use them, but I am planning to do big refactor of everything related to votes, so I think I will remove it in the future

kstdl (approved):

Issue #1281, opened 2021-11-15
End of life of PCRE 


MatusKysel:
Some of our dependencies use PCRE to be build, not sure witch one, but we should update ASAP

MatusKysel:
https://pcre.org/

PR #1283, opened 2021-11-16
chore(building): disable pcre library for cppcheck

MatusKysel:
## Purpose

Disable End of Life library, that is cppcheck dependency, but for feature we are not using

resolves https://github.com/Taraxa-project/taraxa-node/issues/1281


libraries/core_libs/network/rpc/rpc_error_handler.cpp
```diff
@@ -9,7 +9,6 @@ namespace taraxa::net {
 RpcServer::Error handle_rpc_error() {
   RpcServer::Error err;
   try {
```
kstdl, line 11: why do we need this empty try?

conanfile.py
```diff
@@ -14,8 +14,8 @@ class TaraxaConan(ConanFile):
 
     def requirements(self):
         self.requires("boost/1.76.0")
-        self.requires("cppcheck/2.3")
-        self.requires("openssl/1.1.1k")
+        self.requires("cppcheck/2.5") # 2.6 Segmentation fault (core dumped) while building
+        self.requires("openssl/1.1.1l")
```
kstdl, line 18: is there any need for that to be also updated?

MatusKysel: for openssl not, but why not if it is not breaking anything

gaoqi8649:
I test building on Mac on CircleCI, want to make sure will not effect in Mac build.
Failed at both in linux machine and Mac Machine. Looks like build environment setup issue. @Taraxa-project/devops please take a look. Thanks!

MatusKysel:
> I test building on Mac on CircleCI, want to make sure will not effect in Mac build. Failed at both in linux machine and Mac Machine. Looks like build environment setup issue. @Taraxa-project/devops please take a look. Thanks!

it looks like those builds are broken even without my changes

kstdl:
I have no issues with build on my mac. If you are not building from scratch try to rerun cmake to update conan dependencies

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -43,7 +43,6 @@ bool Dag::addVEEs(blk_hash_t const &new_vertex, blk_hash_t const &pivot, std::ve
   // add vertex
   vertex_t ret = add_vertex(new_vertex, graph_);
   boost::get(boost::vertex_index, graph_)[ret] = new_vertex;
-  edge_index_map_t weight_map = boost::get(boost::edge_index, graph_);
```
JakubFornadel, line 46: Yeah thi new copy of edge_index_map_t does not make sense as it is not used but wasn't this supposed to get reference/pointer rather than copy and set the values for weight_map, which is part of graph_ ?

MatusKysel: I do not know, I can just comment this code, so it will stays there?

kstdl (approved):

JakubFornadel (approved):

PR #1285, opened 2021-11-17
chore(peer): fix race condition in pending peers

MatusKysel:
There is race condition between getting peer and getting pending peer as it can change those states between those 2 calls

### DIFF
--- a/libraries/core_libs/network/src/tarcap/shared_states/peers_state.cpp
+++ b/libraries/core_libs/network/src/tarcap/shared_states/peers_state.cpp
@@ -17,7 +17,6 @@ std::shared_ptr<TaraxaPeer> PeersState::getPeer(const dev::p2p::NodeID& node_id)
 }
 
 std::shared_ptr<TaraxaPeer> PeersState::getPendingPeer(const dev::p2p::NodeID& node_id) const {
-  // TODO: pending_peers_ might have different mutex than peers ?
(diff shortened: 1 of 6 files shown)


gaoqi8649 (approved):

libraries/core_libs/network/src/tarcap/shared_states/peers_state.cpp
```diff
@@ -28,6 +27,22 @@ std::shared_ptr<TaraxaPeer> PeersState::getPendingPeer(const dev::p2p::NodeID& n
...
+std::pair<std::shared_ptr<TaraxaPeer>, bool> PeersState::getAnyPeer(const dev::p2p::NodeID& node_id) const {
+  std::shared_lock lock(peers_mutex_);
+
+  auto it_peer = peers_.find(node_id);
+  if (it_peer != peers_.end()) {
+    return {it_peer->second, false};
+  }
+
+  it_peer = pending_peers_.find(node_id);
+  if (it_peer != pending_peers_.end()) {
```
JakubFornadel, line 39: I would use this form for both ifs and not reuse it_peer as peers and pending peers are different structures.... 

```
if (auto it_peer = pending_peers_.find(node_id); it_peer != pending_peers_.end()) {
  ...
}
```

JakubFornadel (approved):

PR #1287, opened 2021-11-18
Rpc acc votes

mfrankovi:
RPC to get account vote and estimate block production

### DIFF
--- a/libraries/core_libs/network/rpc/Test.cpp
+++ b/libraries/core_libs/network/rpc/Test.cpp
@@ -165,6 +165,39 @@ Json::Value Test::get_account_balance(const Json::Value &param1) {
   return res;
 }
 
+Json::Value Test::get_account_votes(const Json::Value &param1) {
+  Json::Value res;
+  try {
+    if (auto node = full_node_.lock()) {
+      addr_t addr = addr_t(param1["address"].asString());
+      uint64_t block_number = param1["block_number"].asUInt64();
+      uint64_t range = 0;
+      if (!param1["range"].isNull()) {
+        range = param1["range"].asUInt64();
+      }
+
+      auto votes = node->getFinalChain()->dpos_eligible_vote_count(block_number, addr);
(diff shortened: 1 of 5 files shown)


libraries/core_libs/network/rpc/Test.cpp
```diff
@@ -165,6 +165,41 @@ Json::Value Test::get_account_balance(const Json::Value &param1) {
...
+      uint64_t range = 0;
+      if (!param1["range"].isNull()) {
+        range = param1["range"].asUInt64();
+      }
+
+      auto votes = node->getFinalChain()->dpos_eligible_vote_count(block_number, addr);
+      auto total_votes = node->getFinalChain()->dpos_eligible_total_vote_count(block_number);
+      res["votes"] = votes;
+      res["total_votes"] = total_votes;
+      double estimate = ((double)votes) / total_votes;
```
JakubFornadel, line 183: C style casts should not be used in modern C++. 
https://stackoverflow.com/questions/1609163/what-is-the-difference-between-static-cast-and-c-style-casting/18414126

libraries/core_libs/network/rpc/Test.cpp
```diff
@@ -165,6 +165,41 @@ Json::Value Test::get_account_balance(const Json::Value &param1) {
   return res;
 }
 
+Json::Value Test::get_account_votes(const Json::Value &param1) {
+  Json::Value res;
+  try {
+    if (auto node = full_node_.lock()) {
```
JakubFornadel, line 171: This nested if is not necessary - always try not to use nested if's due to readability. CAn be refactored to:
`Json::Value Test::get_account_votes(const Json::Value &param1) {
  Json::Value res;

  auto node = full_node_.lock();
  if (!node) {
    return res;
  }

  try {
    addr_t addr = addr_t(param1["address"].asString());
    uint64_t block_number = param1["block_number"].asUInt64();
    ...`

PR #1288, opened 2021-11-19
Genesis

kstdl:
## Purpose

Genesis hash now is not only dag block hash, but hash of whole genesis that includes consensus params. So if user change something in it he wont be able to connect to the network


gaoqi8649 (commented):
We definitely need this in Mainnet. But for Testnet, this will restrict cannot change any parameters for PBFT, sortition, and final chain. Every time we want to adjust parameters will need a complete reset.

JakubFornadel:
@kstdl what about this PR ? We should decide if we are going to use this or not

kstdl:
> @kstdl what about this PR ? We should decide if we are going to use this or not

Most of the changes in develop branch will need to be merged with a hardfork to be applied on the mainnet candidate. So we can try to merge also this. But first we need to implement configs migration 
But it is more just a structural change and I am not sure that it is worth it

JakubFornadel:
> > @kstdl what about this PR ? We should decide if we are going to use this or not
> 
> Most of the changes in develop branch will need to be merged with a hardfork to be applied on the mainnet candidate. So we can try to merge also this. But first we need to implement configs migration But it is more just a structural change and I am not sure that it is worth it

@kstdl can we close it then ?

libraries/core_libs/node/src/node.cpp
```diff
@@ -78,9 +78,10 @@ void FullNode::init() {
                                         conf_.test_params.db_max_open_files, conf_.test_params.db_max_snapshots,
                                         conf_.test_params.db_revert_to_period, node_addr);
     }
-
     if (db_->getNumDagBlocks() == 0) {
       db_->saveDagBlock(conf_.chain.dag_genesis_block);
+      db_->setGenesisHash(conf_.chain.genesisHash());
+      std::cout << "set genesis hash " << conf_.chain.genesisHash() << std::endl;
```
JakubFornadel, line 84: std::cout

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -234,6 +234,23 @@ void DbStorage::disableSnapshots() { snapshots_enabled_ = false; }
 
 void DbStorage::enableSnapshots() { snapshots_enabled_ = true; }
 
+void DbStorage::setGenesisHash(const h256& genesis_hash) {
+  if (!exist(0, Columns::genesis)) {
+    std::cout << "SETTING " << genesis_hash << std::endl;
```
JakubFornadel, line 239: std::cout

libraries/config/src/chain_config.cpp
```diff
@@ -114,4 +114,19 @@ decltype(ChainConfig::predefined_) const ChainConfig::predefined_([] {
...
-}  // namespace taraxa::chain_config
+bytes ChainConfig::rlp() const {
+  dev::RLPStream s;
+  s.appendList(5);
+
+  s << chain_id;
+  s << dag_genesis_block.rlp(true);
+  s << sortition.rlp();
+  s << pbft.rlp();
+  s << final_chain.rlp();
```
JakubFornadel, line 125: Missing gas_price and dag ? 

```
  DagBlock dag_genesis_block;
  GasPriceConfig gas_price;
  SortitionConfig sortition;
  PbftConfig pbft;
  final_chain::Config final_chain;
  DagConfig dag;
```

libraries/core_libs/node/src/node.cpp
```diff
@@ -90,29 +91,31 @@ void FullNode::init() {
   final_chain_ = NewFinalChain(db_, conf_.chain.final_chain, node_addr);
   trx_mgr_ = std::make_shared<TransactionManager>(conf_, db_, final_chain_, node_addr);
 
-  auto genesis_hash = conf_.chain.dag_genesis_block.getHash();
-  auto dag_genesis_hash_from_db = *db_->getBlocksByLevel(0).begin();
-  if (genesis_hash != dag_genesis_hash_from_db) {
-    LOG(log_er_) << "The DAG genesis block hash " << genesis_hash << " in config is different with "
-                 << dag_genesis_hash_from_db << " in DB";
+  auto genesis_hash = conf_.chain.genesisHash();
+  auto genesis_hash_from_db = db_->getGenesisHash();
```
JakubFornadel, line 95: based on return value, genesis_hash_from_db could be empty optional

libraries/cli/include/cli/config.hpp
```diff
@@ -29,7 +29,6 @@ class Config {
   static constexpr const char* CONFIG = "config";
   static constexpr const char* DATA_DIR = "data-dir";
   static constexpr const char* NETWORK_ID = "network-id";
-  static constexpr const char* BOOT_NODE = "boot-node";
```
MatusKysel, line 32: please also remove line `--boot-node true` from docker

MatusKysel: probably all appearances of  `--boot-node`

libraries/core_libs/node/src/node.cpp
```diff
@@ -78,9 +78,10 @@ void FullNode::init() {
                                         conf_.test_params.db_max_open_files, conf_.test_params.db_max_snapshots,
                                         conf_.test_params.db_revert_to_period, node_addr);
     }
-
     if (db_->getNumDagBlocks() == 0) {
       db_->saveDagBlock(conf_.chain.dag_genesis_block);
```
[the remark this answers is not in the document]
kstdl: After removing it I see errors like this. So it still should be separate tesk
```ERROR: Cannot find pivot dag block #4c6c3798```

File: for_devs/local-net [hunk 7]
```diff
@@ -125,7 +125,7 @@ def config(binary, boot_nodes, rpc_nodes):
             cmd += f'--boot-nodes {boot_nodes_flags} '
 
             if type == "boot":
-                cmd += '--boot-node '
+                # cmd += '--boot-node ' // TODO: we should replace this with bootnode binary?
```
MatusKysel, line 128: we do not need boot nodes on local net I guess

MatusKysel (approved):

for_devs/local-net
kstdl: But we have to specify count of bootnodes for some reason. So maybe we should remove it option and logic related to this

kstdl: I will create separate task for this

mfrankovi (approved):

PR #1297, opened 2021-11-24
Deep sync improvement

mfrankovi:
Deep sync state will be true until we are completely synced. With this change, the deep sync state will be set to false as soon as syncing gets close to being synced.

### DIFF
--- a/libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
+++ b/libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
@@ -8,27 +8,34 @@ namespace taraxa::network::tarcap {
 SyncingState::SyncingState(uint16_t deep_syncing_threshold)
     : malicious_peers_(300, 50), kDeepSyncingThreshold(deep_syncing_threshold) {}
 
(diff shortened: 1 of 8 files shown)


JakubFornadel (approved):

libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
```diff
@@ -8,40 +8,47 @@ namespace taraxa::network::tarcap {
...
+void SyncingState::set_peer(const std::shared_ptr<TaraxaPeer> &peer) {
   std::unique_lock lock(peer_mutex_);
-  peer_id_ = peer_id;
+  peer_ = peer;
 }
 
 const dev::p2p::NodeID &SyncingState::syncing_peer() const {
   std::shared_lock lock(peer_mutex_);
-  return peer_id_;
+  return peer_->getId();
```
MatusKysel, line 18: is there a chance, when this can be nullptr?

mfrankovi: There shouldn't be, but I added the nullptr check anyway

MatusKysel (approved):

PR #1299, opened 2021-11-25
fix: not remove old VrfParamsChanges to validate old dag block correctly

kstdl:
## Purpose

Keep old VrfParamsChanges to validate old dag block correctly

### DIFF
--- a/libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
+++ b/libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
@@ -117,7 +117,6 @@ void SortitionParamsManager::cleanup(uint64_t current_period) {
   }
   auto batch = db_->createWriteBatch();
   db_->cleanupDagEfficiencies(current_period);
-  db_->cleanupParamsChanges(batch, config_.changes_count_for_average);
   db_->commitWriteBatch(batch);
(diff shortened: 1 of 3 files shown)


mfrankovi (approved):

MatusKysel:
I do not like this solution, we should try to save as little as possible. I will approve it for now, but this should be just temporary and we should figure out how to deal with old dags

MatusKysel (approved):

JakubFornadel (approved):

justinsnapp:
> I do not like this solution, we should try to save as little as possible. I will approve it for now, but this should be just temporary and we should figure out how to deal with old dags

@kstdl and @MatusKysel can we save only the info needed to recompute these values should we need them for when a very old dag block is generated?

MatusKysel:
> > I do not like this solution, we should try to save as little as possible. I will approve it for now, but this should be just temporary and we should figure out how to deal with old dags
> 
> @kstdl and @MatusKysel can we save only the info needed to recompute these values should we need them for when a very old dag block is generated?

We should merge this so we can confirm that this is the problem and let's try to figure out how to deal with old DAGs, bacause this will be issue for rewards also. With this current approach we need to save a lot of data

PR #1303, opened 2021-11-27
fix: get not availiable param changes from db to validate old dag blocks

kstdl:


libraries/core_libs/storage/src/storage.cpp
```diff
@@ -423,6 +423,16 @@ std::deque<SortitionParamsChange> DbStorage::getLastSortitionParams(size_t count
   return changes;
 }
 
+SortitionParamsChange DbStorage::getParamsChangeForPeriod(uint64_t period) {
+  auto it =
```
JakubFornadel, line 427: why not simply:

auto data = lookup(toSlice(period), Columns::sortition_params_change);
assert(!data.empty());

return SortitionParamsChange::from_rlp(dev::RLP(asBytes(data)));

JakubFornadel (approved):

libraries/core_libs/storage/src/storage.cpp
kstdl: Will it return closest previous element? Because we have changes written only every interval blocks. For example we will have an record on key 50, but will search for 56. In that case `SeekForPrev` will definitely return what we need

JakubFornadel: ah yeah you are right

PR #1304, opened 2021-11-29
fix(build): modify linking library name CONAN_PKG::GTest

gaoqi8649:


JakubFornadel (changes requested):
You cant just change lib name. This would break linux builds... How is it possible that @kstdl can build just fine ?

PR #1305, opened 2021-11-29
Uint rocksdb comparator

MatusKysel:
## Purpose

Add custom UINT rocksdb comparator

### DIFF
--- a/libraries/core_libs/storage/include/storage/storage.hpp
+++ b/libraries/core_libs/storage/include/storage/storage.hpp
@@ -125,7 +125,7 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
     COLUMN(final_chain_receipt_by_trx_hash);
     COLUMN(final_chain_log_blooms_index);
     COLUMN(pbft_block_dag_efficiency);
-    COLUMN_W_COMP(sortition_params_change, getUintComparator());
+    COLUMN_W_COMP(sortition_params_change, getIntComparator<uint64_t>());
 
 #undef COLUMN
 #undef COLUMN_W_COMP
--- a/libraries/core_libs/storage/include/storage/storage.hpp
+++ b/libraries/core_libs/storage/include/storage/storage.hpp
@@ -87,9 +87,9 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
    public:
     static inline auto const& all = all_;
 
+#define COLUMN(__name__) static inline auto const __name__ = all_.emplace_back(#__name__, all_.size())
 #define COLUMN_W_COMP(__name__, ...) \
   static inline auto const __name__ = all_.emplace_back(#__name__, all_.size(), __VA_ARGS__)
-#define COLUMN(__name__) static inline auto const __name__ = all_.emplace_back(#__name__, all_.size())
 
     // do not change/move
     COLUMN(default_column);
--- a/libraries/core_libs/storage/include/storage/storage.hpp
+++ b/libraries/core_libs/storage/include/storage/storage.hpp
@@ -15,6 +15,7 @@
 #include "logger/logger.hpp"
 #include "pbft/pbft_block.hpp"
 #include "pbft/sync_block.hpp"
+#include "storage/uint_comparator.hpp"
 #include "transaction/transaction.hpp"
 
 namespace taraxa {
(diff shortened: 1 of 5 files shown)


libraries/core_libs/storage/src/storage.cpp
```diff
@@ -48,9 +49,12 @@ DbStorage::DbStorage(fs::path const& path, uint32_t db_snapshot_each_n_pbft_bloc
   // https://github.com/facebook/rocksdb/issues/3216#issuecomment-817358217
   // aleth default 256 (state_db is using another 128)
   options.max_open_files = (max_open_files) ? max_open_files : 256;
-  std::vector<rocksdb::ColumnFamilyDescriptor> descriptors;
+
+  std::vector<rocksdb::ColumnFamilyDescriptor> descriptors(Columns::all.size());
   std::transform(Columns::all.begin(), Columns::all.end(), std::back_inserter(descriptors), [](const Column& col) {
-    return rocksdb::ColumnFamilyDescriptor(col.name(), rocksdb::ColumnFamilyOptions());
+    auto options = rocksdb::ColumnFamilyOptions();
+    options.comparator = getUintComparator();
```
kstdl, line 56: I think we dont need it for all columns. Only for `sortition_params_change` column

MatusKysel: this was just to test, your test

libraries/core_libs/storage/include/storage/storage.hpp
```diff
@@ -91,6 +87,8 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
    public:
     static inline auto const& all = all_;
 
+#define COLUMN_W_COMP(__name__, ...) \
```
MatusKysel, line 90: if you know how to have just one define, let me know :)

libraries/core_libs/storage/include/storage/uint_comparator.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+#include "storage/db_utils.hpp"
+
+namespace taraxa {
+class UintComparator : public rocksdb::Comparator {
+ public:
+  UintComparator() = default;
+  ~UintComparator() = default;
+
+  static const char *kClassName() { return "taraxa.UintComparator"; }
```
JakubFornadel, line 13: What is kClassName() for ?

MatusKysel: I wanted to make same as rocksdb https://github.com/facebook/rocksdb/blob/2e09a54c4fb82e88bcaa3e7cfa8ccbbbbf3635d5/util/comparator.cc#L30

libraries/core_libs/storage/include/storage/storage.hpp
```diff
@@ -79,8 +73,10 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
 
    public:
     size_t const ordinal_;
+    const rocksdb::Comparator* comparator_;
```
JakubFornadel, line 76: I dont like this usage of raw pointer. I understand you probably need raw pointer for setting `rocksdb::ColumnFamilyOptions().comparator`, but it seems that you could use unique_ptr here ? :

- change raw ptr to unique_ptr
- make it private:
- create method that would return raw pointer from unique_ptr and use that for setting ColumnFamilyOptions
- getIntComparator() can return unique_ptr too

MatusKysel: uniqe_ptr does not make sense as more columns could use same compactor, I can create shared_ptr,  if you think is better

JakubFornadel: well this comment was about usage of smart_pointers vs raw pointers... choose whatever type of smart pointer is suitable. We have two options:

- write this class according to the best practices(no raw pointers, not using methods we dont need like that kClassName(), ...)
- or we can use this copy paste style of comparator class from rocksdb that implemented and keep the style according to rocksdb

MatusKysel: Actually I do not see any point using shared_ptrs as this approached (copied from rocksdb) is kind of singleton and there is no need to use shared_ptrs over there

MatusKysel: let's discuss this over the meeting

PR #1306, opened 2021-11-29
chore: add hardforks and enable_vrf_adjustion_block option to it. some  small fixes

kstdl:
## Purpose

Add ability to disable vrf params adjusting on start and enable it after with hardfork functionality

### DIFF
--- a/libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
+++ b/libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
@@ -1,6 +1,7 @@
 #include "dag/sortition_params_manager.hpp"
 
 #include "pbft/pbft_block.hpp"
+
 namespace taraxa {
 
 SortitionParamsChange::SortitionParamsChange(uint64_t period, uint16_t efficiency, const VrfParams& vrf,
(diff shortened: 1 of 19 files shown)


libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -173,13 +187,14 @@ std::optional<SortitionParamsChange> SortitionParamsManager::calculateChange(uin
   }
 
   const int32_t change = getChange(period, average_dag_efficiency);
-  config_.vrf += change;
+  VrfParams vrf = config_.vrf;
+  vrf += change;
```
JakubFornadel, line 191: just making sure you wanted to change this from saving the change is class member into not saving it ?

MatusKysel:
maybe you should also increase version or genesis timestamp so when we merge it, it will be ready to deploy

MatusKysel (approved):

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
kstdl: It moved here https://github.com/Taraxa-project/taraxa-node/pull/1306/files#diff-740ffa012af5a6c39db38b0cba5ee86455365dd3397a774838fd38b043e5a8c6R156. I noticed and thought that method with a name "calculateChange" shouldn't change anything. So I moved it and made this method const

JakubFornadel: ah I see, ok

PR #1312, opened 2021-12-01
checkout staking exist before vote verification

gaoqi8649:
After deep syncing, node could receive votes but still behind, may don't have vote sender state in table

### DIFF
--- a/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
+++ b/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
@@ -289,36 +289,25 @@ void VoteManager::verifyVotes(uint64_t pbft_round, size_t sortition_threshold, u
     if (votes_invalid_in_current_final_chain_period_.count(v->getHash())) {
       continue;
     }
-    bool vote_is_valid = true;
+
     auto dpos_votes_count = dpos_eligible_vote_count(v->getVoterAddr());
-    if (!dpos_votes_count) {
-      // After deep syncing, node could receive votes but still behind, may don't have vote sender state in table
-      LOG(log_nf_) << "Node may behind and still in syncing. Cannot find vote sender state in table " << v;
-      vote_is_valid = false;
+    if (dpos_votes_count && v->getStep() == 1) {
+      // We need to handle propose_vote_type
+      dpos_votes_count = 1;
     }
-
-    if (vote_is_valid) {
-      if (v->getStep() == 1) {
-        // We need to handle propose_vote_type
-        dpos_votes_count = 1;
-      }
-      try {
-        v->validate(dpos_votes_count, dpos_total_votes_count, sortition_threshold);
-      } catch (const std::logic_error& e) {
-        LOG(log_er_) << e.what();
-        vote_is_valid = false;
-      }
-    }
-
-    if (vote_is_valid) {
-      addVerifiedVote(v);
-      removeUnverifiedVote(v->getRound(), v->getHash());
-    } else {
+    try {
+      v->validate(dpos_votes_count, dpos_total_votes_count, sortition_threshold);
+    } catch (const std::logic_error& e) {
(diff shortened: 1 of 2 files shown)


MatusKysel (approved):

MatusKysel (commented):
let's change line `LOG(log_er_) << e.what();` to warn, so communing do not freak out

JakubFornadel:
> let's change line `LOG(log_er_) << e.what();` to warn, so communing does not freak out

I dont see any LOG(log_er_) nor  e.what() ?

MatusKysel:
> > let's change line `LOG(log_er_) << e.what();` to warn, so communing does not freak out
> 
> I dont see any LOG(log_er_) nor e.what() ?

check the whole file

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -291,12 +291,17 @@ void VoteManager::verifyVotes(uint64_t pbft_round, size_t sortition_threshold, u
     }
     bool vote_is_valid = true;
```
JakubFornadel, line 292: you can remove this variable  - no need to use it

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -291,12 +291,17 @@ void VoteManager::verifyVotes(uint64_t pbft_round, size_t sortition_threshold, u
...
     auto dpos_votes_count = dpos_eligible_vote_count(v->getVoterAddr());
-    if (v->getStep() == 1 && v->getType() == propose_vote_type) {
-      // We need to handle propose_vote_type
-      if (!dpos_votes_count) vote_is_valid = false;
-      dpos_votes_count = 1;
+    if (!dpos_votes_count) {
+      // After deep syncing, node could receive votes but still behind, may don't have vote sender state in table
+      LOG(log_nf_) << "Node may behind and still in syncing. Cannot find vote sender state in table " << v;
+      vote_is_valid = false;
```
JakubFornadel, line 297: you can use continue here insetad of vote_is_valid + remove if on line 300 so we get rif of one nested if

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -291,12 +291,17 @@ void VoteManager::verifyVotes(uint64_t pbft_round, size_t sortition_threshold, u
     }
     bool vote_is_valid = true;
     auto dpos_votes_count = dpos_eligible_vote_count(v->getVoterAddr());
-    if (v->getStep() == 1 && v->getType() == propose_vote_type) {
-      // We need to handle propose_vote_type
-      if (!dpos_votes_count) vote_is_valid = false;
-      dpos_votes_count = 1;
+    if (!dpos_votes_count) {
+      // After deep syncing, node could receive votes but still behind, may don't have vote sender state in table
+      LOG(log_nf_) << "Node may behind and still in syncing. Cannot find vote sender state in table " << v;
```
JakubFornadel, line 296: I am not sure if I am ok with this - rather no than yes I guess. You handled 1 edge case when when node did not apply all synced pbft blocks yet - this will be happening very rarely at the start of node. But what about malicious/invalid votes ? I mean if you receive vote and voter has 0 dpos stake that should log error(or warning depends how we want to treat it). If you want to be so specific and log case when synced pbft blocks are not synced yet I guess you can check that explicitly ?

justinsnapp: "Node may behind and still in syncing." should be "Node may be still syncing".     But shouldn't the node KNOW if its still syncing?   And why is this an INFO level message?    In addition I think @JakubFornadel  makes valid points.

JakubFornadel (approved):

Issue #1313, opened 2021-12-01
Assert in dag.cpp


MatusKysel:
My local node crashed on this

```
taraxad: /home/matus/dev/taraxa-node/libraries/core_libs/consensus/src/dag/dag.cpp:384: void taraxa::DagManager::worker(): Assertion `trx != nullptr' failed.
```

PR #1314, opened 2021-12-01
Nonfinalized trx fix

mfrankovi:
Transactions that were already finalized in a pbft block were incorrectly included in the nonfinalized_transactions_in_dag_ structure and in db when transactions were included in another dag block creating a memory leak. Fix is checking if the transactions is already in the db before saving it.

The fix includes removing all incorrectly saved transactions in recoverNonfinalizedTransactions from DB on restart so that existing node data is properly processed on startup. This part can later be removed

### DIFF
--- a/libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
@@ -164,8 +164,16 @@ void TransactionManager::saveTransactionsFromDagBlock(SharedTransactions const &
 
 void TransactionManager::recoverNonfinalizedTransactions() {
   // On restart populate nonfinalized_transactions_in_dag_ structure from db
+  auto write_batch = db_->createWriteBatch();
   for (auto const &trx : db_->getNonfinalizedTransactions()) {
-    nonfinalized_transactions_in_dag_.emplace(trx->getHash());
+    if (db_->transactionFinalized(trx->getHash())) {
+      // TODO: This section where transactions are deleted is only to recover from a bug where non-finalized
+      // transactions were not properly deleted from the DB, once there are no nodes with such corrupted data, this
+      // line can be removed oe replaced with an assert
+      db_->removeTransactionToBatch(trx->getHash(), write_batch);
+    } else {
+      nonfinalized_transactions_in_dag_.emplace(trx->getHash());
+    }
   }
 }
 
--- a/libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
@@ -133,6 +133,7 @@ uint32_t TransactionManager::insertBroadcastedTransactions(const SharedTransacti
 unsigned long TransactionManager::getTransactionCount() const { return trx_count_; }
 
 std::shared_ptr<Transaction> TransactionManager::getTransaction(trx_hash_t const &hash) const {
+  std::shared_lock transactions_lock(transactions_mutex_);
   auto trx_it = transactions_pool_.get(hash);
   if (trx_it.second) {
     return trx_it.first;
@@ -151,16 +152,16 @@ void TransactionManager::saveTransactionsFromDagBlock(SharedTransactions const &
   auto trx_in_db = db_->transactionsInDb(trx_hashes);
   for (uint64_t i = 0; i < trxs.size(); i++) {
     auto const &trx_hash = trx_hashes[i];
+    // We only save transaction if it has not already been saved
+    if (!trx_in_db[i]) {
+      db_->addTransactionToBatch(*trxs[i], write_batch);
+      nonfinalized_transactions_in_dag_.emplace(trx_hash);
+    }
     if (transactions_pool_.erase(trx_hash)) {
       LOG(log_dg_) << "Transaction " << trx_hash << " removed from trx pool";
       // Transactions are counted when included in DAG
       trx_count_++;
       transaction_accepted_.emit(trx_hash);
-    }
-    // We only save transaction if it has not already been saved
-    if (!trx_in_db[i]) {
-      db_->addTransactionToBatch(*trxs[i], write_batch);
-      nonfinalized_transactions_in_dag_.emplace(trx_hash);
     }
   }
   db_->addStatusFieldToBatch(StatusDbField::TrxCount, trx_count_, write_batch);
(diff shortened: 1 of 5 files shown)


libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -140,26 +140,40 @@ std::shared_ptr<Transaction> TransactionManager::getTransaction(trx_hash_t const
...
       trx_count_++;
+      transaction_accepted_.emit(trx->getHash());
     }
-    transaction_accepted_.emit(trx->getHash());
+    // We only save transaction if it has not already been saved
+    if (!db_->transactionInDb(trx->getHash())) {
+      db_->addTransactionToBatch(*trx, write_batch);
+      nonfinalized_transactions_in_dag_.emplace(trx->getHash());
+    }
+    db_->commitWriteBatch(write_batch);
```
JakubFornadel, line 160: this commit should be probably outside of for loop ?

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -140,26 +140,40 @@ std::shared_ptr<Transaction> TransactionManager::getTransaction(trx_hash_t const
...
   }
   db_->saveStatusField(StatusDbField::TrxCount, trx_count_);
```
JakubFornadel, line 162: this could be part of write batch and not separate call

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -140,26 +140,40 @@ std::shared_ptr<Transaction> TransactionManager::getTransaction(trx_hash_t const
...
-    nonfinalized_transactions_in_dag_.emplace(trx->getHash());
     if (transactions_pool_.erase(trx->getHash())) {
       LOG(log_dg_) << "Transaction " << trx->getHash() << " removed from trx pool";
       // Transactions are counted when included in DAG
       trx_count_++;
+      transaction_accepted_.emit(trx->getHash());
     }
-    transaction_accepted_.emit(trx->getHash());
+    // We only save transaction if it has not already been saved
+    if (!db_->transactionInDb(trx->getHash())) {
```
JakubFornadel, line 156: Here is potentionally up to 250 separate call to db. Maybe we could use 1 call to db(multiget or what we use for getting multiple keys) before loop, save hashes in unordered_set and then check that set here ?

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -140,26 +140,40 @@ std::shared_ptr<Transaction> TransactionManager::getTransaction(trx_hash_t const
...
+      nonfinalized_transactions_in_dag_.emplace(trx->getHash());
+    }
+    db_->commitWriteBatch(write_batch);
   }
   db_->saveStatusField(StatusDbField::TrxCount, trx_count_);
 }
 
 void TransactionManager::recoverNonfinalizedTransactions() {
   // On restart populate nonfinalized_transactions_in_dag_ structure from db
+  auto write_batch = db_->createWriteBatch();
```
JakubFornadel, line 167: you are not commiting the batch anywhere ?

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -140,26 +140,40 @@ std::shared_ptr<Transaction> TransactionManager::getTransaction(trx_hash_t const
...
   for (auto const &trx : db_->getNonfinalizedTransactions()) {
-    nonfinalized_transactions_in_dag_.emplace(trx->getHash());
+    if (db_->transactionFinalized(trx->getHash())) {
```
JakubFornadel, line 169: again multiget would be much more efficient

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -140,26 +140,40 @@ std::shared_ptr<Transaction> TransactionManager::getTransaction(trx_hash_t const
...
+      // TODO: This section where transactions are deleted is only to recover from a bug where non-finalized
+      // transactions were not properly deleted from the DB, once there are no nodes with such corrupted data, this
+      // line can be removed oe replaced with an assert
+      db_->removeTransactionToBatch(trx->getHash(), write_batch);
+    } else {
+      nonfinalized_transactions_in_dag_.emplace(trx->getHash());
```
JakubFornadel, line 175: there is no lock for nonfinalized_transactions_in_dag_, are we sure there is no possible concurrent access ?

mfrankovi: I fixed all your comments. This one I don't think there is a need to protect it since it is called from constructor but I added the lock anyway since we do protect it on all other places.

MatusKysel (approved):

MatusKysel:
resolves https://github.com/Taraxa-project/taraxa-node/issues/1313

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -133,34 +133,62 @@ uint32_t TransactionManager::insertBroadcastedTransactions(const SharedTransacti
...
   // Unique lock here makes sure that transactions we are removing are not reinserted in transactions_pool_
   std::unique_lock transactions_lock(transactions_mutex_);
-  for (auto const &trx : trxs) {
-    nonfinalized_transactions_in_dag_.emplace(trx->getHash());
-    if (transactions_pool_.erase(trx->getHash())) {
-      LOG(log_dg_) << "Transaction " << trx->getHash() << " removed from trx pool";
+  auto write_batch = db_->createWriteBatch();
+  vec_trx_t trx_hashes;
+  std::transform(trxs.begin(), trxs.end(), std::back_inserter(trx_hashes),
+                 [](std::shared_ptr<Transaction> const &t) { return t->getHash(); });
```
JakubFornadel, line 151: 1 last minor thing: we agreed to use `const Type&` style but you use everywhere the opposite Type const& style

JakubFornadel (approved):

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
mfrankovi: Do you think I should change it everywhere in the transaction_manager.cpp or just in the places I touched. We do have a mixture in the code now. Maybe we should just change this everywhere in the code at once?

JakubFornadel: I think anything you touch is ok, we can change it everywhere in separate PR that will do it throughout the codebase....

PR #1315, opened 2021-12-01
chore(votes): test binominal dist equality for same values

MatusKysel:
## Purpose
 test equality of binominal distribution calculation for same values


justinsnapp:
huh? what is this really testing?

MatusKysel:
> huh? what is this really testing?

@gaoqi8649 was asking in slack, if weigh of vote (binominal dist.) is constant across the nodes for same values, because he saw a lot of errors. So this test should show, for same input we always calculate same output

PR #1316, opened 2021-12-01
Non finalized transactions fix

mfrankovi:
Fix for non finalized transactions memory leak

### DIFF
--- a/libraries/core_libs/consensus/include/transaction_manager/transaction_manager.hpp
+++ b/libraries/core_libs/consensus/include/transaction_manager/transaction_manager.hpp
@@ -31,9 +31,9 @@ class TransactionManager : public std::enable_shared_from_this<TransactionManage
   SharedTransactions packTrxs(uint16_t max_trx_to_pack = 0);
 
   /**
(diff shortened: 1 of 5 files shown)


MatusKysel (approved):

JakubFornadel (commented):
So why dont we rather merge develop (which already has it) into the testnet ?

justinsnapp:
> So why dont we rather merge develop (which already has it) into the testnet ?

Is that only change on develop?

JakubFornadel:
> > So why dont we rather merge develop (which already has it) into the testnet ?
> 
> Is that only change on develop?

I think so, will double check

justinsnapp:
> > > So why dont we rather merge develop (which already has it) into the testnet ?
> > 
> > 
> > Is that only change on develop?
> 
> I think so, will double check
> 
> Checked and yes this is the only change

ok then @JakubFornadel  if you want to close this and merge develop to testnet, please do.  :-)

justinsnapp:
Merged develop to testnet, and this change already included in similar PR to develop today.  Thanks @mfrankovi !

PR #1319, opened 2021-12-02
chore(syncing): we can only set peer to nullptr if both syncing flags…

MatusKysel:
… are false

### DIFF
--- a/libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
+++ b/libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
@@ -10,6 +10,10 @@ SyncingState::SyncingState(uint16_t deep_syncing_threshold)
 
 void SyncingState::set_peer(std::shared_ptr<TaraxaPeer> &&peer) {
   std::unique_lock lock(peer_mutex_);
+  // we set peer to null if dag and pbft syncing are not enabled
+  if (peer == nullptr && (pbft_syncing_ || dag_syncing_)) {
+    return;
+  }
   peer_ = std::move(peer);
 }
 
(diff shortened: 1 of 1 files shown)


libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
```diff
@@ -42,14 +43,19 @@ void SyncingState::set_pbft_syncing(bool syncing, uint64_t current_period, std::
...
 
-void SyncingState::set_dag_syncing(bool syncing, std::shared_ptr<TaraxaPeer> peer) {
+void SyncingState::set_dag_syncing(bool syncing, std::shared_ptr<TaraxaPeer> peer /*=nullptr*/) {
   assert((syncing && peer) || !syncing);
   dag_syncing_ = syncing;
-  set_peer(std::move(peer));
+  // We set a peer if peer is not null and we are syncing or
+  // we set peer to null if dag and pbft syncing are not enabled
+  if ((syncing && peer) || (!syncing && !pbft_syncing_)) {
+    set_peer(std::move(peer));
```
mfrankovi, line 57: I think we are basically checking the same thing in two places in set_pbft_syncing and set_dag_syncing, maybe on both places we only call set_peer and then inside of set_peer we do the check. But I am fine as is as well

mfrankovi (approved):

libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
MatusKysel: you are right, that is better

libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
```diff
@@ -10,6 +10,10 @@ SyncingState::SyncingState(uint16_t deep_syncing_threshold)
 
 void SyncingState::set_peer(std::shared_ptr<TaraxaPeer> &&peer) {
   std::unique_lock lock(peer_mutex_);
+  // we set peer to null if dag and pbft syncing are not enabled
+  if (!peer && (pbft_syncing_ || dag_syncing_)) {
```
kstdl, line 14: I think it is the tough part of code to read. Could we explicit set it to null at least?

kstdl: I guess it is just easier to read then that one. But I am not sure if there could be a situation that value are other then null value here. 
```
if (!pbft_syncing_ && !dag_syncing) {
  peer_ = nullptr;
}
```

kstdl (approved):

PR #1321, opened 2021-12-02
chore: difficulty adjustment testnet fork

mfrankovi:
This PR is made only for the testnet to enable difficulty adjustment changes in a hard fork without changing the db format. A different change is needed for develop branch.

k_threshold_testnet_hard_fork_period defines the period on which the vrf upper bound will be reduced even though lower bound reached 0. To implement this without modifying the current testnet db format, there is a range constant k_threshold_range which defines a constant difference between lover and upper bound to be able to recover once the lower bound reaches 0 and upper bound is reduced below this range.

Upper bound can now be reduced up to k_threshold_upper_min_value = 0x50

### DIFF
--- a/libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
+++ b/libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
@@ -77,6 +77,8 @@ class SortitionParamsManager {
   int32_t getChange(uint64_t period, uint16_t efficiency) const;
   void cleanup(uint64_t current_period);
 
+  const uint64_t k_threshold_testnet_hard_fork_period = 50000;
+
   LOG_OBJECTS_DEFINE
 };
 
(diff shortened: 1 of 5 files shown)


libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
```diff
@@ -77,6 +77,8 @@ class SortitionParamsManager {
   int32_t getChange(uint64_t period, uint16_t efficiency) const;
   void cleanup(uint64_t current_period);
 
+  const uint64_t k_threshold_testnet_hard_fork_period = 50000;
```
MatusKysel, line 80: I know it is only for testnet but we vote, that constant values naming should be kConstatValue

MatusKysel:
do we need to change this? https://github.com/Taraxa-project/taraxa-node/blob/develop/libraries/core_libs/network/src/tarcap/taraxa_capability.cpp#L371 so we will disconnect peers with old fork version?

justinsnapp (approved):

PR #1322, opened 2021-12-02
chore: vrf range improvement

mfrankovi:
Improvement to enable vrf upper range to be lowered up to a minimum level even if lower bound is at zero

### DIFF
--- a/libraries/vdf/src/sortition.cpp
+++ b/libraries/vdf/src/sortition.cpp
@@ -10,7 +10,10 @@ VdfSortition::VdfSortition(SortitionParams const& config, vrf_sk_t const& sk, by
   difficulty_ = calculateDifficulty(config);
 }
 
-bool VdfSortition::isOmitVdf(SortitionParams const& config) const { return threshold <= config.vrf.threshold_lower; }
+bool VdfSortition::isOmitVdf(SortitionParams const& config) const {
+  return config.vrf.threshold_upper >= config.vrf.threshold_range &&
+         threshold <= config.vrf.threshold_upper - config.vrf.threshold_range;
+}
 
 bool VdfSortition::isStale(SortitionParams const& config) const { return threshold > config.vrf.threshold_upper; }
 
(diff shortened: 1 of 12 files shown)


libraries/vdf/src/sortition.cpp
```diff
@@ -10,7 +10,10 @@ VdfSortition::VdfSortition(SortitionParams const& config, vrf_sk_t const& sk, by
   difficulty_ = calculateDifficulty(config);
 }
 
-bool VdfSortition::isOmitVdf(SortitionParams const& config) const { return threshold <= config.vrf.threshold_lower; }
+bool VdfSortition::isOmitVdf(SortitionParams const& config) const {
+  return config.vrf.threshold_upper >= config.vrf.threshold_range &&
```
kstdl, line 14: I guess that one should be easier to understand 
```threshold <= std::max(config.vrf.threshold_upper - config.vrf.threshold_range, 0)```

File: libraries/vdf/src/sortition.cpp [hunk 2]
```diff
@@ -96,7 +99,7 @@ void VdfSortition::verifyVdf(SortitionParams const& config, bytes const& vrf_inp
       throw InvalidVdfSortition(
           "VDF solution verification failed. Incorrect difficulty. VDF input " + bytes2str(vdf_input) + ", lambda " +
           std::to_string(config.vdf.lambda_bound) + ", difficulty " + std::to_string(getDifficulty()) +
-          ", expected: " + std::to_string(expected) + ", vrf_params: (" + std::to_string(config.vrf.threshold_lower) +
+          ", expected: " + std::to_string(expected) + ", vrf_params: (" + std::to_string(config.vrf.threshold_range) +
```
kstdl, line 102: We should change the log because now it is not looks like a range

libraries/vdf/src/sortition.cpp
mfrankovi: Both of these values are unsigned integers. Does this still work even if threshold_upper is smaller than threshold_range?

libraries/vdf/src/sortition.cpp
mfrankovi: Everywhere in the code we now use upper and range including the config files. I think that it makes sense that it is in the log as well. Maybe we can add some additional text in the log if it is unclear

kstdl: I am just talking about clearer text. Because (num1, num2) looks like a range, but it is not now
I expect something like `vrf_params: range=500, threshold_upper: 1000` would be better

libraries/vdf/src/sortition.cpp
kstdl: I dont sure it is defined behaviour, but I test it and it was negative. But we can also cast one of that values to int32_t for example and expect it to be valid  in any situation

kstdl (approved):

libraries/vdf/src/sortition.cpp
mfrankovi: Your solution is shorter and more readable but I think that casting to int_32 is less efficient. Maybe @MatusKysel or @JakubFornadel give their opinion on this

justinsnapp: LGTM as is

PR #1327, opened 2021-12-06
chore(final_chain): move block to final chain instead reading from DB

MatusKysel:


gaoqi8649 (approved):

JakubFornadel:
All the functionality from this PR is implemented in my rewards PR so if we merge this, it is going to be just 1 huge conflict for me :/

MatusKysel:
> All the functionality from this PR is implemented in my rewards PR so if we merge this, it is going to be just 1 huge conflict for me :/

so why did not you create this PR before? it is not good practice to combine all changes in one huge PR if it is possible to split it

JakubFornadel:
> > All the functionality from this PR is implemented in my rewards PR so if we merge this, it is going to be just 1 huge conflict for me :/
> 
> so why did not you create this PR before? it is not good practice to combine all changes in one huge PR if it is possible to split it

well my mistake then. To be honest I have been working on it for such a long time now and faced so many frustrating problems that I simply did not separate it  as it was developed incrementally all together so I would have to create new branch and copy paste there only some part of the new code, which was like naah... 

I will gladly let you do the next huge task(even all of them) so you can split it into the 10 separate PR's and do it the right way :)

JakubFornadel (approved):

kstdl:
Maybe for develop we should wait for @JakubFornadel PR to be merged to not spend time on merging? If we want it for some reason on testnet than we can merge it to testnet directly

PR #1328, opened 2021-12-07
chore(dag): create seen cache to not request same block multiple times

MatusKysel:
## Purpose

When we received dag block with missing something from node A, we request missing blocks, but when meanwhile we receive same block from node B,C ... we to the same thing. So, we basically DDoS ourselfs. I have create small cache that should solve this behaviour

### DIFF
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
@@ -20,9 +20,8 @@ DagBlockPacketHandler::DagBlockPacketHandler(std::shared_ptr<PeersState> peers_s
                               std::move(pbft_chain), std::move(pbft_mgr), std::move(dag_mgr), std::move(dag_blk_mgr),
                               std::move(db), node_addr, "DAG_BLOCK_PH"),
       test_state_(std::move(test_state)),
-      trx_mgr_(std::move(trx_mgr)) {}
-
-thread_local std::mt19937_64 DagBlockPacketHandler::urng_{std::mt19937_64(std::random_device()())};
+      trx_mgr_(std::move(trx_mgr)),
+      seen_dags_(10000, 100) {}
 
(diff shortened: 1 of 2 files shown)


gaoqi8649 (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
```diff
@@ -31,6 +30,12 @@ void DagBlockPacketHandler::process(const PacketData &packet_data, const std::sh
...
 
+  if (!seen_dags_.insert(block.getHash())) {
+    LOG(log_dg_) << "Received dag block" << block.getHash() << " (from " << packet_data.from_node_id_.abridged()
+                 << ") already seen.";
+    return;
+  }
+
   if (dag_blk_mgr_) {
     // Do not process this block in case we already have it
     if (dag_blk_mgr_->isDagBlockKnown(block.getHash())) {
```
JakubFornadel, line 41: so how is your new cache different to this one ? I dont like the idea having two caches that are doing basically the same thing, just checked on different places

MatusKysel: isDagBlockKnown saves only valid dag blocks

MatusKysel: we need to have cache for requested blocks and this should kind of be that, maybe I should implement it only for requested blocks.

JakubFornadel: I see, ok... It would be nice differentiate then between these caches somehow. If you open this in 1 month and one cache is called seen_dags, another is called seen_blocks you will have no idea what is the difference between them. So at least some comment should be there on declaration and maybe the change name of the other cache so it is clear that is caches only valid_seen_blocks and this one all of the blocks...

JakubFornadel (approved):

PR #1332, opened 2021-12-09
chore(bootnode): improve bootnode settings

MatusKysel:


kstdl (approved):

rattrap (approved):
LGTM

programs/taraxa-bootnode/main.cpp
```diff
@@ -152,7 +152,7 @@ int main(int argc, char** argv) {
 
   auto net_conf = public_ip.empty() ? dev::p2p::NetworkConfig(listen_ip, listen_port, false)
                                     : dev::p2p::NetworkConfig(public_ip, listen_ip, listen_port, false);
-  net_conf.allowLocalDiscovery = allowLocalDiscovery;
+  net_conf.allowLocalDiscovery = !denyLocalDiscovery;
```
JakubFornadel, line 155: out of curiosity: why did you change it from allow to deny and program arg is deny but in config is allow ? :D

MatusKysel: netconfig is also part of taraxa node, I would need to change it everywhere in code, but we need this reverse logic just for bootnodes

JakubFornadel: yeah I am just curious what was the reasoning behind this change. Isnt having "allow" set to false same thing as "deny" to true ?

JakubFornadel (approved):

programs/taraxa-bootnode/main.cpp
MatusKysel: it is just because of boost program options 
```
addNetworkingOption("deny-local-discovery", po::bool_switch(&denyLocalDiscovery),
```
if you use this flag you can set it only to true

MatusKysel: and we want to enable by default, so we need flag to disable it

JakubFornadel: aaah ok now I get it

PR #1335, opened 2021-12-09
Hardfork 1.1

MatusKysel:


MatusKysel:
I want to include this one https://github.com/Taraxa-project/taraxa-node/pull/1333 when it is approved

MatusKysel:
@mfrankovi does not work on already created DB, we need to "recreate" column I guess
```
Dynamic exception type: taraxa::DbException
std::exception::what: Db error. Status code: 4 SubCode: 0 Message:Invalid argument: taraxa.UintComparator: does not match existing comparator leveldb.BytewiseComparator
```

gaoqi8649:
I recommend revert "change parameters for propose_block vote":

I review the PR1331 change parameters for propose block vote. I still think the old approach is better.
As I explained, In current Testnet: There are 713 delegation nodes, dpos total votes count is 2615
threshold = committee size = 1000
sortition rate = threshold / dpos_total_votes_count = 1000/2615 = 38%
Since in proposal state all delegation nodes only have 1 chance to win sortition, all delegation nodes have 38% rate to propose a PBFT block. They have the same probability. And there should be 713 * 38% = 273 proposal PBFT blocks for each period. I think that’s enough.

When a node add in Testnet as staking 1, will also add 2 stakes to our one of 6 consensus nodes. In order to make our 6 CN have majority.
Assume number of players as “n”, in mathematical model the proposal PBFT blocks per period should be (6 + n) * (threshold / (6 + 3n)). When “n” goes to be a big number, so there should have 1000 / 3 = 333 proposal PBFT blocks per period constantly. I think that’s enough.

In the new approach:
Sortition rate = threshold / total voters count, and all players have 1 chance to win the sortition that's same as old approach
Proposal PBFT blocks per period = total players * sortition rate = n * threshold / n = threshold, “n” is number of players. When n goes to a big number, that will propose 1000 PBFT blocks per period.

Since we only finalize 1 PBFT block per period, I think 1000 proposal blocks are too much and will cause network traffic. And the old approach 333 is better.

PR #1337, opened 2021-12-10
Binary Search for Binomial Distribution

gaoqi8649:


tests/crypto_test.cpp
```diff
@@ -332,6 +332,19 @@ TEST_F(CryptoTest, sortition_rate) {
   EXPECT_EQ(count, valid_sortition_players * round);
 }
 
+TEST_F(CryptoTest, binomial_distribution) {
```
MatusKysel, line 335: this test does not test anything

libraries/types/vote/src/vote.cpp
```diff
@@ -33,16 +33,28 @@ bytes VrfPbftSortition::getRlpBytes() const {
 
 uint64_t VrfPbftSortition::getBinominalDistribution(uint64_t stake, double dpos_total_votes_count, double threshold,
                                                     const uint512_t& output) {
+  if (!stake) return 0;  // Stake is 0
+
   auto l = static_cast<uint512_t>(output).convert_to<boost::multiprecision::mpfr_float>();
   auto division = l / kMax512bFP;
   double ratio = division.convert_to<double>();
```
MatusKysel, line 40: let's improve this and make all of them `const`

libraries/types/vote/src/vote.cpp
```diff
@@ -33,16 +33,28 @@ bytes VrfPbftSortition::getRlpBytes() const {
...
-    // Found the correct boundary, break
-    if (ratio <= cdf(dist, j)) {
-      return j;
+
+  // Binary search for find the lowest stake <= cdf
+  uint64_t start = 0;
+  uint64_t end = stake - 1;
+  while (start + 1 < end) {
+    auto mid = start + (end - start) / 2;
+    auto target = cdf(dist, mid);
```
MatusKysel, line 48: also `const`

tests/crypto_test.cpp
gaoqi8649: I like your test. I modify it a litter bit. It could test both cases with "total count < committee size" and "total count >= committee size". The test could test if the getBinominalDistribution always return same value with introduce binary search.
If you don't mind, I prefer to keep the test here.

libraries/types/vote/src/vote.cpp
gaoqi8649: fixed

libraries/types/vote/src/vote.cpp
gaoqi8649: Fixed

tests/crypto_test.cpp
MatusKysel: I just do no understand when `getBinominalDistribution` with same params could return different result

MatusKysel (approved):

PR #1342, opened 2021-12-13
Revert "chore(votes): change parameters for propose_block vote"

MatusKysel:
This reverts commit c005bfacdaa6fa208d7625dc63cb8065040e0c33.

### DIFF
--- a/libraries/core_libs/consensus/include/final_chain/final_chain.hpp
+++ b/libraries/core_libs/consensus/include/final_chain/final_chain.hpp
@@ -56,7 +56,7 @@ class FinalChain {
                                           std::optional<EthBlockNumber> blk_n = {},
                                           std::optional<state_api::ExecutionOptions> const& opts = {}) const = 0;
 
-  virtual uint64_t dpos_eligible_address_count(EthBlockNumber blk_num) const = 0;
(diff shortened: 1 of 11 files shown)


gaoqi8649 (approved):

gaoqi8649 (changes requested):
As @justinsnapp request, I add a proposal committee size for PBFT proposal votes/blocks.
https://github.com/Taraxa-project/taraxa-node/pull/1343
I think we don't need revert it

justinsnapp:
> As @justinsnapp request, I add a proposal committee size for PBFT proposal votes/blocks. #1343 I think we don't need revert it

Why would that mean we don't need to revert this change?

gaoqi8649:
> > As @justinsnapp request, I add a proposal committee size for PBFT proposal votes/blocks. #1343 I think we don't need revert it
> 
> Why would that mean we don't need to revert this change?

Because @MatusKysel change to do sortition for proposal PBFT votes/blocks base on total voters instead of total votes. And in https://github.com/Taraxa-project/taraxa-node/pull/1343, I add a proposal committee size as 200. That means the committee size is 200 for proposal state, and for other states the committee size is still 1000

JakubFornadel (approved):

PR #1343, opened 2021-12-14
Add proposal committee size for PBFT proposal votes and blocks

gaoqi8649:


MatusKysel (approved):

justinsnapp:
Should we go all in and make a committee size for each vote type?

gaoqi8649:
> Should we go all in and make a committee size for each vote type?

I think soft , cert, next vote type should have the same committee size. Why do you think we should make different committee size for them?

JakubFornadel (approved):

PR #1344, opened 2021-12-14
Syncing improvements

mfrankovi:
This PR is missing a more consistent and comprehensive handling of malicious nodes which I believe is another large task that needs to be done. As part of this PR I have left a lot of TODO: for places where such malicious node handling might be needed. 

Changes are:

**PBFT syncing**

Pbft manager no longer controls the syncing process. Status messages are used for restarting pbft syncing if needed. Previously a lot of time restartSyncing was called in cases where there was no need for it. Pbft syncing is started only if:
1. There is a node at least 2 pbft periods ahead
2. There is a node 1 pbft period ahead for the same period for two consecutive status messages  

**DAG syncing**

DAG syncing is refactored and simplified:
Each connected peer will request all the missing DAG blocks level from a connected peer in only one of these two cases:
1. On connecting if the pbft periods of both nodes are matching
2. On receiving a first dag block with mising tip/pivot if node is pbft synced

This request can only be performed once. After this is performed it is expected that nodes are DAG synced. If after this sync, another block with missing tip/pivot is received a node is disconnected.

Main difference to what we had before is that there is no global dag_syncing flag anymore. We are now ensuring dag syncing state between each connected peer. As long as this is true DAG gossiping should work without any missing tip/pivot errors.

### DIFF
--- a/libraries/core_libs/consensus/src/dag/dag.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag.cpp
@@ -375,15 +375,18 @@ void DagManager::worker() {
       continue;
     }
     level_limit = false;
-    SharedTransactions transactions;
     if (pivotAndTipsAvailable(*blk)) {
-      // Retrieve all the transactions
-      for (auto const &trx_hash : blk->getTrxs()) {
-        auto trx = trx_mgr_->getTransaction(trx_hash);
-        // DAG block should not have been verified if it is missing a transaction
-        assert(trx != nullptr);
-        transactions.emplace_back(std::move(trx));
+      // Retrieve pool transactions
+      auto [transactions, missing_trxs] = trx_mgr_->getPoolTransactions(blk->getTrxs());
+
+      // DAG block should not have been verified if it is missing a transaction
+      // This check is taking some resources so in time we can remove it but it is safe to have it as a sanity check
+      size_t trx_found_count = 0;
+      for (const auto &b : db_->transactionsInDb(missing_trxs)) {
+        if (b) trx_found_count++;
       }
+      assert(missing_trxs.size() == trx_found_count);
+
       addDagBlock(*blk, std::move(transactions));
       LOG(log_time_) << "Broadcast block " << blk->getHash() << " at: " << getCurrentTimeMilliSeconds();
     } else {
@@ -414,6 +417,13 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
       trx_mgr_->saveTransactionsFromDagBlock(trxs);
       // Save the dag block
       db_->saveDagBlock(blk);
+
+      // TODO: This is an ugly temporary fix for testnet, a better solution is needed for dag block race condition
+      if (db_->getDagBlockPeriod(blk.getHash()) != nullptr) {
+        db_->removeDagBlock(blk.getHash());
+        LOG(log_er_) << "Block already in DB: " << blk.getHash();
+        return;
+      }
     }
     auto blk_hash = blk.getHash();
     auto pivot_hash = blk.getPivot();
@@ -434,7 +444,7 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
     if (save) {
       block_verified_.emit(blk);
       if (auto net = network_.lock()) {
-        net->onNewBlockVerified(blk, proposed);
+        net->onNewBlockVerified(blk, proposed, std::move(trxs));
       }
     }
   }
@@ -605,6 +615,25 @@ void DagManager::recoverDag() {
 
   for (auto &lvl : db_->getNonfinalizedDagBlocks()) {
     for (auto &blk : lvl.second) {
+      // These are some sanity checks that difficulty is correct and block is truly non-finalized.
+      // This is only done on startup
+      auto period = db_->getDagBlockPeriod(blk.getHash());
+      if (period != nullptr) {
+        LOG(log_er_) << "Nonfinalized Dag Block actually finalized in period " << period->first;
+        break;
+      } else {
+        auto propose_period = dag_blk_mgr_->getProposalPeriod(blk.getLevel());
+        // Verify VDF solution
+        try {
+          blk.verifyVdf(dag_blk_mgr_->sortitionParamsManager().getSortitionParams(propose_period.first));
+        } catch (vdf_sortition::VdfSortition::InvalidVdfSortition const &e) {
+          LOG(log_er_) << "DAG block " << blk.getHash() << " with " << blk.getLevel()
+                       << " level failed on VDF verification with pivot hash " << blk.getPivot() << " reason "
+                       << e.what();
+          break;
+        }
+      }
+
       addDagBlock(std::move(blk), {}, false, false);
     }
   }
--- a/libraries/core_libs/consensus/src/dag/dag.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag.cpp
@@ -611,9 +611,9 @@ void DagManager::recoverDag() {
   trx_mgr_->recoverNonfinalizedTransactions();
 }
 
-const std::map<uint64_t, std::vector<blk_hash_t>> DagManager::getNonFinalizedBlocks() const {
+const std::pair<uint64_t, std::map<uint64_t, std::vector<blk_hash_t>>> DagManager::getNonFinalizedBlocks() const {
   SharedLock lock(mutex_);
-  return non_finalized_blks_;
+  return {period_, non_finalized_blks_};
 }
 
 std::pair<size_t, size_t> DagManager::getNonFinalizedBlocksSize() const {
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
@@ -20,7 +20,7 @@ class GetDagSyncPacketHandler : public PacketHandler {
 
   virtual ~GetDagSyncPacketHandler() = default;
 
-  void sendBlocks(dev::p2p::NodeID const& peer_id, std::vector<std::shared_ptr<DagBlock>> blocks,
+  void sendBlocks(const dev::p2p::NodeID& peer_id, std::vector<std::shared_ptr<DagBlock>>&& blocks,
                   uint64_t request_period, uint64_t period);
 
  private:
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
@@ -20,7 +20,8 @@ class GetDagSyncPacketHandler : public PacketHandler {
 
   virtual ~GetDagSyncPacketHandler() = default;
 
-  void sendBlocks(dev::p2p::NodeID const& peer_id, std::vector<std::shared_ptr<DagBlock>> blocks);
+  void sendBlocks(dev::p2p::NodeID const& peer_id, std::vector<std::shared_ptr<DagBlock>> blocks,
+                  uint64_t request_period, uint64_t period);
 
  private:
   void process(const PacketData& packet_data, const std::shared_ptr<TaraxaPeer>& peer) override;
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/pbft_block_packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/pbft_block_packet_handler.hpp
@@ -6,14 +6,18 @@ namespace taraxa {
 class PbftBlock;
 class PbftChain;
 class PbftManager;
+class DagBlockManager;
 }  // namespace taraxa
 
 namespace taraxa::network::tarcap {
 
+class SyncingState;
+
 class PbftBlockPacketHandler : public PacketHandler {
  public:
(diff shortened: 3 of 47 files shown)


libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -611,9 +611,9 @@ void DagManager::recoverDag() {
   trx_mgr_->recoverNonfinalizedTransactions();
 }
 
-const std::map<uint64_t, std::vector<blk_hash_t>> DagManager::getNonFinalizedBlocks() const {
+const std::pair<uint64_t, std::map<uint64_t, std::vector<blk_hash_t>>> DagManager::getNonFinalizedBlocks() const {
   SharedLock lock(mutex_);
-  return non_finalized_blks_;
+  return {period_, non_finalized_blks_};
```
gaoqi8649, line 616: Non finalized DAG blocks should have no period yet. What is the period number?

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
```diff
@@ -46,23 +41,28 @@ void DagBlockPacketHandler::process(const PacketData &packet_data, const std::sh
...
-      status.second.insert(hash);
-      requestDagBlocks(packet_data.from_node_id_, status.second, DagSyncRequestType::MissingHashes);
-      return;
+      if (peer->peer_dag_syncing) {
+        LOG(log_dg_) << "Ignore new dag block " << hash.abridged() << ", dag syncing is on";
+      } else {
+        if (peer->peer_dag_synced) {
+          LOG(log_er_) << "DagBlock" << block.getHash() << " has missing pivot or/and tips " << status.second
+                       << " . Peer " << packet_data.from_node_id_.abridged() << " will be disconnected.";
+          disconnect(peer->getId(), dev::p2p::UserReason);
```
gaoqi8649, line 56: I think here is dangerous. DAG is growing constantly, that's very hard to determined peer DAG has synced or not

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_block_packet_handler.cpp
```diff
@@ -28,6 +33,19 @@ void PbftBlockPacketHandler::process(const PacketData &packet_data, const std::s
     peer->pbft_chain_size_ = peer_pbft_chain_size;
   }
 
+  // Ignore any pbft block that is missing dag anchor
```
gaoqi8649, line 36: If this is a good PBFT proposal block, just DAG has not synced yet. We should not drop the block

gaoqi8649:
After carefully thinking of this, I have several questions:

// PBFT anchor gossiping
PBFT is using votes to determine voting value, not by proposal blocks. If a node receive votes but drop the PBFT proposal block(because of missing DAG blocks). That will still vote on the value until timeout giving up on the value. 
If DAG blocks proposal in a high rate, that will cause there is no enough(2t+1) nodes have the same DAG in time. And that will cause continue to propose PBFT blocks and never finalize blocks. 
Since the main bug is DAG blocks syncing problem for missing DAG blocks issue. The change just hidden the bug in PBFT side, not fix the bug. If missing DAG blocks issue still there, PBFT still cannot finalize blocks, and instead of proposing new blocks constantly. If the missing DAG blocks bug get fixed, we don’t need change anything here without consider malicious players.

So I think this is only working for protect from malicious players. I think should be no help on DAG syncing or PBFT finalization. 
And how to check the PBFT block sender is malicious players and not because of DAG has not synced yet?

// PBFT syncing. 
This is OK for Testnet, but I think that’s not good for Mainnet. Actually I have a completely opposite view. I think we should avoid sending any syncing request by state messages because state messages cannot verify, malicious players could send out state message with any number.

// DAG syncing
DAG is growing constantly. How to determine a node’s DAG has synced or not?

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: This is the current period we are in. Basically the same values as current pbft chain size

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_block_packet_handler.cpp
mfrankovi: I have modified this

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -105,49 +106,52 @@ void StatusPacketHandler::process(const PacketData& packet_data, const std::shar
...
-    selected_peer->pbft_round_ = (*it++).toPositiveInt64();
+    selected_peer->pbft_round_ = (*it++).toInt<uint64_t>();
     selected_peer->pbft_previous_round_next_votes_size_ = (*it++).toInt<unsigned>();
 
+    // TODO: Address malicious status
+    if (!syncing_state_->is_pbft_syncing()) {
+      if (pbft_synced_period < selected_peer->pbft_chain_size_) {
+        LOG(log_nf_) << "Restart PBFT chain syncing. Own synced PBFT at period " << pbft_synced_period
+                     << ", peer PBFT chain size " << selected_peer->pbft_chain_size_;
+        if (pbft_synced_period < selected_peer->pbft_chain_size_ + 1) {
```
gaoqi8649, line 120: Dose this should be "(pbft_synced_period + 1) < selected_peer->pbft_chain_size_"?

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
mfrankovi: This flag indicates that we at least once did pull all the non-finalized dag blocks from the node sending us this message. If this has already happened and if there is a missing tip/pivot node is disconnected since this should not be possible with honest node

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
mfrankovi: Yes. Good catch. I fixed it

libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
```diff
@@ -20,7 +20,8 @@ class GetDagSyncPacketHandler : public PacketHandler {
 
   virtual ~GetDagSyncPacketHandler() = default;
 
-  void sendBlocks(dev::p2p::NodeID const& peer_id, std::vector<std::shared_ptr<DagBlock>> blocks);
+  void sendBlocks(dev::p2p::NodeID const& peer_id, std::vector<std::shared_ptr<DagBlock>> blocks,
```
JakubFornadel, line 23: std::vector<std::shared_ptr<DagBlock>> should be probably `const ref` - a lot of data has to be copied here

libraries/core_libs/network/include/network/tarcap/packets_handlers/pbft_block_packet_handler.hpp
```diff
@@ -6,10 +6,13 @@ namespace taraxa {
 class PbftBlock;
 class PbftChain;
 class PbftManager;
+class DagBlockManager;
 }  // namespace taraxa
 
 namespace taraxa::network::tarcap {
 
+class SyncingState;
```
JakubFornadel, line 14: why are these forwards declarations in this header when they are not used here ?

libraries/core_libs/network/include/network/tarcap/taraxa_capability.hpp
```diff
@@ -83,7 +83,8 @@ class TaraxaCapability : public dev::p2p::CapabilityFace {
 
   // METHODS USED IN TESTS ONLY
   void sendBlock(dev::p2p::NodeID const &id, DagBlock const &blk);
-  void sendBlocks(dev::p2p::NodeID const &id, std::vector<std::shared_ptr<DagBlock>> blocks);
+  void sendBlocks(dev::p2p::NodeID const &id, std::vector<std::shared_ptr<DagBlock>> blocks, uint64_t request_period,
```
JakubFornadel, line 86: also here `const ref` for blocks

libraries/core_libs/network/include/network/tarcap/taraxa_peer.hpp
```diff
@@ -70,6 +70,12 @@ class TaraxaPeer : public boost::noncopyable {
   std::atomic<uint64_t> pbft_chain_size_ = 0;
   std::atomic<uint64_t> pbft_round_ = 1;
   std::atomic<size_t> pbft_previous_round_next_votes_size_ = 0;
+  std::atomic<uint64_t> last_status_pbft_chain_size_ = 0;
+  std::atomic_bool peer_dag_synced = false;
+  std::atomic_bool peer_dag_syncing = false;
```
JakubFornadel, line 75: missing underscore at the end

libraries/core_libs/network/include/network/tarcap/threadpool/packets_blocking_mask.hpp
```diff
@@ -6,6 +6,8 @@
 #include <map>
 #include <shared_mutex>
 
+#include "config/config.hpp"
+#include "logger/logger.hpp"
```
JakubFornadel, line 10: why were these included added when not used here ?

libraries/core_libs/network/src/network.cpp
```diff
@@ -128,9 +128,10 @@ void Network::sendBlock(dev::p2p::NodeID const &id, DagBlock const &blk) {
   LOG(log_dg_) << "Sent Block:" << blk.getHash().toString();
 }
 
-void Network::sendBlocks(dev::p2p::NodeID const &id, std::vector<std::shared_ptr<DagBlock>> blocks) {
+void Network::sendBlocks(dev::p2p::NodeID const &id, std::vector<std::shared_ptr<DagBlock>> blocks,
```
JakubFornadel, line 131: hmm I see that we have a lot of chaos in all these `sendBlocks` methods :D So we have 2 options:

- either use std::vector<std::shared_ptr<DagBlock>>&&
- or const std::vector<std::shared_ptr<DagBlock>>& 

depends what is more appropriate

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
```diff
@@ -46,23 +41,28 @@ void DagBlockPacketHandler::process(const PacketData &packet_data, const std::sh
     }
 
     if (auto status = checkDagBlockValidation(block); !status.first) {
-      // Ignore new block packets when pbft syncing
+      // Ignore new block packets when pbft syncing or dag syncing
```
JakubFornadel, line 44: either the comment is wrong or is_pbft_syncing() name is wrong here

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
```diff
@@ -55,12 +79,10 @@ void DagSyncPacketHandler::process(const PacketData& packet_data, const std::sha
     dag_blk_mgr_->insertBroadcastedBlock(std::move(block));
   }
 
-  if (missing_blks.size() > 0) {
-    requestDagBlocks(packet_data.from_node_id_, missing_blks, DagSyncRequestType::MissingHashes);
-  }
-  syncing_state_->set_dag_syncing(false);
+  peer->peer_dag_synced = true;
+  peer->peer_dag_syncing = false;
```
JakubFornadel, line 83: so as dag_syncing is moved out from syncing_state, once we are pbft synced (but not yet dag synced), received dag blocks from other peers will probably fail because we are not dag synced yet, is that right and are we ok with it ?

JakubFornadel: + didnt you changed testnet that we are sending dag blocks in batches of 100 ? In such case this would not work properly ?

JakubFornadel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
mfrankovi: The batch of 100 is removed from testnet as well. This was there only because I originally did not realize what was causing the slow processing. And yes we are ok with received dag blocks failing with tips/pivot missing until dag_synced is true, once dag_synced is true for a peer any missing tip/pivot should indicate a malicious node. Currently I disconnect such a node but we should have some blacklisting as well not to reconnect to it again

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -414,6 +417,13 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
       trx_mgr_->saveTransactionsFromDagBlock(trxs);
       // Save the dag block
       db_->saveDagBlock(blk);
+
+      // TODO: This is an ugly temporary fix for testnet, a better solution is needed for dag block race condition
+      if (db_->getDagBlockPeriod(blk.getHash()) != nullptr) {
+        db_->removeDagBlock(blk.getHash());
+        LOG(log_er_) << "Block already in DB: " << blk.getHash();
+        return;
+      }
```
gaoqi8649, line 426: I am not sure if you will fix it in a better way, or just leave it as this

gaoqi8649 (approved):

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: I will fix this but I will need another PR for this since the solution is not simple

PR #1346, opened 2021-12-15
fix: syncing stall

mfrankovi:
This condition happens often if one block behind and at the same time a pbft block is pushed by pbft consensus and syncing at the same time. Unless restartSyincingPbft is called, node gets stuck in syncing state.


justinsnapp (approved):

justinsnapp:
This is essentially an "off by one" type issue where we are close-in to being synced but have fallen behind due to a block just having been pushed (that we weren't in sync for consensus for),  correct?

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -73,6 +73,7 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
   if (sync_block.pbft_blk->getPeriod() != pbft_mgr_->pbftSyncingPeriod() + 1) {
     LOG(log_dg_) << "Block " << pbft_blk_hash << " period unexpected: " << sync_block.pbft_blk->getPeriod()
                  << ". Expected period: " << pbft_mgr_->pbftSyncingPeriod() + 1;
+    restartSyncingPbft(true);
```
MatusKysel, line 76: don't we need to disable syncing by `syncing_state_->set_pbft_syncing(false);`

mfrankovi: No, since restartsyncing with force flag will set it to true or false inside

MatusKysel (approved):

PR #1348, opened 2021-12-16
Rewards only for dags

JakubFornadel:
## Purpose

Distribute rewards based on unique transactions included in dag blocks. This is version without rewarding votes as that might complicate things and we want to have something easy & stable asap. 

## For reviewers

taraxa-evm go changes can be found in this separate PR: https://github.com/Taraxa-project/taraxa-evm/pull/21/files

Note: Go part is using old eth model for generating rewards and all rewards are assigned to the pbft block author so existing state_api tests pass. Code and internal structures are adjusted and saved in map `address -> rewards` -> prepared for writing into the mapped storage of smartcontract. Once we have this smartcontract and interface to it, it should be fairly easy and quick to use current go changes to work with the contract...

Note 2: do not review doc -> it will be changed later...


libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -76,45 +77,72 @@ class FinalChainImpl final : public FinalChain {
...
-      SyncBlock sync_block(period_raw);
-      DB::MultiGetQuery db_query(db_, sync_block.transactions.size());
-      for (auto const& trx : sync_block.transactions) {
-        db_query.append(DB::Columns::final_chain_transaction_location_by_hash, trx.getHash());
+
+    // Create dag & transactions stats for rewards distribution
+    RewardsStats rewards_stats(sync_block.getTransactions().size(), sync_block.getDagBlocks().size());
+    for (const auto& dag_block : sync_block.getDagBlocks()) {
+      // TODO: this assert throws in test FinalChainTest.contract, either fix the test or delete the assert
+      // assert(dag_block.getTrxs().size());
```
gaoqi8649, line 97: I think should keep the assert here. DAG block with no transactions should be an invalid block, and need to drop

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1460,16 +1477,19 @@ bool PbftManager::pushCertVotedPbftBlockIntoChain_(taraxa::blk_hash_t const &cer
...
+  if (!comparePbftBlockScheduleWithDAGblocks_(pbft_block)) {
     LOG(log_nf_) << "DAG has not build up for PBFT block " << cert_voted_block_hash;
     return false;
   }
-  cert_sync_block_.cert_votes = std::move(cert_votes_for_round);
-  if (!pushPbftBlock_(cert_sync_block_, *dag_blocks_order)) {
+  cert_sync_block_.setCertVotes(std::move(cert_votes_for_round));
+
+  // TODO: check if cert_sync_block_ has to be created again inside comparePbftBlockScheduleWithDAGblocks_ even
+  //  after pushPbftBlock. If yes, shared_ptr to const SyncBlock should be probably used ?
```
gaoqi8649, line 1487: This should never happen. PBFT manager is serialize, not parallel

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -47,35 +47,37 @@ void VoteManager::retreieveVotes_() {
...
-bool VoteManager::addUnverifiedVote(std::shared_ptr<Vote> const& vote) {
+bool VoteManager::addUnverifiedVote(std::shared_ptr<Vote>&& vote) {
   uint64_t pbft_round = vote->getRound();
   const auto& hash = vote->getHash();
   vote->getVoterAddr();  // this will cache object variables - speed up
 
+  if (voteInUnverifiedMap(pbft_round, hash)) {
+    LOG(log_dg_) << "Vote " << hash << " is already in unverified map";
+    return false;
+  }
```
gaoqi8649, line 58: Why need check voteInUnverifiedMap first? What's the benefits?

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -33,18 +33,27 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
...
-  if (vote_mgr_->voteInVerifiedMap(vote) || !vote_mgr_->addUnverifiedVote(vote)) {
-    LOG(log_dg_) << "Received PBFT vote " << vote_hash << " (from " << packet_data.from_node_id_.abridged()
-                 << ") already saved in (un)verified queues.";
+  // No need to check unverified map as it is checked inside addUnverifiedVote
+  if (vote_mgr_->voteInVerifiedMap(vote)) {
+    LOG(log_dg_) << "Vote " << vote_hash << " (from " << packet_data.from_node_id_.abridged()
+                 << ") already saved in verified queue";
+    return;
+  }
+  auto vote_ptr_copy = vote;
```
gaoqi8649, line 42: I think could remove the "vote_ptr_copy", and remove "std::move" in "if (!vote_mgr_->addUnverifiedVote(std::move(vote)))"

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
```diff
@@ -61,16 +61,17 @@ void VotesSyncPacketHandler::process(const PacketData &packet_data, const std::s
                      << ") already saved in queue.";
         continue;
       }
+      auto vote_ptr_copy = vote_n;
```
gaoqi8649, line 64: same here. remove "vote_ptr_copy", and remove "std::move" in "if (!vote_mgr_->addUnverifiedVote(std::move(vote_n)))"

libraries/core_libs/consensus/src/final_chain/final_chain.cpp
JakubFornadel: When I think about it now, we need such condition/or assert inside proposing/receiving dag blocs and not here... This is just finalize() function that is processing rewards so at this moment such invalid dag bock is in database, dag, everywhere...

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
JakubFornadel: It might happen quite often that you receive the same vote from multiple nodes and because we are processing votes concurrently, you want to eliminate "unique" locking as much as possible because it blocks the other threads. By checking voteInUnverifiedMap first, you use shared_lock, which does not block other threads - this might improve performance a lot

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
JakubFornadel: no you cant, addUnverifiedVote has an argument of type `std::shared_ptr<Vote>&&`, which means you gotta use std::move

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
JakubFornadel: same answer

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
gaoqi8649: I think the purpose of using std::move is for saving space for passing parameters. Here in order to use std::move, have to make a copy first. 
If like that, I think we should use reference "std::shared_ptr<Vote>&" that should be better

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1485,16 +1516,22 @@ bool PbftManager::pushCertVotedPbftBlockIntoChain_(taraxa::blk_hash_t const &cer
...
     return false;
   }
-  cert_sync_block_.cert_votes = std::move(cert_votes_for_round);
-  if (!pushPbftBlock_(cert_sync_block_, *dag_blocks_order)) {
+  cert_sync_block_.setCertVotes(std::move(cert_votes_for_round));
+
+  // Pbft block from cert_sync_block_ was already processed and saved in db
+  if (pbftSyncBlockProcessed(cert_sync_block_)) {
+    return false;
+  }
```
gaoqi8649, line 1528: We have the checking inside pushPbftBlock_ already.

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1504,21 +1541,26 @@ void PbftManager::pushSyncedPbftBlocksIntoChain_() {
...
-      auto pbft_block_hash = sync_block.pbft_blk->getBlockHash();
+      auto sync_block = processSyncBlock();
+      if (!sync_block.has_value()) {
+        continue;
+      }
+
+      // Pbft block from sync_block was already processed and saved in db
+      if (pbftSyncBlockProcessed(sync_block.value())) {
+        continue;
+      }
```
gaoqi8649, line 1552: Same thing here. In processSyncBlock(), will check if have added the block in PBFT chain
Don't need add pbftSyncBlockProcessed function

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
JakubFornadel: First of all, when you comment on something and request change, you should understand the topic you write about. Move semantics does not really save space for passing parameters, it's purpose is completely different. Plus you are missing big picture here. I made argument of `addUnverifiedVote` method `Vote&&` so I would optimize `addUnverifiedVotes` function, that is important in terms of performance as it pushes hundreds of votes and does it regurarly inside pbft_manager. It is ok that we make copies in packet handlers as long as the performance in some more critical code(pbft manager) is optimized thanks to it... There are always other ways how to do it, but you are clearly not suggesting a new way, just commenting on one line without considering the bigger picture. I guess I can consider this conversation finished

PR #1349, opened 2021-12-16
fix: stale blocks

mfrankovi:
Do not propose stale blocks if any new DAG block appeared in the network

### DIFF
--- a/libraries/core_libs/consensus/include/pbft/block_proposer.hpp
+++ b/libraries/core_libs/consensus/include/pbft/block_proposer.hpp
@@ -46,13 +46,15 @@ class SortitionPropose : public ProposeModelFace {
       : dag_mgr_(std::move(dag_mgr)), dag_blk_mgr_(std::move(dag_blk_mgr)), trx_mgr_(std::move(trx_mgr)) {
     LOG_OBJECTS_CREATE("PR_MDL");
     LOG(log_nf_) << "Set sortition DAG block proposal" << dag_blk_mgr_->sortitionParamsManager().getSortitionParams();
+    // Add a random component in proposing stale blocks so that not all nodes propose stale blocks at the same time
+    max_num_tries_ += (node_addr_[0] % max_num_tries_);
   }
   ~SortitionPropose() {}
   bool propose() override;
 
  private:
   int num_tries_ = 0;
-  const int max_num_tries_ = 20;  // Wait 2000(ms)
+  int max_num_tries_ = 20;  // Wait 2000(ms)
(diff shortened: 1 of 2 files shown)


MatusKysel (approved):

libraries/core_libs/consensus/include/pbft/block_proposer.hpp
```diff
@@ -46,13 +46,15 @@ class SortitionPropose : public ProposeModelFace {
...
+    // Add a random component in proposing stale blocks so that not all nodes propose stale blocks at the same time
+    max_num_tries_ += (node_addr_[0] % max_num_tries_);
   }
   ~SortitionPropose() {}
   bool propose() override;
 
  private:
   int num_tries_ = 0;
-  const int max_num_tries_ = 20;  // Wait 2000(ms)
+  int max_num_tries_ = 20;  // Wait 2000(ms)
```
gaoqi8649, line 57: The comment of wait 2000ms is not true anymore. Since in previous approach for each DAG block proposal, wait 100ms constantly, 20 times waiting that's 2000ms

gaoqi8649 (commented):
In my opinion, I don't like the idea. We introduce sortition params in order to control the proposing rate. If nodes find the DAG proposing rate is too high, should use sortition to chose a reasonable % proposing rate. There should not happen all of nodes are in stale for VDF and no DAG blocks proposing at the moment. So I think the correct fix should adjust sortition, not like this.

mfrankovi:
> In my opinion, I don't like the idea. We introduce sortition params in order to control the proposing rate. If nodes find the DAG proposing rate is too high, should use sortition to chose a reasonable % proposing rate. There should not happen all of nodes are in stale for VDF and no DAG blocks proposing at the moment. So I think the correct fix should adjust sortition, not like this.

But this is basically fixing a bug. The idea is that stale block should only be produced if there is no other blocks being produced. Currently if tips blocks are produced we still keep producing stale blocks. I think this is a bug and needs to be fixed.

libraries/core_libs/consensus/include/pbft/block_proposer.hpp
mfrankovi: This should be fixed differently on testnet but I still think 2000ms gives a proper reference to what 20 means so the comment is good. The variable is changed later with another random component and it will be between 2000 and 4000 but at this point it does apply to 2000ms

gaoqi8649 (approved):

justinsnapp (approved):

PR #1350, opened 2021-12-17
chore: check stale block proposal by frontier

mfrankovi:
Use frontier(pivot and tips) for stale block selection. Using proposal period as before, stale block would be computed even if new blocks appear on the network.

### DIFF
--- a/libraries/types/dag_block/include/dag/dag_block.hpp
+++ b/libraries/types/dag_block/include/dag/dag_block.hpp
@@ -97,21 +97,11 @@ struct DagFrontier {
   }
 
   bool isEqual(const DagFrontier &frontier) {
-    if (pivot == frontier.pivot && tips.size() == frontier.tips.size()) {
-      bool tips_changed = false;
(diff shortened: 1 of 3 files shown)


justinsnapp (approved):

libraries/types/dag_block/include/dag/dag_block.hpp
```diff
@@ -95,6 +95,25 @@ struct DagFrontier {
...
+      }
+      if (tips_changed) {
+        return false;
+      }
+    } else {
+      return false;
+    }
+    return true;
+  }
+
```
kstdl, line 116: Could be simplified to
``` cpp
bool isEqual(const DagFrontier &frontier) {
    if (pivot != frontier.pivot) {
        return false;
    }
    
    return std::equal(tips.begin(), tips.end(), frontier.tips.begin(), frontier.tips.end());
}
```

mfrankovi: Done

kstdl (approved):

PR #1351, opened 2021-12-18
Stale blocks fix

mfrankovi:
Update minor version to force users to update version
Add larger random delay before creating stale blocks
Add a small delay after calculating vdf for a stale block to allow any unprocessed DAG blocks to be added in the DAG


libraries/core_libs/consensus/src/pbft/block_proposer.cpp
```diff
@@ -54,14 +54,21 @@ bool SortitionPropose::propose() {
 
   vdf.computeVdfSolution(sortition_params, frontier.pivot.asBytes());
   if (vdf.isStale(sortition_params)) {
+    // Computing VDF for a stale block is CPU extensive, there is a possibility that some dag blocks are in a queue,
```
kstdl, line 57: Will block with stale difficulty be produced only after some time?
If so that second is useless because we can receive block right after this second. This is kind of random process and you cant know when will receive a block

mfrankovi: So computeVdfSolution takes 100% CPU. I see that this sometimes blocks/delays processing of dag blocks and transactions which are received over the network while computeVdfSolution is running . Delay after this allows the blocks to be processed and add to the DAG. So block reaches our node before this delay but it is not processed and verified in time because of this CPU intensive calculation

kstdl: Got it now

kstdl (approved):

gaoqi8649 (approved):

JakubFornadel (approved):

PR #1362, opened 2021-12-22
Exponential backoff testnet

gaoqi8649:
Hi @justinsnapp, as your request, I cherry pick the PBFT exponential backoff and generate PR for Testnet branch.
In our case, for example like one node CN4 down many hours and cause PBFT stalled. Then restart CN4 and want to catch up with others nodes steps. This cannot help, and may need take longer time.

The benefits: could save space for many votes when PBFT stalled
The disadvantage: when more than 1/3 weighted nodes down, and restart nodes want to catch up. That may take longer time because votes not send often as before (in old approach, send out 2 votes every 2 seconds for each node)

### DIFF
--- a/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
+++ b/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
@@ -169,6 +169,7 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
   u_long const LAMBDA_ms_MIN;
   u_long LAMBDA_ms = 0;
   u_long LAMBDA_backoff_multiple = 1;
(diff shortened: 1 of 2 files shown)


mfrankovi:
Does this need to be in a fork or it will work without a fork on testnet?

gaoqi8649:
> Does this need to be in a fork or it will work without a fork on testnet?

This will work without a fork, but need to increase node version

JakubFornadel:
lets merge this into the develop ? What do you think @gaoqi8649 ?

gaoqi8649:
> lets merge this into the develop ? What do you think @gaoqi8649 ?

Develop branch has the change already. This is for testnet branch. Yeah, let's merge it

PR #1363, opened 2021-12-23
feat: add hardfork2. excluding stale blocks from efficiency calculation

kstdl:


justinsnapp (approved):

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -107,7 +113,28 @@ SortitionParams SortitionParamsManager::getSortitionParams(std::optional<uint64_
...
+    }
+    return s + b.getTrxs().size();
+  };
+  size_t total_count = std::accumulate(block.dag_blocks.begin(), block.dag_blocks.end(), 0, accLambda);
+  uint16_t res = 100 * kOnePercent;
+  if (total_count != 0) {
+    res = block.transactions.size() * 100 * kOnePercent / total_count;
+  }
+  // this needed because stale blocks could have efficient transactions and efficiency would be bigger than 100%
+  res = std::min(uint16_t(100 * kOnePercent), res);
```
JakubFornadel, line 129: I guess it is by design that number 10000 represents 100% ?

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -107,7 +113,28 @@ SortitionParams SortitionParamsManager::getSortitionParams(std::optional<uint64_
...
+  auto accLambda = [&stale_difficulty](uint32_t s, const DagBlock& b) -> uint32_t {
+    if (b.getDifficulty() == stale_difficulty) {
+      return s;
+    }
+    return s + b.getTrxs().size();
+  };
+  size_t total_count = std::accumulate(block.dag_blocks.begin(), block.dag_blocks.end(), 0, accLambda);
+  uint16_t res = 100 * kOnePercent;
+  if (total_count != 0) {
+    res = block.transactions.size() * 100 * kOnePercent / total_count;
```
JakubFornadel, line 126: not 100% sure if thats what we really want if we want to ignore stale blocks and everything that is inside, I think you should iterate over the dag blocks in for loop, check if block is stale or not (as you do in lambda) and count only unique txs from non-stale blocks instead of `block.transactions.size()`

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
kstdl: Yes. It is not a good idea to use float/double for such calculations. So I defining one percent as 100. I guess that precision is enough for that calculations.

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
justinsnapp: Yes it seems like we exclude stale blocks only from the denominator, but not the numerator?   Which is why we have to check that we dont exceed 100%.   But why not exclude blocks from both denominator and numerator as @JakubFornadel  says?

kstdl: Yeah, that makes sense. But to check if that transaction is unique we should check it in the database, because transaction could be finalized in previous pbft blocks

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -1,8 +1,11 @@
 #include "dag/sortition_params_manager.hpp"
 
 #include "pbft/pbft_block.hpp"
+
 namespace taraxa {
 
+constexpr uint16_t k_testnet_hardfork2_sortition_interval_length = 200;
```
JakubFornadel, line 7: I am pretty confused and the wording seems weird. 

- This constant is basically computation_interval (200) and the original computation interval should be renamed to something like changing_interval(50) ? 
- Why is this not part of config ? If it was part of config, you would not need to use hardfork2 if conditions [here](https://github.com/Taraxa-project/taraxa-node/pull/1363/files#diff-740ffa012af5a6c39db38b0cba5ee86455365dd3397a774838fd38b043e5a8c6R157) and [here](https://github.com/Taraxa-project/taraxa-node/pull/1363/files#diff-740ffa012af5a6c39db38b0cba5ee86455365dd3397a774838fd38b043e5a8c6R80)

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -126,13 +153,22 @@ uint16_t SortitionParamsManager::averageCorrectionPerPercent() const {
 }
 
 void SortitionParamsManager::cleanup(uint64_t current_period) {
-  dag_efficiencies_.clear();
+  uint16_t efficiencies_to_leave = 0;
+  if (current_period >= k_testnet_hardfork2_block_num) {
+    efficiencies_to_leave = k_testnet_hardfork2_sortition_interval_length - config_.computation_interval;
```
JakubFornadel, line 158: as mentioned above, if I understand it correctly, config_.computation_interval is now changing_interval and is computation_interval - confusing wording here

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
kstdl: Now excluding from both, but not checking if transaction was finalised in previous pbft blocks

JakubFornadel: hmm thats going to be pretty expensive but yeah I guess to be sure if tx is unique or not we need to also know if it wasnt included in some previous pbft block

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
kstdl: Changing the config structure(introducing a new field, or even change a name) will create and additional task how to deal with that config change for node runners. 
For the develop branch it will be sure in the config, but for the testnet with a hardfork I am not sure we should spend time on that

JakubFornadel: well we are getting into the situation where most of the lateest changes in testnet are like "lets merge this and do it properly in develop", but does anybody even keep track of it ? It will be pretty big mess if we continue do it like this. If authors keep track of it I guess it might be ok but I am not sure if it is the case

mfrankovi (approved):

PR #1372, opened 2021-12-27
Set pbft step

mfrankovi:
Override pbft step by cmd line option


JakubFornadel (approved):

gaoqi8649:
I think this is not help for our nodes running on cloud, because this need a reboot and setting step value on command line.
There is setPbftStep in PBFT manager, I am not sure if we could create a RPC call for that.

JakubFornadel:
@mfrankovi can we close this PR ?

mfrankovi:
> @mfrankovi can we close this PR ?

@justinsnapp do we still want to have this?

gaoqi8649:
I think for testing purpose, that's better in RPC call instead of CLI

mfrankovi:
I am closing this. Lets keep the branch alive a bit more if we ever decide to use it

PR #1376, opened 2021-12-28
udp port for boot node config

mfrankovi:


rattrap (approved):

JakubFornadel (changes requested):
- lets delete also tcp ports from devnet/testnet configs l
- ibraries/cli/src/tools.cpp has still some references to  "tcp_port" 
- resolveHost function inside TaraxaCapability::initBootNodes seems like it is complicated with no valid reason. bootnode main parses it in like 5 lines compared to what we have

PR #1381, opened 2021-12-28
Difficulty check hf

kstdl:


libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -152,12 +152,16 @@ void SortitionParamsManager::cleanup(uint64_t current_period) {
...
-  }
+  auto hardfork = [&](uint64_t block_num, const VrfParams& vrf) {
+    if (block.pbft_blk->getPeriod() == block_num) {
+      SortitionParamsChange pc{block_num, 0, vrf};
+      params_changes_.push_back(pc);
+      db_->saveSortitionParamsChange(block_num, pc, batch);
+    }
+  };
+  hardfork(3630, {0x5000, 0x1300});
+  hardfork(5250, {0x1770, 0xbb8});
```
JakubFornadel, line 163: Explorer shows some weird state, based on taraxa-nodes I can see we stoped at pbft period 5275

JakubFornadel: just force pushed it

PR #1393, opened 2021-12-31
chore(hardfork): multiply genesis_balances by 1e18

kstdl:


justinsnapp (approved):

gaoqi8649 (commented):
I don't understand the unit test. Why is at the 9000 final block that balance change to 10^18?
I checked the eth_mainnet_blocks_0_300000.rlp is a binary file, also don't know what does that do?
I am not familiar with EVM.

justinsnapp (changes requested):
Doesn’t seem to pass the unit test.

kstdl:
@justinsnapp Thats because I changed hardfork block number to 12000,but didnt changed it in the test.

kstdl:
> I don't understand the unit test. Why is at the 9000 final block that balance change to 10^18?
> 
> I checked the eth_mainnet_blocks_0_300000.rlp is a binary file, also don't know what does that do?
> 
> I am not familiar with EVM.

That file is not related to this test. I just forget to remove it when I was changing the test logic. I will clean it a little bit later

JakubFornadel (changes requested):
Ok so we launched mainnet candidate and I think we all agreed that since that event we are going to follow certain rules related to [github practices](https://github.com/Taraxa-project/taraxa-node/blob/develop/doc/git_practices.md) and versioning - TODO: we should also create som document and agreed on it. 

- it is proper hardfork, it should have a name ! We could base naming on ethereum -> [read here](https://eips.ethereum.org/), but we can call it TIP (Taraxa Improvement Proposal) instead.  We will probably use similar multi-step approval process for hardforks in the future where community is somehow involved...
- hardfork block should not be just hardcoded somewhere in the evm like it is now. It should be somewhere in config class, something like `tip_XYZ_block` and then passed passed to the evm. You choose to support hardfork or not by running a certain version. Version is set in C++ code so no hardfork block num can be hardcoded in evm ! It will be hardcoded in C++ by setting this value in C++ class config probably ?
- we definitely need to change node version ! But version to use ? TODO: create doc for versioning rules
- this branch should be called hotfix/* - I would also include new version into the name of branch

PR #1395, opened 2022-01-10
Improved discovery

MatusKysel:
So idea is that node can report different IP (external IP), that should be use for discovery. I implemented basic mappings that will keep track of node ID and it's real ip (this could change) and real IP with external IP, so we know what should we report to the rest of the network

### DIFF
--- a/libraries/config/src/config.cpp
+++ b/libraries/config/src/config.cpp
@@ -23,10 +23,19 @@ Json::Value getConfigData(Json::Value root, std::vector<string> const &path, boo
   return root;
 }
 
-std::string getConfigDataAsString(Json::Value const &root, std::vector<string> const &path) {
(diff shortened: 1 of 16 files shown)


libraries/config/src/config.cpp
```diff
@@ -23,10 +23,19 @@ Json::Value getConfigData(Json::Value root, std::vector<string> const &path, boo
   return root;
 }
 
-std::string getConfigDataAsString(Json::Value const &root, std::vector<string> const &path) {
+std::string getConfigDataAsString(Json::Value const &root, std::vector<string> const &path, bool optional = false,
+                                  std::string value = {}) {
```
JakubFornadel, line 27: what is this value argument good for ?

MatusKysel: default value, check the uint version

JakubFornadel (approved):

mfrankovi (approved):

PR #1396, opened 2022-01-11
fix: dag sync deadlock

mfrankovi:
On devnet one of the nodes experienced a deadlock which blocked some of the threads and the process ran out of memory.
This was caused by DagManager::mutex_ and TaraxaPeer::mutex_for_sending_dag_blocks_ mutexes which were invoked in reverse order on gossiping the blocks compared to syncing dag blocks.

I have refactored this code that the functionality of both mutexes is now handled by mutex in DagManager and mutex_for_sending_dag_blocks_ is no longer needed. The purpose of mutex in DagManager now is:
1. Protect any multi threaded access to DagManager class members
2. Ensures that dag blocks are added into DAG in correct order
3. Ensures that dag blocks are gossiped in the correct order (pivot/tip is always sent before the block) 
4. Ensures that any dag syncing request is processed and sent out before any new dag block is gossiped - prevent race condition between gossiping and syncing

### DIFF
--- a/libraries/core_libs/consensus/include/dag/dag.hpp
+++ b/libraries/core_libs/consensus/include/dag/dag.hpp
@@ -28,6 +28,9 @@
 #include "pbft/pbft_chain.hpp"
 #include "storage/storage.hpp"
 #include "transaction_manager/transaction_manager.hpp"
+namespace taraxa::network::tarcap {
+class TaraxaPeer;
+}
 namespace taraxa {
 
 /**
@@ -164,6 +167,16 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
   }
 
   const std::pair<uint64_t, std::map<uint64_t, std::vector<blk_hash_t>>> getNonFinalizedBlocks() const;
+
+  /**
+   * @brief Sends non finalized dag blocks excluding the known block to the peer
+   *
+   * @param blocks_hashes Blocks known to the peer
+   * @param peer Peer to send the blocks to
+   * @param peer_period Peer period
+   */
+  void sendNonFinalizedBlocks(std::unordered_set<blk_hash_t> &&blocks_hashes,
+                              const std::shared_ptr<network::tarcap::TaraxaPeer> &peer, uint64_t peer_period);
 
   DagFrontier getDagFrontier();
 
--- a/libraries/core_libs/consensus/src/dag/dag.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag.cpp
@@ -13,6 +13,7 @@
 
 #include "dag/dag.hpp"
 #include "network/network.hpp"
+#include "network/tarcap/taraxa_peer.hpp"
 #include "transaction_manager/transaction_manager.hpp"
 
 namespace taraxa {
@@ -450,6 +451,35 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
   }
   LOG(log_dg_) << " Update frontier after adding block " << blk.getHash() << "anchor " << anchor_
                << " pivot = " << frontier_.pivot << " tips: " << frontier_.tips;
+}
+
+void DagManager::sendNonFinalizedBlocks(std::unordered_set<blk_hash_t> &&blocks_hashes,
+                                        const std::shared_ptr<network::tarcap::TaraxaPeer> &peer,
+                                        uint64_t peer_period) {
+  // Sending of blocks is in DagManager class to use DagManager mutex_ which ensures that there is no race condition
+  // between syncing and gossiping dag blocks
+  SharedLock lock(mutex_);
+  std::vector<std::shared_ptr<DagBlock>> dag_blocks;
+  // There is no point in sending blocks if periods do not match
+  if (peer_period == period_) {
+    peer->syncing_ = false;
+    for (auto &level_blocks : non_finalized_blks_) {
+      for (auto &block : level_blocks.second) {
+        const auto hash = block;
+        if (blocks_hashes.count(hash) == 0) {
+          if (auto blk = dag_blk_mgr_->getDagBlock(hash); blk) {
+            dag_blocks.emplace_back(blk);
+          } else {
+            LOG(log_er_) << "NonFinalizedBlock " << hash << " not in DB";
+            assert(false);
+          }
+        }
+      }
+    }
+  }
+  if (auto net = network_.lock()) {
+    net->sendBlocks(peer->getId(), std::move(dag_blocks), peer_period, period_);
+  }
 }
 
 void DagManager::drawGraph(std::string const &dotfile) const {
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/get_dag_sync_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/get_dag_sync_packet_handler.cpp
@@ -18,13 +18,8 @@ GetDagSyncPacketHandler::GetDagSyncPacketHandler(std::shared_ptr<PeersState> pee
       dag_blk_mgr_(std::move(dag_blk_mgr)),
       db_(std::move(db)) {}
 
-void GetDagSyncPacketHandler::process(const PacketData &packet_data,
-                                      [[maybe_unused]] const std::shared_ptr<TaraxaPeer> &peer) {
-  // This lock prevents race condition between syncing and gossiping dag blocks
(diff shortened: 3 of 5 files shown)


libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -452,6 +453,35 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
...
+  // Sending of blocks is in DagManager class to use DagManager mutex_ which ensures that there is no race condition
+  // between syncing and gossiping dag blocks
+  SharedLock lock(mutex_);
+  std::vector<std::shared_ptr<DagBlock>> dag_blocks;
+  // There is no point in sending blocks if periods do not match
+  if (peer_period == period_) {
+    peer->syncing_ = false;
+    for (auto &level_blocks : non_finalized_blks_) {
+      for (auto &block : level_blocks.second) {
+        const auto hash = block;
```
gaoqi8649, line 468: I think that could remove line 468, if do "for (auto &hash : level_blocks.second)"

gaoqi8649 (approved):

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -452,6 +453,35 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
...
+          } else {
+            LOG(log_er_) << "NonFinalizedBlock " << hash << " not in DB";
+            assert(false);
+          }
+        }
+      }
+    }
+  }
+  if (auto net = network_.lock()) {
+    net->sendBlocks(peer->getId(), std::move(dag_blocks), peer_period, period_);
```
MatusKysel, line 481: I believe this method is used only for testing purposes, maybe better approach would be to return `dag_blocks` and send it from handler

libraries/core_libs/network/src/tarcap/packets_handlers/get_dag_sync_packet_handler.cpp
```diff
@@ -34,25 +29,7 @@ void GetDagSyncPacketHandler::process(const PacketData &packet_data,
...
-          } else {
-            LOG(log_er_) << "NonFinalizedBlock " << hash << " not in DB";
-            assert(false);
-          }
-        }
-      }
-    }
-  }
-  sendBlocks(packet_data.from_node_id_, std::move(dag_blocks), peer_period, period);
+  dag_mgr_->sendNonFinalizedBlocks(std::move(blocks_hashes), std::move(peer), peer_period);
```
MatusKysel, line 32: no move on peer object

libraries/core_libs/consensus/src/dag/dag.cpp
JakubFornadel: const auto &hash

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -452,6 +453,35 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
...
+  // between syncing and gossiping dag blocks
+  SharedLock lock(mutex_);
+  std::vector<std::shared_ptr<DagBlock>> dag_blocks;
+  // There is no point in sending blocks if periods do not match
+  if (peer_period == period_) {
+    peer->syncing_ = false;
+    for (auto &level_blocks : non_finalized_blks_) {
+      for (auto &block : level_blocks.second) {
+        const auto hash = block;
+        if (blocks_hashes.count(hash) == 0) {
```
JakubFornadel, line 469: better readability if this form is used:

```
if (blocks_hashes.count(hash) != 0) {
  continue:
}
```

libraries/core_libs/consensus/include/dag/dag.hpp
```diff
@@ -165,6 +168,16 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
...
 
+  /**
+   * @brief Sends non finalized dag blocks excluding the known block to the peer
+   *
+   * @param blocks_hashes Blocks known to the peer
+   * @param peer Peer to send the blocks to
+   * @param peer_period Peer period
+   */
+  void sendNonFinalizedBlocks(std::unordered_set<blk_hash_t> &&blocks_hashes,
+                              const std::shared_ptr<network::tarcap::TaraxaPeer> &peer, uint64_t peer_period);
```
MatusKysel, line 179: I think, we should not pass peer object, but return something like `getNonFinalizedBlocks`, but instead of hashes we return blocks for sending and period

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -452,6 +453,35 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
...
+          if (auto blk = dag_blk_mgr_->getDagBlock(hash); blk) {
+            dag_blocks.emplace_back(blk);
```
JakubFornadel, line 471: std::move(blk)

JakubFornadel:
We are also doing copies in void `DagBlockPacketHandler::sendBlock(dev::p2p::NodeID const &peer_id, taraxa::DagBlock block, ...)` -> should be `const DagBlock& block`

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -452,6 +453,35 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
                << " pivot = " << frontier_.pivot << " tips: " << frontier_.tips;
 }
 
+void DagManager::sendNonFinalizedBlocks(std::unordered_set<blk_hash_t> &&blocks_hashes,
+                                        const std::shared_ptr<network::tarcap::TaraxaPeer> &peer,
+                                        uint64_t peer_period) {
+  // Sending of blocks is in DagManager class to use DagManager mutex_ which ensures that there is no race condition
+  // between syncing and gossiping dag blocks
```
JakubFornadel, line 460: So how is mutex_ solving this race condition ? I sort of struggle to find the new synchronization mechnism that replaced mutex_for_sending_dag_blocks_. Checked sendBlock() method, which is related to the gossiping and I dont see anything that would solve this race condition there ?

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: but in that case sending would not be under the dagmanager mutex

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: So blocks for gossiping are sent from DagManager::addDagBlock which is under the DagManager::mutex_, blocks for dag syncing is now also sent from DagManager::sendNonFinalizedBlocks which is under the same DagManager::mutex_, so when we get sync request until it is processed we will not be gossiping blocks. I spent a whole day thinking on this and this was the cleanest/safest way I could think of to solve the issue. I am open for discussion if there is a better way to do this

libraries/core_libs/consensus/include/dag/dag.hpp
mfrankovi: Same as the comment I left for Jakub. In that case it is not under the mutex. I am open for discussion on what would be a cleaner solution

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/get_dag_sync_packet_handler.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/dag/dag.cpp
JakubFornadel: hmm yeah I see. I forgot what is actually the problem in sending sync dag blocks and gossiping new dag blocks concurrently though ?

PR #1397, opened 2022-01-12
fix(compile): fix -Wrange-loop-bind-reference compile error

gaoqi8649:


libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -382,7 +382,7 @@ void DagManager::worker() {
       // DAG block should not have been verified if it is missing a transaction
       // This check is taking some resources so in time we can remove it but it is safe to have it as a sanity check
       size_t trx_found_count = 0;
-      for (const auto &b : db_->transactionsInDb(missing_trxs)) {
+      for (auto &&b : db_->transactionsInDb(missing_trxs)) {
```
MatusKysel, line 385: we are not changing `b` so why would you remove `const`? `for (const auto b : db_->transactionsInDb(missing_trxs))` -> this would be better

gaoqi8649: Fixed, Thanks

MatusKysel (approved):

PR #1399, opened 2022-01-12
Dag sync deadlocalk fix v2

mfrankovi:
On devnet one of the nodes experienced a deadlock which blocked some of the threads and the process ran out of memory.
This was caused by DagManager::mutex_ and TaraxaPeer::mutex_for_sending_dag_blocks_ mutexes which were invoked in reverse order on gossiping the blocks compared to syncing dag blocks.

I have excluded gossiping from the DagManager::mutex_ but in order to do that and preserve the order of the dag blocks being gossiped another mutex is used in addDagBlock which prevents proposed block to be gossiped at the same time as blocks we get from the queue.

### DIFF
--- a/libraries/core_libs/consensus/include/dag/dag.hpp
+++ b/libraries/core_libs/consensus/include/dag/dag.hpp
@@ -114,8 +114,8 @@ class FullNode;
 
 class DagManager : public std::enable_shared_from_this<DagManager> {
  public:
-  using ULock = boost::unique_lock<boost::shared_mutex>;
-  using SharedLock = boost::shared_lock<boost::shared_mutex>;
+  using ULock = std::unique_lock<std::shared_mutex>;
+  using SharedLock = std::shared_lock<std::shared_mutex>;
 
   explicit DagManager(blk_hash_t const &genesis, addr_t node_addr, std::shared_ptr<TransactionManager> trx_mgr,
                       std::shared_ptr<PbftChain> pbft_chain, std::shared_ptr<DagBlockManager> dag_blk_mgr,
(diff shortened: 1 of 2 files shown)


libraries/core_libs/consensus/src/pbft/block_proposer.cpp
```diff
@@ -179,7 +179,7 @@ void BlockProposer::proposeBlock(DagFrontier&& frontier, level_t level, SharedTr
   // simply add the block to the DAG
   DagBlock blk(frontier.pivot, std::move(level), std::move(frontier.tips), std::move(trx_hashes), std::move(vdf),
                node_sk_);
-  dag_mgr_->addDagBlock(blk, std::move(trxs), true);
+  dag_blk_mgr_->pushVerifiedBlockToFront(blk);
```
MatusKysel, line 182: I think trxs is not needed anymore, maybe we could change `SharedTransactions shared_trxs = proposer->getShardedTrxs();` to return just hashes?

gaoqi8649 (changes requested):
We add the own proposed DAG blocks into DAG directly, because we hit the issue of a node may propose many DAG blocks in the same level when process DAG blocks on queue get slow in previous time.
The change may cause the issue happen again. 
If really want to change in this way, I think we have to add a variable to remember the last proposal level. And checking for sure a node only be able to propose one DAG block in each one level.

libraries/core_libs/consensus/src/pbft/block_proposer.cpp
mfrankovi: I have made changes so this does not longer apply

mfrankovi:
> We add the own proposed DAG blocks into DAG directly, because we hit the issue of a node may propose many DAG blocks in the same level when process DAG blocks on queue get slow in previous time. The change may cause the issue happen again. If really want to change in this way, I think we have to add a variable to remember the last proposal level. And checking for sure a node only be able to propose one DAG block in each one level.

The reason 

> We add the own proposed DAG blocks into DAG directly, because we hit the issue of a node may propose many DAG blocks in the same level when process DAG blocks on queue get slow in previous time. The change may cause the issue happen again. If really want to change in this way, I think we have to add a variable to remember the last proposal level. And checking for sure a node only be able to propose one DAG block in each one level.

I have changed my implementation and reverted the change you were mentioning. I am not using a separate mutex which ensures gossiping order

JakubFornadel (approved):

mfrankovi:
> We add the own proposed DAG blocks into DAG directly, because we hit the issue of a node may propose many DAG blocks in the same level when process DAG blocks on queue get slow in previous time. The change may cause the issue happen again. If really want to change in this way, I think we have to add a variable to remember the last proposal level. And checking for sure a node only be able to propose one DAG block in each one level.

PR #1409, opened 2022-01-13
chore: improve logging

mfrankovi:
I have refactored logging for the following log channels:
DAG_BLOCK_PH
DAG_SYNC_PH
GET_DAG_SYNC_PH
GET_PBFT_SYNC_PH
PBFT_BLOCK_PH
PBFT_SYNC_PH
TRANSACTION_PH
DAGBLKMGR
DAGMGR

Main focus is on logging on INFO level and any level above(WARN, ERROR)
These channels will now on INFO level provide all the data needed to track gossip and syncing for transactions, dag blocks and pbft blocks. INFO level logs are made to be as minimal as possible with no repetitions. Only new transactions and blocks are logged, any packets containing data that is already known are not logged on INFO level.

### DIFF
--- a/libraries/cli/include/cli/default_config.hpp
+++ b/libraries/cli/include/cli/default_config.hpp
@@ -42,30 +42,6 @@ const char *default_json = R"foo({
         "verbosity": "ERROR",
         "channels": [
           {
-            "name": "DAG_BLOCK_PH",
-            "verbosity": "INFO"
-          },{
-            "name": "DAG_SYNC_PH",
-            "verbosity": "INFO"
-          },{
-            "name": "GET_DAG_SYNC_PH",
-            "verbosity": "INFO"
-          },{
-            "name": "GET_PBFT_SYNC_PH",
-            "verbosity": "INFO"
-          },{
-            "name": "PBFT_BLOCK_PH",
-            "verbosity": "INFO"
-          },{
-            "name": "TRANSACTION_PH",
-            "verbosity": "INFO"
-          },{
-            "name": "DAGBLKMGR",
-            "verbosity": "INFO"
-          },{
-            "name": "DAGMGR",
-            "verbosity": "INFO"
-          },{
             "name": "SUMMARY",
             "verbosity": "INFO"
           }
--- a/libraries/cli/include/cli/default_config.hpp
+++ b/libraries/cli/include/cli/default_config.hpp
@@ -42,6 +42,30 @@ const char *default_json = R"foo({
         "verbosity": "ERROR",
         "channels": [
           {
+            "name": "DAG_BLOCK_PH",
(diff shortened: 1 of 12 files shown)


JakubFornadel (commented):
I think the main tool that sets channels on our networks is taraxa-testnet. Not sure if it overrides config or how exactly it works :) Check out https://github.com/Taraxa-project/taraxa-testnet/blob/development/tests/taraxa-node/mainnet.values.yaml

there is also devnet/testnet.values.yaml

You should create PR also in the repo and set the channels

libraries/cli/include/cli/default_config.hpp
```diff
@@ -42,6 +42,30 @@ const char *default_json = R"foo({
...
+          },{
+            "name": "TRANSACTION_PH",
+            "verbosity": "INFO"
+          },{
+            "name": "DAGBLKMGR",
+            "verbosity": "INFO"
+          },{
+            "name": "DAGMGR",
+            "verbosity": "INFO"
+          },{
```
JakubFornadel, line 68: Is there any reason why we should not change also testnet&devnet configs ?

libraries/core_libs/consensus/include/transaction_manager/transaction_manager.hpp
```diff
@@ -52,6 +52,14 @@ class TransactionManager : public std::enable_shared_from_this<TransactionManage
    */
   uint32_t insertBroadcastedTransactions(const SharedTransactions &txs);
 
+  /**
+   * @brief Checks the cache if transaction is already seen
+   *
+   * @param trx_hash transaction hash
+   * @return true if seen
+   */
+  bool transactionSeen(const trx_hash_t &trx_hash);
```
MatusKysel, line 61: bool transactionSeen(const trx_hash_t &trx_hash) **const**;

MatusKysel (approved):

JakubFornadel (approved):

libraries/cli/include/cli/default_config.hpp
mfrankovi: Actually this change made it in by mistake. I will revert it. I would like to first make these changes on devnet and test it on devnet before applying the changes to other networks as well

libraries/core_libs/consensus/include/transaction_manager/transaction_manager.hpp
mfrankovi: Fixed

JakubFornadel: how is this fixed ? You are still missing const at the end of function declaration

JakubFornadel (approved):

PR #1412, opened 2022-01-14
PBFT Produce Blocks with NULL Anchor

gaoqi8649:
PBFT produce blocks with NULL HASH DAG block as anchor when there is no new DAG blocks generate

### DIFF
--- a/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
+++ b/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
@@ -61,10 +61,10 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
    * @param prev_blk_hash previous PBFT block hash
    * @param anchor_hash proposed DAG pivot block hash for finalization
    * @param order_hash the hash of all DAG blocks include in the PBFT block
-   * @return PBFT block
+   * @return PBFT block hash
    */
-  std::shared_ptr<PbftBlock> generatePbftBlock(const blk_hash_t &prev_blk_hash, const blk_hash_t &anchor_hash,
-                                               const blk_hash_t &order_hash);
+  blk_hash_t generatePbftBlock(const blk_hash_t &prev_blk_hash, const blk_hash_t &anchor_hash,
+                               const blk_hash_t &order_hash);
   std::shared_ptr<Vote> generateVote(blk_hash_t const &blockhash, PbftVoteTypes type, uint64_t round, size_t step);
 
   size_t getDposTotalVotesCount() const;
--- a/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
+++ b/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
@@ -162,7 +162,7 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
 
   std::atomic<bool> stopped_ = true;
 
-  // Only be able to propose one PBFT block for each period
+  // Ensures that only one PBFT block per period can be proposed
   blk_hash_t proposed_block_hash_ = NULL_BLOCK_HASH;
 
   std::unique_ptr<std::thread> daemon_;
--- a/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
+++ b/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
@@ -56,6 +56,15 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
   size_t getTwoTPlusOne() const;
   void setPbftStep(size_t const pbft_step);
 
+  /**
+   * @brief Generate PBFT block, push into unverified queue, and broadcast to peers
+   * @param prev_blk_hash previous PBFT block hash
+   * @param anchor_hash proposed DAG pivot block hash for finalization
+   * @param order_hash the hash of all DAG blocks include in the PBFT block
+   * @return PBFT block
+   */
+  std::shared_ptr<PbftBlock> generatePbftBlock(const blk_hash_t &prev_blk_hash, const blk_hash_t &anchor_hash,
+                                               const blk_hash_t &order_hash);
   std::shared_ptr<Vote> generateVote(blk_hash_t const &blockhash, PbftVoteTypes type, uint64_t round, size_t step);
 
   size_t getDposTotalVotesCount() const;
@@ -116,9 +125,9 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
 
   uint64_t getThreshold(PbftVoteTypes vote_type) const;
 
-  std::pair<blk_hash_t, bool> proposeMyPbftBlock_();
+  blk_hash_t proposePbftBlock_();
 
-  std::pair<blk_hash_t, bool> identifyLeaderBlock_();
+  blk_hash_t identifyLeaderBlock_();
 
   h256 getProposal(const std::shared_ptr<Vote> &vote) const;
 
@@ -129,7 +138,7 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
   bool broadcastAlreadyThisStep_() const;
 
   bool comparePbftBlockScheduleWithDAGblocks_(blk_hash_t const &pbft_block_hash);
-  std::optional<vec_blk_t> comparePbftBlockScheduleWithDAGblocks_(std::shared_ptr<PbftBlock> pbft_block);
+  std::pair<vec_blk_t, bool> comparePbftBlockScheduleWithDAGblocks_(std::shared_ptr<PbftBlock> pbft_block);
 
   bool pushCertVotedPbftBlockIntoChain_(blk_hash_t const &cert_voted_block_hash,
                                         std::vector<std::shared_ptr<Vote>> &&cert_votes_for_round);
@@ -154,8 +163,9 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
   std::shared_ptr<PbftBlock> getUnfinalizedBlock_(blk_hash_t const &block_hash);
 
   std::atomic<bool> stopped_ = true;
-  // Using to check if PBFT block has been proposed already in one period
-  std::pair<blk_hash_t, bool> proposed_block_hash_ = std::make_pair(NULL_BLOCK_HASH, false);
+
+  // Only be able to propose one PBFT block for each period
+  blk_hash_t proposed_block_hash_ = NULL_BLOCK_HASH;
 
   std::unique_ptr<std::thread> daemon_;
   std::shared_ptr<DbStorage> db_;
--- a/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
+++ b/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
@@ -78,8 +78,6 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
   // Notice: Test purpose
   // TODO: Add a check for some kind of guards to ensure these are only called from within a test
   void setSortitionThreshold(size_t const sortition_threshold);
-  std::vector<std::vector<uint>> createMockTrxSchedule(
-      std::shared_ptr<std::vector<std::pair<blk_hash_t, std::vector<bool>>>> trx_overlap_table);
   size_t getPbftCommitteeSize() const { return COMMITTEE_SIZE; }
   u_long getPbftInitialLambda() const { return LAMBDA_ms_MIN; }
   void setLastSoftVotedValue(blk_hash_t soft_voted_value);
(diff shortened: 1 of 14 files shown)


libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
```diff
@@ -56,6 +56,15 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
...
   void setPbftStep(size_t const pbft_step);
 
+  /**
+   * @brief Generate PBFT block, push into unverified queue, and broadcast to peers
+   * @param prev_blk_hash previous PBFT block hash
+   * @param anchor_hash proposed DAG pivot block hash for finalization
+   * @param order_hash the hash of all DAG blocks include in the PBFT block
+   * @return PBFT block
+   */
+  std::shared_ptr<PbftBlock> generatePbftBlock(const blk_hash_t &prev_blk_hash, const blk_hash_t &anchor_hash,
```
MatusKysel, line 66: You never used result of this function, just the hash of pbft block

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -597,20 +603,28 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
...
-      addToDag(anchor_, blk_hash_t(), vec_blk_t(), 0, true);
-
+      if (anchor) {
+        anchor_ = anchor;
+        LOG(log_nf_) << "Recover anchor " << anchor_;
+        addToDag(anchor_, blk_hash_t(), vec_blk_t(), 0, true);
+        pbft_block_hash = pbft_block.getPrevBlockHash();
+        break;
+      }
       pbft_block_hash = pbft_block.getPrevBlockHash();
```
MatusKysel, line 618: line 615 & 618 are same

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -597,20 +603,28 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
...
-        old_anchor_ = dag_block_hash_as_anchor;
+    }
+
+    while (pbft_block_hash) {
+      auto pbft_block = pbft_chain_->getPbftBlockInChain(pbft_block_hash);
+      auto anchor = pbft_block.getPivotDagBlockHash();
+      if (anchor) {
+        old_anchor_ = anchor;
+        LOG(log_nf_) << "Recover old anchor " << old_anchor_;
+        break;
```
MatusKysel, line 627: This can be done in the same loop as searching by anchor. You can used just one `bool` variable to distinguish between search for anchor and old one

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1083,6 +1079,23 @@ void PbftManager::secondFinish_() {
   loop_back_finish_state_ = elapsed_time_in_round_ms_ > end_time_for_step;
 }
 
+std::shared_ptr<PbftBlock> PbftManager::generatePbftBlock(const blk_hash_t &prev_blk_hash,
+                                                          const blk_hash_t &anchor_hash, const blk_hash_t &order_hash) {
+  auto propose_period = pbft_chain_->getPbftChainSize() + 1;
+  auto pbft_block =
+      std::make_shared<PbftBlock>(prev_blk_hash, anchor_hash, order_hash, propose_period, node_addr_, node_sk_);
```
MatusKysel, line 1086: can be `const`

tests/network_test.cpp
```diff
@@ -1261,7 +1261,6 @@ TEST_F(NetworkTest, node_full_sync) {
               nodes[0]->getDagManager()->getNumVerticesInDag().first);
     EXPECT_EQ(nodes[i]->getDagManager()->getNumVerticesInDag().first, nodes[i]->getDB()->getNumDagBlocks());
     EXPECT_EQ(nodes[i]->getDagManager()->getNumEdgesInDag().first, nodes[0]->getDagManager()->getNumEdgesInDag().first);
-    EXPECT_TRUE(!nodes[i]->getNetwork()->pbft_syncing());
```
MatusKysel, line 1264: why this was removed?

gaoqi8649: @mfrankovi Could you please answer @MatusKysel question? Why remove this line? Thanks!

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
```diff
@@ -149,8 +156,9 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
   std::shared_ptr<PbftBlock> getUnfinalizedBlock_(blk_hash_t const &block_hash);
 
   std::atomic<bool> stopped_ = true;
-  // Using to check if PBFT block has been proposed already in one period
-  std::pair<blk_hash_t, bool> proposed_block_hash_ = std::make_pair(NULL_BLOCK_HASH, false);
+
+  // Only be able to propose one PBFT block for each period
```
JakubFornadel, line 160: weird wording.... -> Ensures that only one PBFT block per period can be proposed ?

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -542,6 +542,12 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
     return 0;
   }
 
+  if (new_anchor == blk_hash_t(0)) {
```
JakubFornadel, line 545: NULL_BLOCK_HASH is used everywhere, why not also here ?

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -542,6 +542,12 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
...
+    period_ = period;
+    LOG(log_nf_) << "Set new period " << period << " with anchor " << new_anchor;
```
JakubFornadel, line 547: `LOG(log_nf_) << "Set new period " << period << " with NULL_BLOCK_HASH anchor;` would be probably easier to notice when going through logs

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -597,20 +603,28 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
 
 void DagManager::recoverDag() {
   if (pbft_chain_) {
-    blk_hash_t pbft_block_hash = pbft_chain_->getLastPbftBlockHash();
-    if (pbft_block_hash) {
-      PbftBlock pbft_block = pbft_chain_->getPbftBlockInChain(pbft_block_hash);
-      blk_hash_t dag_block_hash_as_anchor = pbft_block.getPivotDagBlockHash();
+    auto pbft_block_hash = pbft_chain_->getLastPbftBlockHash();
+    while (pbft_block_hash) {
```
JakubFornadel, line 607: are you checking if this is != NULL_BLOCK_HASH here ? If yes, it might be better to write it explicitely

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1166,10 +1198,11 @@ std::pair<blk_hash_t, bool> PbftManager::proposeMyPbftBlock_() {
   LOG(log_dg_) << "GHOST size " << ghost.size();
   // Looks like ghost never empty, at lease include the last period dag anchor block
```
JakubFornadel, line 1199: at least

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1529,25 +1527,33 @@ void PbftManager::finalize_(SyncBlock &&sync_block, std::vector<h256> &&finalize
...
+
   if (sync) {
     result.wait();
   }
 }
 
 bool PbftManager::pushPbftBlock_(SyncBlock &&sync_block, vec_blk_t &&dag_blocks_order) {
   auto const &pbft_block_hash = sync_block.pbft_blk->getBlockHash();
-  if (db_->pbftBlockInDb(pbft_block_hash)) {
+  if (pbft_chain_->getPbftChainSize() && db_->pbftBlockInDb(pbft_block_hash)) {
```
JakubFornadel, line 1556: what is this check for ?

tests/network_test.cpp
mfrankovi: Some of our tests counted on the fact that once there is no new transactions being sent to the nodes there is no possibility for node to add new pbft blocks. With this change it is possible even when everything is synced up that pbft progress and that makes it possible that node can get into pbft syncing state even with no new transactions.

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: Fixed

MatusKysel (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
gaoqi8649: Check PBFT chain is not empty. In initialization, pbft chain is empty and I hit crash at db_->pbftBlockInDb

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
gaoqi8649: Fixed, thanks

libraries/core_libs/consensus/src/dag/dag.cpp
gaoqi8649: Fixed, thanks

libraries/core_libs/consensus/src/dag/dag.cpp
gaoqi8649: Fixed, thanks

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
gaoqi8649: Fixed, thanks

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
gaoqi8649: Check PBFT chain is not empty. If Pbft chain doesn't have any blocks, should not check pbftBlockInDb that will check if the pbft_block_hash is in chain already

JakubFornadel: so when does pbft chain not have any blocks except the very start of the node ?

JakubFornadel:
Is this empy pbft block generation also included in some of the test ? If not, can you create a test for it ? Shouldnt bee too difficult ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
gaoqi8649: This is for skipping the first block check when first start of a node

JakubFornadel: I know, but it does not make much sense. `pbft_chain_->getPbftChainSize()` is going to be checked thousands of times every day, but the only case when it helps is at the very first start of the node.... After that it is just useless code that needs to be checked and does not bring any value because if always returns value greater than 0

gaoqi8649: Do you have any idea how we can instead of the check?

gaoqi8649:
> Is this empty pbft block generation also included in some of the tests ? If not, can you create a test for it ? Shouldnt bee too difficult ?

Add an unit test for PBFT produce empty blocks

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
gaoqi8649: Remove pbft_chain_->getPbftChainSize(). pbftBlockInDb doesn't need check if pbft chain empty.

JakubFornadel (approved):

PR #1423, opened 2022-01-17
fix: race condition in retrieving blocks with transactions

mfrankovi:
When processing a getDagSyncPacket non finalized blocks are sent only with unique set of transactions that are not finalized or included in a previous DAG block. Other transactions should already be present on the peer requesting it.

There was a race condition where it was possible that between requesting dag blocks with getNonFinalizedBlocks and transactions with db_->getNonfinalizedTransactions(trx_to_query), some blocks were actually finalized which would lead to sending some missing transactions in the packet. This has been fixed that now blocks and transactions are retrieved with a new method DagManager::getNonFinalizedBlocksWithTransactions which ensures that blocks are retrieved under the DagManager mutex which ensures that blocks cannot be finalized while retrieving this data.

At the same time I have made block finalization in pbft manager atomic by putting all the db and memory pbft finalization of dag blocks and transactions under the transaction and dag manager mutex. Previously when I proposed the same, I know that there were comments that this might hurt performance but I do not believe that this will cause performance issues. I have tested this on devnet which currently had a large blockchain with 40GB database and 270000 blocks. I measured the finalization time which took only about 8 milliseconds on average. The finalization happens every 4 seconds so I do not believe that this should have any significant impact on processing transactions/dag blocks.

### DIFF
--- a/libraries/core_libs/consensus/src/dag/dag.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag.cpp
@@ -422,13 +422,6 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
         trx_mgr_->saveTransactionsFromDagBlock(trxs);
         // Save the dag block
         db_->saveDagBlock(blk);
-
-        // TODO: This is an ugly temporary fix for testnet, a better solution is needed for dag block race condition
-        if (db_->getDagBlockPeriod(blk.getHash()) != nullptr) {
-          db_->removeDagBlock(blk.getHash());
-          LOG(log_er_) << "Block already in DB: " << blk.getHash();
-          return;
-        }
       }
       auto blk_hash = blk.getHash();
       auto pivot_hash = blk.getPivot();
--- a/libraries/core_libs/consensus/src/dag/dag.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag.cpp
@@ -542,7 +542,6 @@ std::vector<blk_hash_t> DagManager::getDagBlockOrder(blk_hash_t const &anchor, u
 }
 
 uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period, vec_blk_t const &dag_order) {
-  ULock lock(mutex_);
   LOG(log_dg_) << "setDagBlockOrder called with anchor " << new_anchor << " and period " << period;
   if (period != period_ + 1) {
     LOG(log_er_) << " Inserting period (" << period << ") anchor " << new_anchor
--- a/libraries/core_libs/consensus/src/dag/dag.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag.cpp
@@ -656,6 +656,39 @@ const std::pair<uint64_t, std::map<uint64_t, std::unordered_set<blk_hash_t>>> Da
   return {period_, non_finalized_blks_};
 }
 
+const std::tuple<uint64_t, std::vector<std::shared_ptr<DagBlock>>, SharedTransactions>
+DagManager::getNonFinalizedBlocksWithTransactions(const std::unordered_set<blk_hash_t> &known_hashes) const {
+  SharedLock lock(mutex_);
+  std::vector<std::shared_ptr<DagBlock>> dag_blocks;
+  std::unordered_set<trx_hash_t> unique_trxs;
+  std::vector<trx_hash_t> trx_to_query;
+  SharedTransactions transactions;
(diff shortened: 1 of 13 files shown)


gaoqi8649 (approved):

JakubFornadel (changes requested):
So I was thinkg about this race condition for quite som time and I think I might have more elegant solution:

The problem is that between calls to blks = getNonFinalizedBlocks and db_->getNonfinalizedTransactions(trx_to_query) some of the non-finalized "blks" were finalized so getNonfinalizedTransactions did not return txs included in "blks", right ?

How about we send in getDagSyncPacket also peer's pbft period. When you receive such requets, you can check peer_period:
- if the difference between peer_period and our period is > 1/2/???, there is no need to even send him non-finalized dag blocks as he is behind and should receive pbft sync block...
- if the difference is <= 1/2/2??? we will proceed in sending him nonfinalized dag blocks
- now inside db->getNonfinalizedTransactions wee look for txs in Columns::transactions column, if we dont find it there(due to race condition), we can lookup up transaction period index and if the tx is saved in a period >= peer_period, we will retrieve it and send it

This way, there is no need for any excessive locking and it should solve this rare race-condition by retrieving few txs from period-data

mfrankovi:
> So I was thinkg about this race condition for quite som time and I think I might have more elegant solution:
> 
> The problem is that between calls to blks = getNonFinalizedBlocks and db_->getNonfinalizedTransactions(trx_to_query) some of the non-finalized "blks" were finalized so getNonfinalizedTransactions did not return txs included in "blks", right ?
> 
> How about we send in getDagSyncPacket also peer's pbft period. When you receive such requets, you can check peer_period:
> 
> * if the difference between peer_period and our period is > 1/2/???, there is no need to even send him non-finalized dag blocks as he is behind and should receive pbft sync block...
> * if the difference is <= 1/2/2??? we will proceed in sending him nonfinalized dag blocks
> * now inside db->getNonfinalizedTransactions wee look for txs in Columns::transactions column, if we dont find it there(due to race condition), we can lookup up transaction period index and if the tx is saved in a period >= peer_period, we will retrieve it and send it
> 
> This way, there is no need for any excessive locking and it should solve this rare race-condition by retrieving few txs from period-data

So the problem that I see with your approach is that you say "if we don't find it there", is that you don't know if you have found them or not. You are not looking for some specific set of transactions. I am looking only for nonfinalized transactions, not all transactions. If your idea is that we always look for all transactions(the exception would be if all transactions are non finalized) also in transaction period index I believe that this can hurt performance more than the locking but I guess it would have to be checked.

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -650,6 +642,39 @@ const std::pair<uint64_t, std::map<uint64_t, std::vector<blk_hash_t>>> DagManage
...
+      if (unique_trxs.emplace(trx).second) {
+        trx_to_query.emplace_back(trx);
+      }
+    }
+  }
+  auto trxs = db_->getNonfinalizedTransactions(trx_to_query);
+  for (auto t : trxs) {
+    transactions.emplace_back(t);
+  }
+  return {period_, std::move(dag_blocks), std::move(transactions)};
```
gaoqi8649, line 675: I think don't need "transactions" here. 
return {period_, std::move(dag_blocks), std::move(trxs)};

libraries/core_libs/network/src/tarcap/packets_handlers/get_dag_sync_packet_handler.cpp
```diff
@@ -37,64 +36,46 @@ void GetDagSyncPacketHandler::process(const PacketData &packet_data,
...
-          } else {
-            LOG(log_er_) << "NonFinalizedBlock " << hash << " not in DB";
-            assert(false);
-          }
-        }
-      }
-    }
+  } else {
+    // There is no point in sending blocks if periods do not match
+    blocks.clear();
```
gaoqi8649, line 44: Does it need transactions clear also?

JakubFornadel (approved):

MatusKysel (approved):

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/get_dag_sync_packet_handler.cpp
mfrankovi: Fixed

Issue #1402, opened 2022-01-12
Override config based on node version


JakubFornadel:
# Task Title

Task: Override config based on node version

## Task Description

We need to have a way to automatically override config in case the new node version changes config params names, etc...

PR #1456, opened 2022-01-19
feat(config): introduce config version updater

MatusKysel:
#1402


gaoqi8649 (approved):

libraries/cli/src/config_updater.cpp
```diff
@@ -0,0 +1,44 @@
...
+    old_conf["network_listen_ip"] = old_conf["network_address"];
+    old_conf.removeMember("network_address");
+  } else if (old_conf["network_listen_ip"].isNull()) {
+    old_conf["network_listen_ip"] = new_conf["network_listen_ip"];
+  }
+}
+//// NetworkIPChange
+
+ConfigUpdater::ConfigUpdater(int network_id) {
+  new_conf_ = Tools::generateConfig((Config::NetworkIdType)network_id);
```
JakubFornadel, line 25: static_cast

libraries/cli/src/config_updater.cpp
```diff
@@ -0,0 +1,44 @@
...
+  // Regiser changes that should applay
```
JakubFornadel, line 26: typo applay

CMakeLists.txt
```diff
@@ -14,6 +14,8 @@ set(TARAXA_DB_MAJOR_VERSION 2)
 # Minor version should be modified when changes to the database are made in the tables that can be rebuilt from the
 # basic tables
 set(TARAXA_DB_MINOR_VERSION 3)
+# If config params are changed we should increase the version
+set(TARAXA_CONFIG_VERSION 1)
```
JakubFornadel, line 18: why dont we use node version instead of this ?

libraries/cli/src/config.cpp
```diff
@@ -168,6 +169,12 @@ Config::Config(int argc, const char* argv[]) {
       fs::create_directories(data_dir);
     }
 
+    // if there is not chain id it is a test
```
JakubFornadel, line 172: Why do we create config updater only if this condition pass ? Shouldnt it be created every time ?

libraries/cli/include/cli/config_updater.hpp
```diff
@@ -0,0 +1,39 @@
...
+class ConfigUpdater {
+ public:
+  ConfigUpdater(int network_id);
+  ~ConfigUpdater() = default;
+
+  void UpdateConfig(const std::string& config_name, Json::Value& old_conf);
+
+ private:
+  Json::Value new_conf_;
+  std::vector<std::unique_ptr<ConfigChange>> config_changes_;
```
JakubFornadel, line 36: Do we need this while boilerplate around ConfigChange class and derived classes.... ? Wouldnt be somehting like this  simpler and achieve the same thing ?

struct ConfigChange {
  VersionType version;
  std::function<void(const std::string& config_name, Json::Value& old_conf)> apply;
}

std::vector<ConfigChange> config_changes_;

private:
void registerConfigChanges() {
  // register all config changes functions

  // IP config change
  ConfigChange networkIPChange(1.0.0, [](Json::Value& old_conf, const Json::Value& new_conf){ code... });
 config_changes_.push_back(networkIPChange)
}

CMakeLists.txt
MatusKysel: changing of node version does not mean, changing the config. But I guess we can do it like that

libraries/cli/src/config.cpp
MatusKysel: read the comment above you line. As test do not have chain id, I can not apply update

JakubFornadel (approved):

PR #1466, opened 2022-01-20
fix: initial vote sync

mfrankovi:
Verified votes retrieved from db on startup were broadcasted with net->onNewPbftVotes but on startup no peer is actually connected yet. As a temporary fix I delay the broadcast for 2 minutes when it is expected that there will be peers connected. A better fix is needed on develop.

### DIFF
--- a/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
+++ b/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
@@ -34,6 +34,14 @@ void VoteManager::retreieveVotes_() {
   auto verified_votes = db_->getVerifiedVotes();
   const auto pbft_step = db_->getPbftMgrField(PbftMgrRoundStep::PbftStep);
   for (auto const& v : verified_votes) {
+    addVerifiedVote(v);
+    LOG(log_dg_) << "Retrieved verified vote " << *v;
+  }
+
+  // TODO: Implement votes sync properly
+  // Since this is invoked on startup no peers are connected yet, invoke two minutes after startup
+  thisThreadSleepForSeconds(120);
+  for (auto const& v : verified_votes) {
     // Rebroadcast our own next votes in case we were partitioned...
     if (v->getStep() >= FIRST_FINISH_STEP && pbft_step > EXTENDED_PARTITION_STEPS) {
       std::vector<std::shared_ptr<Vote>> votes = {v};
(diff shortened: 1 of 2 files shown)


libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -33,6 +33,14 @@ void VoteManager::retreieveVotes_() {
...
+    addVerifiedVote(v);
+    LOG(log_dg_) << "Retrieved verified vote " << *v;
+  }
+
+  // TODO: Implement votes sync properly
+  // Since this is invoked on startup no peers are connected yet, invoke two minutes after startup
+  thisThreadSleepForSeconds(120);
   for (auto const& v : verified_votes) {
     // Rebroadcast our own next votes in case we were partitioned...
     if (v->getStep() >= FIRST_FINISH_STEP && pbft_step > EXTENDED_PARTITION_STEPS) {
```
MatusKysel, line 46: hmm, and what about `pbft_step > EXTENDED_PARTITION_STEPS`. With back-off this definitely does not make sense, as we were not reaching steps above 1k

mfrankovi: all of our consensus nodes are now over 1000, they are currently on step 8000 but yes with the backoff 1000 is too much

JakubFornadel (approved):

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -20,19 +20,38 @@ VoteManager::VoteManager(addr_t node_addr, std::shared_ptr<DbStorage> db, std::s
...
   const auto pbft_step = db_->getPbftMgrField(PbftMgrRoundStep::PbftStep);
+  for (auto const& v : verified_votes) {
+    addVerifiedVote(v);
+    LOG(log_dg_) << "Retrieved verified vote " << *v;
+  }
+
+  // TODO: Implement votes sync properly
+  // Since this is invoked on startup no peers are connected yet, invoke two minutes after startup
+  for (int i = 0; i < 120; i++) {
+    if (stopped_ == false) {
```
gaoqi8649, line 50: Since retreieveVotes_ only be called once in constructor, I think we don't need the "stopped_" variable. Why we need the stopped_ here?

gaoqi8649 (approved):

PR #1470, opened 2022-01-21
Hotfix new testnet

MatusKysel:


for_devs/local-net
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env /usr/local/bin/python3
```
JakubFornadel, line 1: this should not have been changed probably ?

rattrap: We want to use the python version from `venv` so this is incorrect

rattrap (changes requested):

for_devs/local-net
kstdl: Sorry, missed that. That should be only my local change

justinsnapp:
Forced push this branch to testnet.

PR #1472, opened 2022-01-21
chore: fix bootnodes for private networks

MatusKysel:


libraries/cli/include/cli/config.hpp
```diff
@@ -19,7 +19,7 @@ class Config {
   // Retrieves loaded node configuration
   FullNodeConfig getNodeConfiguration();
 
-  enum class NetworkIdType { Mainnet = 1, Testnet, Devnet };
+  enum class NetworkIdType { Mainnet = 1, Testnet, Devnet, LastNetworkId };
```
gaoqi8649, line 22: What does the LastNetworkId use for?
Looks like in Mainnet, testnet and devnet will do boot host add nodes. Not for LastNetworkId

JakubFornadel (approved):

libraries/cli/include/cli/config.hpp
MatusKysel: It is used in check as you described, but this is generic approach in case we add new network/s. e.g. I do not need to change anything if we add `supernet` because enum will be `Mainnet = 1, Testnet, Devnet, Super, LastNetworkId` and my condition will be still valid

gaoqi8649 (approved):

Issue #1476, opened 2022-01-21
PBFT get stalled for not enough votes in Testnet


gaoqi8649:
## Task Description

Usually we have more than 2/3 votes in Testnet. We lost majority votes in Testnet, and seems like have 1/3 votes have get partition. And there are no enough (2t + 1) votes in Testnet for PBFT to make progress

PR #1477, opened 2022-01-21
Hotfix not enough votes in Testnet

gaoqi8649:
## Purpose
Usually we have more than 2/3 votes in Testnet. We lost majority votes in Testnet, and seems like have 1/3 votes have get partition. And there are no enough (2t + 1) votes in Testnet for PBFT to make progress
Fix https://github.com/Taraxa-project/taraxa-node/issues/1476

Current node weighted votes = dpos weighted votes * (committee size / DPOS total votes)
The hotfix will change the node weighted votes = dpos weighted votes
The fix will increase the weight in our consensus nodes.


JakubFornadel:
I would rather do the reset than this

justinsnapp:
We all agreed NOT to go this route.

PR #1479, opened 2022-01-24
chore(syncing): validate sync pbft block in handler

### LINKED ISSUE #1160: Validate sync blocks in packet handler
## Is your feature request related to a problem? Please describe.

- Most of the checks that we do in PbftManager::processSyncBlock() are static checks that can be done as soon as packet with sync blocks is received. If one of the checks fails, such block should not be pushed into the sync_queue. What we do now instead is that we simply push everything into the queue and then perform those checks after the block is popped and processed inside pbft_manager and basically trigger the functionality from tarcap(disconnect, set malicious peer, etc..) that should have been triggered as soon as packet was received 

- sync_queue should have probably stayed inside pbft_chain and not inside pbft_manager - check what is better...

## Describe the solution you'd like.

Do you have ideas regarding the implementation of this feature?
Are you willing to implement this feature?

MatusKysel:
fixes #1160


libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -46,31 +46,72 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
...
   SyncBlock sync_block(*it);
-
-  string received_dag_blocks_str;
-  for (auto const &block : sync_block.dag_blocks) {
-    received_dag_blocks_str += block.getHash().toString() + " ";
-    if (peer->dag_level_ < block.getLevel()) {
-      peer->dag_level_ = block.getLevel();
+  const auto pbft_blk_hash = sync_block.pbft_blk->getBlockHash();
+
+  {  // This is just log related stuff
```
gaoqi8649, line 51: Not just log related stuff, also update max peer DAG level

gaoqi8649 (approved):

mfrankovi (approved):

PR #1486, opened 2022-01-25
chore(trxs): correct transaction validation

### LINKED ISSUE #1483: No forefront tx execution check on tmp state
## Task Description

We allow processing of invalid txs. Nice example is current mainnet where all txs fail due to insufficient balance but we broadcast them as valid, they are included in dag&pbft blocks, drain resources of our validators and they are useless...

##### Solution
Tx could be executed on some tmp test state and if it does not succeed, it will be rejected.

## Epic Parent

MatusKysel:
fixes #1483


mfrankovi (approved):

gaoqi8649 (approved):
Looks good to me

PR #1487, opened 2022-01-26
PBFT Rebroadcast Soft Votes

gaoqi8649:
## Purpose

If have seen enough (2t+1) soft votes in the current round, PBFT should rebroadcast them in second finish state
fix https://github.com/Taraxa-project/taraxa-node/issues/1488

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -738,18 +738,10 @@ bool PbftManager::updateSoftVotedBlockForThisRound_() {
 
     if (soft_voted_block_for_this_round_.first != NULL_BLOCK_HASH && soft_voted_block_for_this_round_.second) {
       // Have enough soft votes for a value other other than NULL BLOCK HASH...
-
-      if (state_ == finish_polling_state) {
-        LOG(log_dg_) << "Node has seen enough soft votes voted at " << soft_voted_block_for_this_round_.first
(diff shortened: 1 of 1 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1035,6 +1027,18 @@ void PbftManager::secondFinish_() {
   auto end_time_for_step = (2 + step_ - startingStepInRound_) * LAMBDA_ms - POLLING_INTERVAL_ms;
 
   updateSoftVotedBlockForThisRound_();
+  if (soft_voted_block_for_this_round_.first != NULL_BLOCK_HASH && soft_voted_block_for_this_round_.second) {
```
justinsnapp, line 1030: And we can call this every second finishing step without re-gossiping, correct?

gaoqi8649: Yes

mfrankovi (approved):

MatusKysel (approved):

kstdl (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: @justinsnapp I will resolve this conversation to be able to test the fix on devnet to get it ustuck. If there is something still unclear please follow up

PR #1490, opened 2022-01-27
Transaction prio queue

### LINKED ISSUE #1481: Dag block txs ordering
## Task Description

Ledger is not correctly ordering transactions when creating dag blocks. We keep unexecuted txs inside unordered map and when creating dag block, 250 txs are retrieved from unordered map by iterating over the map from the beginning - this means those 250 txs are basically random. Then they are ordered by sender & nonce. Problem is that for example if there is 10k txs in tx pool already and sender A sends 2 txs: TX1 with nonce 1 and fee 10 and TX2 with nonce 2 and fee 5, it is possible that due to randomness TX2 will be saved at the beggining of the transaction pool and TX1 at the end, so TX2 will be executed first and fails due to invalid nonce.... This should never happen

Solution
Create much more sophisticated tx pool, which would serve 250 ordered txs based on fee & sender & nonce ?
See eth ordering algo here:
https://github.com/ethereum/go-ethereum/blob/290e851f57f5d27a1d5f0f7ad784c836e017c337/core/types/transaction.go#L372

## Epic Parent

MatusKysel:
fixes #1481

### DIFF
--- a/libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
+++ b/libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
@@ -62,24 +62,11 @@ class TransactionQueue {
   size_t size() const;
 
  private:
-  struct PriorityCompare {
-    TransactionQueue& queue;
-    // Compare transaction by nonce height and gas price.
-    bool operator()(const std::shared_ptr<Transaction>& first, const std::shared_ptr<Transaction>& second) const {
-      const auto& height1 = first->getNonce() - queue.nonce_queue_[first->getSender()].begin()->first;
-      const auto& height2 = second->getNonce() - queue.nonce_queue_[second->getSender()].begin()->first;
-      if (first->getSender() == second->getSender()) {
-        return height1 < height2 || (height1 == height2 && first->getGasPrice() > second->getGasPrice());
-      } else {
-        return first->getGasPrice() > second->getGasPrice();
-      }
-    }
-  };
+  typedef std::function<bool(const std::shared_ptr<Transaction>&, const std::shared_ptr<Transaction>&)> ComperType;
   // It has to be multiset as two trx could have same value (nonce and gas price)
-  using PriorityQueue = std::multiset<std::shared_ptr<Transaction>, PriorityCompare>;
+  using PriorityQueue = std::multiset<std::shared_ptr<Transaction>, ComperType>;
   PriorityQueue priority_queue_;
   std::unordered_map<trx_hash_t, PriorityQueue::iterator> hash_queue_;
-  std::unordered_map<addr_t, std::map<u256, PriorityQueue::iterator>> nonce_queue_;
 };
 
 }  // namespace taraxa
--- a/libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
+++ b/libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
@@ -19,7 +19,7 @@ class TransactionQueue {
  private:
   struct PriorityCompare {
     TransactionQueue& queue;
-    /// Compare transaction by nonce height and gas price.
+    // Compare transaction by nonce height and gas price.
     bool operator()(const std::shared_ptr<Transaction>& first, const std::shared_ptr<Transaction>& second) const {
       const auto& height1 = first->getNonce() - queue.nonce_queue_[first->getSender()].begin()->first;
       const auto& height2 = second->getNonce() - queue.nonce_queue_[second->getSender()].begin()->first;
--- a/libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
+++ b/libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
@@ -3,17 +3,62 @@
 #include "transaction/transaction.hpp"
 
 namespace taraxa {
-
+/**
+ * @brief this is NOT thread safe class. It is proteced only by transactions_mutex_ in the transaction mamanger !!!
+ *
+ */
 class TransactionQueue {
  public:
   TransactionQueue();
 
+  /**
+   * @brief insert a transaction into the queue, sorted by priority
+   *
+   * @param transaction
+   * @return true
+   * @return false
+   */
   bool insert(const std::shared_ptr<Transaction>& transaction);
+
+  /**
+   * @brief removed transaction from queue
+   *
+   * @param hash
+   * @return true
+   * @return false
+   */
   bool erase(const trx_hash_t& hash);
 
+  /**
+   * @brief returns the transaction or null
+   *
+   * @param hash
+   * @return std::shared_ptr<Transaction>
+   */
   std::shared_ptr<Transaction> get(const trx_hash_t& hash) const;
+
+  /**
+   * @brief returns up to the number of requested transaction sorted by priority
+   *
+   * @param count
+   * @return std::vector<std::shared_ptr<Transaction>>
+   */
   std::vector<std::shared_ptr<Transaction>> get(uint64_t count = 0) const;
+
+  /**
+   * @brief returns true/false if the transaction is in the queue
+   *
+   * @param hash
+   * @return true
+   * @return false
+   */
   bool contains(const trx_hash_t& hash) const;
+
+  /**
+   * @brief returns size of priority queue
+   *
+   * @return size_t
+   */
   size_t size() const;
 
  private:
@@ -28,7 +73,7 @@ class TransactionQueue {
   };
   // It has to be multiset as two trx could have same value (nonce and gas price)
   using PriorityQueue = std::multiset<std::shared_ptr<Transaction>, PriorityCompare>;
-  PriorityQueue prio_queue_;
+  PriorityQueue priority_queue_;
   std::unordered_map<trx_hash_t, PriorityQueue::iterator> hash_queue_;
   std::unordered_map<addr_t, std::map<u256, PriorityQueue::iterator>> nonce_queue_;
 };
--- a/libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
+++ b/libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
@@ -1,0 +1,36 @@
+#pragma once
+
+#include "transaction/transaction.hpp"
+
+namespace taraxa {
+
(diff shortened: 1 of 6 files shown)


libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
```diff
@@ -0,0 +1,36 @@
...
+    /// Compare transaction by nonce height and gas price.
+    bool operator()(const std::shared_ptr<Transaction>& first, const std::shared_ptr<Transaction>& second) const {
+      const auto& height1 = first->getNonce() - queue.nonce_queue_[first->getSender()].begin()->first;
+      const auto& height2 = second->getNonce() - queue.nonce_queue_[second->getSender()].begin()->first;
+      return height1 < height2 || (height1 == height2 && first->getGasPrice() > second->getGasPrice());
+    }
+  };
+  // It has to be multiset as two trx could have same value (nonce and gas price)
+  using PriorityQueue = std::multiset<std::shared_ptr<Transaction>, PriorityCompare>;
+  PriorityQueue prio_queue_;
```
JakubFornadel, line 31: prio_queue ? why dont you call priority_queue ? :D

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -254,26 +233,8 @@ size_t TransactionManager::getNonfinalizedTrxSize() const {
...
-        count++;
-      }
-    }
-  }
-
-  // sort trx based on sender and nonce
-  std::sort(to_be_packed_trx.begin(), to_be_packed_trx.end(), trxComp);
-
-  return to_be_packed_trx;
+  std::shared_lock transactions_lock(transactions_mutex_);
```
JakubFornadel, line 236: I think all locking that was protecting previous tx_pool should be moved into the TransactionQueue class

MatusKysel: this is not possible because transactions_mutex_ is locking more then to this pool

MatusKysel: it is used for DB and pool synchronization and many more

JakubFornadel: in such case it would be nice to have a comment in tx_pool header explaining why it does not use locks

libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
```diff
@@ -0,0 +1,36 @@
+#pragma once
+
+#include "transaction/transaction.hpp"
+
+namespace taraxa {
+
+class TransactionQueue {
```
JakubFornadel, line 7: If you use doxygen comments for class description and some important methods, people would not need much much less time to uderstand how exactly this pool works (all the comparing based on nonce, fees, etc...) and it take 5 minutes of your time. Would be nice to comment each class and method, which is not just some dummy data holder.... Definitely saves time => money if someone needs to touch it in the future

libraries/core_libs/consensus/src/transaction_manager/transaction_queue.cpp
```diff
@@ -0,0 +1,65 @@
...
+  // Remove even address from nonce map
+  if (nonce_it->second.empty()) nonce_queue_.erase(nonce_it);
+
+  // Clean the rest
+  prio_queue_.erase(it->second);
+  hash_queue_.erase(it);
+  return true;
+}
+
+bool TransactionQueue::insert(const std::shared_ptr<Transaction> &transaction) {
```
JakubFornadel, line 53: shouldnt be here:
`assert(transaction) ? `

libraries/core_libs/consensus/src/transaction_manager/transaction_queue.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+bool TransactionQueue::erase(const trx_hash_t &hash) {
+  // Find the hash
+  const auto it = hash_queue_.find(hash);
+  if (it == hash_queue_.end()) return false;
+
+  // Remove trx from nonce map
+  const auto nonce_it = nonce_queue_.find((*it->second)->getSender());
+  assert(nonce_it != nonce_queue_.end());
+  nonce_it->second.erase((*it->second)->getNonce());
```
JakubFornadel, line 43: you are missing the same "assert pattern" as you used for sender, so it could happen that you dont find a bug in case there are some nonces for sender saved, but this particular tx is missing there...

libraries/core_libs/consensus/src/transaction_manager/transaction_queue.cpp
```diff
@@ -0,0 +1,65 @@
...
+  // Remove even address from nonce map
+  if (nonce_it->second.empty()) nonce_queue_.erase(nonce_it);
+
+  // Clean the rest
+  prio_queue_.erase(it->second);
```
JakubFornadel, line 48: probably "assert pattern" also for priority_queue ?

libraries/core_libs/consensus/src/transaction_manager/transaction_queue.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+std::vector<std::shared_ptr<Transaction>> TransactionQueue::get(uint64_t count) const {
+  if (count == 0 || prio_queue_.size() <= count) {
+    std::vector<std::shared_ptr<Transaction>> ret(prio_queue_.begin(), prio_queue_.end());
+    return ret;
+  }
+
+  std::vector<std::shared_ptr<Transaction>> ret;
+  ret.reserve(count);
+  uint16_t counter = 0;
```
JakubFornadel, line 26: counters are size_t

libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
```diff
@@ -0,0 +1,36 @@
...
+  bool insert(const std::shared_ptr<Transaction>& transaction);
+  bool erase(const trx_hash_t& hash);
+
+  std::shared_ptr<Transaction> get(const trx_hash_t& hash) const;
+  std::vector<std::shared_ptr<Transaction>> get(uint64_t count = 0) const;
+  bool contains(const trx_hash_t& hash) const;
+  size_t size() const;
+
+ private:
+  struct PriorityCompare {
```
JakubFornadel, line 20: isnt it possible to use lambda instead of struct comparator ?

libraries/core_libs/consensus/src/transaction_manager/transaction_queue.cpp
JakubFornadel, line 53: how and where do we deal with a situation when someone just sends the tx with the same nonce and higher fee ?

libraries/core_libs/consensus/src/transaction_manager/transaction_queue.cpp
MatusKysel: this assert was there just for development, if I am cleaning nonce_q correctly, there is no need for it anymore

MatusKysel: there is no way, that something can be missing anymore

libraries/core_libs/consensus/src/transaction_manager/transaction_queue.cpp
MatusKysel: no need

libraries/core_libs/consensus/src/transaction_manager/transaction_queue.cpp
MatusKysel: copy&pasted old code

libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
MatusKysel: I wanted to do so, but I have no idea how to do it with `this` so I can read internal structures. This is only way I make it work, but if anyone knows how to do it newer way I would appreciated it

libraries/core_libs/consensus/src/transaction_manager/transaction_queue.cpp
MatusKysel: comparator will solve this

libraries/core_libs/consensus/include/transaction_manager/transaction_queue.hpp
JakubFornadel: First step is to make people even look at this PR :)

MatusKysel: Everyone was tagged? what you mean?

JakubFornadel:
Just a note to be left here:

sender A:
 - TXA1, nonce 1, fee 5
 - TXA2, nonce 2, fee 6
 
sender B
 - TXB1, nonde 1, fee 4
 
 If we look at the fees, ordering should be:
TXA1, TXA2, TXB1
 
, but it will be:
 will be TXA1, TXB1, TXA2

MatusKysel:
> Just a note to be left here:
> 
> sender A:
> 
>     * TXA1, nonce 1, fee 5
> 
>     * TXA2, nonce 2, fee 6
> 
> 
> sender B
> 
>     * TXB1, nonde 1, fee 4
> 
> 
> If we look at the fees, ordering should be: TXA1, TXA2, TXB1
> 
> , but it will be: will be TXA1, TXB1, TXA2

I do not know 

> Just a note to be left here:
> 
> sender A:
> 
>     * TXA1, nonce 1, fee 5
> 
>     * TXA2, nonce 2, fee 6
> 
> 
> sender B
> 
>     * TXB1, nonde 1, fee 4
> 
> 
> If we look at the fees, ordering should be: TXA1, TXA2, TXB1
> 
> , but it will be: will be TXA1, TXB1, TXA2

I was thinking about that and we can change it, so it will behaves as you want

MatusKysel:
@JakubFornadel check this https://github.com/Taraxa-project/taraxa-node/pull/1490/commits/5da54b719fba251d3468a9e8c08ae9e4e7f50152 now the ordering should be correct even in your case

JakubFornadel (approved):

mfrankovi (approved):

Issue #1491, opened 2022-01-28
Fix PBFT rebroadcast soft votes for current round in Mainnet


gaoqi8649:
## Task Description

Same issue https://github.com/Taraxa-project/taraxa-node/issues/1488, need fix in Mainnet

JakubFornadel:
@gaoqi8649 we are not using flow where you create separate PR for develop/testnet and master branches... So these issue probably should not have been created

gaoqi8649:
> 

Emm..., If only create issue and have a fix have merged to develop, then the issue will be closed. How to record the issue if have been fix in Mainnet/Testnet when fix have merged into develop?

gaoqi8649:
> @gaoqi8649 we are not using flow where you create separate PR for develop/testnet and master branches... So these issue probably should not have been created

Emm..., If only create issue and have a fix have merged to develop, then the issue will be closed. How to record the issue if have been fix in Mainnet/Testnet when fix have merged into develop?

PR #1492, opened 2022-01-28
Pbft rebroadcast soft votes in master branch

gaoqi8649:
## Purpose
Fix https://github.com/Taraxa-project/taraxa-node/issues/1491

## Changes made
Cherry pick from https://github.com/Taraxa-project/taraxa-node/pull/1487

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -725,18 +725,10 @@ bool PbftManager::updateSoftVotedBlockForThisRound_() {
 
     if (soft_voted_block_for_this_round_.first != NULL_BLOCK_HASH && soft_voted_block_for_this_round_.second) {
       // Have enough soft votes for a value other other than NULL BLOCK HASH...
-
-      if (state_ == finish_polling_state) {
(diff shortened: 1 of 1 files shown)


JakubFornadel (changes requested):
@gaoqi8649 we are not just creating random PRs into the master anymore... The only situation when we would create PR like this into the master is:

- there is a bug in mainnet candidate that needs to be fixed immediately 
- hotfix branch will the actual fix will be created, merged into the master, new tag created, then merged into the testet/develo
- this process must be first discussed and agreed on between others

gaoqi8649:
> @gaoqi8649 we are not just creating random PRs into the master anymore... The only situation when we would create PR like this into the master is:
> 
> * there is a bug in mainnet candidate that needs to be fixed immediately
> * hotfix branch will the actual fix will be created, merged into the master, new tag created, then merged into the testet/develo
> * this process must be first discussed and agreed on between others

The bug is a race condition, and will happen more rarely when nodes size get bigger. But this is still a hidden danger in Mainnet and Testnet.
The fix will not need force all players to upgrade and also don't need set a hardfork. 
I recommend merge the fix to make our mainnet more strong.

JakubFornadel:
closing this as it is already merged in develop

PR #1495, opened 2022-01-31
chore: malicious nodes handling

mfrankovi:
Malicious nodes are now marked as malicious for a specific amount of time defined in network_peer_blacklist_timeout. The default is 600 seconds(10 minutes). It is also possible to disable blacklisting with disable_peer_blacklist flag. Main reason for this is that we had cases that nodes were acting malicious because of a bug or an old version. It could be even because of the config file misconfigured. Any such node could in time be corrected and it should not be blacklisted forever.

All of the blacklisting is memory bases so any node restart will clear it as well.

PR handles most of the cases where invalid data is sent, like invalid blocks, transactions with incorrect difficulty and similar. Node sending blocks with missing transactions. The PR does not handle DDOS attacks where valid data is sent too many times.

### DIFF
--- a/libraries/cli/include/cli/default_config.hpp
+++ b/libraries/cli/include/cli/default_config.hpp
@@ -16,6 +16,7 @@ const char *default_json = R"foo({
   "network_max_peer_count": 50,
   "network_sync_level_size": 10,
   "network_packets_processing_threads": 10,
+  "network_peer_blacklist_timeout" : 600,
   "deep_syncing_threshold" : 10,
   "network_boot_nodes": [
     {
--- a/libraries/config/include/config/config.hpp
+++ b/libraries/config/include/config/config.hpp
@@ -27,6 +27,8 @@ struct NodeConfig {
 };
 
 struct NetworkConfig {
+  static constexpr uint16_t kBlacklistTimeoutDefaultInSeconds = 600;
+
   std::string json_file_name;
   bool network_is_boot_node = 0;
   std::string network_public_ip;
@@ -43,6 +45,8 @@ struct NetworkConfig {
   uint16_t network_performance_log_interval = 0;
   uint16_t network_num_threads = std::max(uint(1), uint(std::thread::hardware_concurrency() / 2));
   uint16_t network_packets_processing_threads = 10;
+  uint16_t network_peer_blacklist_timeout = kBlacklistTimeoutDefaultInSeconds;
+  bool disable_peer_blacklist = false;
   uint16_t deep_syncing_threshold = 10;
 };
 
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
@@ -1,7 +1,6 @@
 #pragma once
 
 #include "network/tarcap/packets_handlers/common/packet_handler.hpp"
-#include "network/tarcap/shared_states/syncing_state.hpp"
 
 namespace taraxa {
 class DagManager;
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "network/tarcap/packets_handlers/common/packet_handler.hpp"
+#include "network/tarcap/shared_states/syncing_state.hpp"
 
 namespace taraxa {
 class DagManager;
@@ -11,12 +12,14 @@ class TransactionManager;
 
 namespace taraxa::network::tarcap {
 
+class SyncingState;
+
 class GetDagSyncPacketHandler : public PacketHandler {
  public:
   GetDagSyncPacketHandler(std::shared_ptr<PeersState> peers_state, std::shared_ptr<PacketsStats> packets_stats,
-                          std::shared_ptr<TransactionManager> trx_mgr, std::shared_ptr<DagManager> dag_mgr,
-                          std::shared_ptr<DagBlockManager> dag_blk_mgr, std::shared_ptr<DbStorage> db,
-                          const addr_t& node_addr);
+                          std::shared_ptr<SyncingState> syncing_state, std::shared_ptr<TransactionManager> trx_mgr,
+                          std::shared_ptr<DagManager> dag_mgr, std::shared_ptr<DagBlockManager> dag_blk_mgr,
+                          std::shared_ptr<DbStorage> db, const addr_t& node_addr);
 
   virtual ~GetDagSyncPacketHandler() = default;
 
@@ -26,6 +29,7 @@ class GetDagSyncPacketHandler : public PacketHandler {
  private:
   void process(const PacketData& packet_data, const std::shared_ptr<TaraxaPeer>& peer) override;
 
+  std::shared_ptr<SyncingState> syncing_state_;
   std::shared_ptr<TransactionManager> trx_mgr_;
   std::shared_ptr<DagManager> dag_mgr_;
   std::shared_ptr<DagBlockManager> dag_blk_mgr_;
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
@@ -12,14 +12,12 @@ class TransactionManager;
 
 namespace taraxa::network::tarcap {
 
-class SyncingState;
-
 class GetDagSyncPacketHandler : public PacketHandler {
  public:
   GetDagSyncPacketHandler(std::shared_ptr<PeersState> peers_state, std::shared_ptr<PacketsStats> packets_stats,
-                          std::shared_ptr<SyncingState> syncing_state, std::shared_ptr<TransactionManager> trx_mgr,
-                          std::shared_ptr<DagManager> dag_mgr, std::shared_ptr<DagBlockManager> dag_blk_mgr,
-                          std::shared_ptr<DbStorage> db, const addr_t& node_addr);
+                          std::shared_ptr<TransactionManager> trx_mgr, std::shared_ptr<DagManager> dag_mgr,
+                          std::shared_ptr<DagBlockManager> dag_blk_mgr, std::shared_ptr<DbStorage> db,
+                          const addr_t& node_addr);
 
   virtual ~GetDagSyncPacketHandler() = default;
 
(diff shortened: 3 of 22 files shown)


JakubFornadel (changes requested):
Maybe we should not have put `malicious_peers` container into the SyncingState class. Logically it does not belong there and it became more obvious now when you had to add syncing_state class member into the transaction packet handler, etc... just to access malicious_peers. Maybe it would make more sense to put into the PacketHandler base class as we might add this this malicious behaviour detection also in other handlers in the future ? Or  some other place but syncing_state does not seem like a right place for it

Otherwise looks ok :)

libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
```diff
@@ -65,16 +64,31 @@ bool SyncingState::is_actively_syncing() const {
...
+    return false;
+  }
+
+  // Peers are marked malicious for the time defined in conf_.network_peer_blacklist_timeout
+  if (auto i = malicious_peers_.find(peer_id); i != malicious_peers_.end()) {
+    if (std::chrono::duration_cast<std::chrono::seconds>(std::chrono::steady_clock::now() - i->second).count() <=
+        conf_.network_peer_blacklist_timeout) {
+      return true;
+    } else {
+      malicious_peers_.erase(i);
```
JakubFornadel, line 87: is_peer_maliciou is called from TaraxaCapability::onConnect. Are we certain that onConnect si 100% synchronous ? If not, we need mutex here

libraries/config/include/config/config.hpp
```diff
@@ -43,6 +45,8 @@ struct NetworkConfig {
   uint16_t network_performance_log_interval = 0;
   uint16_t network_num_threads = std::max(uint(1), uint(std::thread::hardware_concurrency() / 2));
   uint16_t network_packets_processing_threads = 10;
+  uint16_t network_peer_blacklist_timeout = kBlacklistTimeoutDefaultInSeconds;
```
MatusKysel, line 48: maybe we can reuse `network_peer_blacklist_timeout` with value of `0` instead of new param `disable_peer_blacklist`

libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
```diff
@@ -11,12 +12,14 @@ class TransactionManager;
 
 namespace taraxa::network::tarcap {
 
+class SyncingState;
```
MatusKysel, line 15: is there packet handler that does not use `SyncingState` ? if there is non or just one or two, maybe it should be part of `PacketHandler` class

libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
```diff
@@ -62,9 +63,9 @@ class SyncingState {
   std::atomic<bool> deep_pbft_syncing_{false};
   std::atomic<bool> pbft_syncing_{false};
 
-  ExpirationCache<dev::p2p::NodeID> malicious_peers_;
+  std::unordered_map<dev::p2p::NodeID, std::chrono::steady_clock::time_point> malicious_peers_;
```
MatusKysel, line 66: now it is not thread-safe, mutex is needed

libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
```diff
@@ -62,9 +63,9 @@ class SyncingState {
...
 
-  const uint16_t kDeepSyncingThreshold;
+  NetworkConfig conf_;
```
MatusKysel, line 68: this could be const

libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
```diff
@@ -5,8 +5,7 @@
 
 namespace taraxa::network::tarcap {
 
-SyncingState::SyncingState(uint16_t deep_syncing_threshold)
-    : malicious_peers_(300, 50), kDeepSyncingThreshold(deep_syncing_threshold) {}
```
MatusKysel, line 9: `malicious_peers_` should still have maximum size

libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
JakubFornadel: you can see it in registerHandlers function, I think 7 out of 11 are now using SyncingState but they use syncing state only because they need to access malicious_peers, which could be part of PacketHandler rather than syncing_state - check out my first comment

libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
MatusKysel, line 87: so if I do not call `is_peer_malicious` peer will be never erased? this is bad design, whole array should be cleaned here

libraries/config/include/config/config.hpp
mfrankovi: I remember now that when I originally thought about this I wanted to use 0 for infinite timeout although I guess I did not implement it that way. Do you think it is better to have 0 as disable or having 0 as infinite timeout?

libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
mfrankovi: why? wouldnt that just made us open for easier attacks? Is it really realistic that someone tries millions of connections with different IDs that makes malicious_peers create significant memory leak?

mfrankovi:
> Maybe we should not have put `malicious_peers` container into the SyncingState class. Logically it does not belong there and it became more obvious now when you had to add syncing_state class member into the transaction packet handler, etc... just to access malicious_peers. Maybe it would make more sense to create separate class for it and make it PacketHandler class member as we might add this this malicious behaviour detection also in other handlers in the future ? Or some other place but syncing_state does not seem like a right place for it
> 
> Otherwise looks ok :)

I will move it to PeersState, the description of PeersState is: "PeersState contains common members and functions related to peers, host, etc... that are shared among multiple * classes"

So I think it logically belongs here and it is already a member of packethandler

JakubFornadel:
> > Maybe we should not have put `malicious_peers` container into the SyncingState class. Logically it does not belong there and it became more obvious now when you had to add syncing_state class member into the transaction packet handler, etc... just to access malicious_peers. Maybe it would make more sense to create separate class for it and make it PacketHandler class member as we might add this this malicious behaviour detection also in other handlers in the future ? Or some other place but syncing_state does not seem like a right place for it
> > Otherwise looks ok :)
> 
> I will move it to PeersState, the description of PeersState is: "PeersState contains common members and functions related to peers, host, etc... that are shared among multiple * classes"
> 
> So I think it logically belongs here and it is already a member of packethandler

yeah PeersState is the right place for it !

libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
mfrankovi: malicious peers is a thread safe map now

libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
mfrankovi: Moved to PeersState

libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
mfrankovi: map is now threadsafe

libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
mfrankovi: I added a cleanup although I am not that convinced that is better. This does make us check cleanup on every call. But this is called only on node connecting so it should not be that often, I guess it is fine.

libraries/core_libs/network/include/network/tarcap/shared_states/syncing_state.hpp
mfrankovi: fixed

libraries/core_libs/network/src/tarcap/shared_states/syncing_state.cpp
MatusKysel: if there is an cleanup, the we do need to limit size

libraries/config/include/config/config.hpp
MatusKysel: but then this map can be huge, if we never clean it up, maybe there should be some hard limit like 24h or something

libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
```diff
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "network/tarcap/packets_handlers/common/packet_handler.hpp"
+#include "network/tarcap/shared_states/syncing_state.hpp"
```
MatusKysel, line 4: this is probably forgotten change

libraries/core_libs/network/src/tarcap/shared_states/peers_state.cpp
```diff
@@ -102,4 +103,35 @@ std::shared_ptr<TaraxaPeer> PeersState::setPeerAsReadyToSendMessages(dev::p2p::N
...
+            conf_.network_peer_blacklist_timeout) {
+      return true;
+    } else {
+      malicious_peers_.erase(peer_id);
+    }
+  }
+
+  // Delete any expired item from the list
+  if (conf_.network_peer_blacklist_timeout > 0) {
+    malicious_peers_.erase([this](std::chrono::steady_clock::time_point value) {
```
MatusKysel, line 128: std::chrono::steady_clock::time_point**&** value

libraries/core_libs/network/include/network/tarcap/packets_handlers/get_dag_sync_packet_handler.hpp
mfrankovi: Done

libraries/core_libs/network/src/tarcap/shared_states/peers_state.cpp
mfrankovi: Done

MatusKysel (approved):

libraries/cli/include/cli/default_config.hpp
```diff
@@ -16,6 +16,7 @@ const char *default_json = R"foo({
   "network_max_peer_count": 50,
   "network_sync_level_size": 10,
   "network_packets_processing_threads": 10,
+  "network_peer_blacklist_timeout" : 600,
```
kstdl, line 19: I am not sure, but maybe we should have a config migration for this. I believe it is already merged to develop.
But maybe it is better to do it in PR to testnet\mainnet branch.

mfrankovi: this is an optional parameter which will default to this value if it is missing so I don't think we need to have config migration

JakubFornadel: @kstdl what ? PR to testnet/mainnet branch ? why ?

JakubFornadel (approved):

libraries/cli/include/cli/default_config.hpp
mfrankovi: I would like to merge this before reseting the devnet so I am resolving this conversation. If this is something unresolved please create an issue for it

PR #1504, opened 2022-02-03
Bandwith period max stats

JakubFornadel:
## Purpose

- all packets stats logs transformed into the json strings
- period max bandwidth stats are returned also by get_packets_stats rpc call

### DIFF
--- a/libraries/cli/include/cli/default_config.hpp
+++ b/libraries/cli/include/cli/default_config.hpp
@@ -91,10 +91,6 @@ const char *default_json = R"foo({
           {
             "name": "SUMMARY",
             "verbosity": "INFO"
-          },
-          {
-            "name": "NETPER",
-            "verbosity": "DEBUG"
           }
         ],
         "outputs": [
--- a/libraries/cli/include/cli/default_config.hpp
+++ b/libraries/cli/include/cli/default_config.hpp
@@ -91,6 +91,10 @@ const char *default_json = R"foo({
           {
             "name": "SUMMARY",
             "verbosity": "INFO"
+          },
+          {
+            "name": "NETPER",
+            "verbosity": "DEBUG"
           }
         ],
         "outputs": [
--- a/libraries/core_libs/network/src/tarcap/stats/packets_stats.cpp
+++ b/libraries/core_libs/network/src/tarcap/stats/packets_stats.cpp
@@ -1,35 +1,56 @@
 #include "network/tarcap/stats/packets_stats.hpp"
 
+#include "json/writer.h"
+
 namespace taraxa::network::tarcap {
+
+/**
+ * @param value
+ * @return unstyled json string (without new lines and whitespaces).
+ */
+std::string jsonToUnstyledString(const Json::Value& value) {
+  Json::StreamWriterBuilder builder;
+  builder["indentation"] = "";  // assume default for comments is None
+  return Json::writeString(builder, value);
+}
 
 PacketsStats::PacketsStats(const addr_t& node_addr) { LOG_OBJECTS_CREATE("NETPER"); }
 
-void PacketsStats::addReceivedPacket(const dev::p2p::NodeID& node, const std::string& packet_type,
-                                     const SinglePacketStats& packet) {
+void PacketsStats::addReceivedPacket(const std::string& packet_type, const SinglePacketStats& packet) {
   received_packets_stats_.addPacket(packet_type, packet);
 
-  LOG(log_dg_) << "(\"" << node << "\") received " << packet_type << " packet from (\"" << packet.node_
-               << "\"). Stats: " << packet;
+  auto packet_json = packet.getStatsJson();
+  packet_json["type"] = packet_type;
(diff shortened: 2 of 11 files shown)


libraries/core_libs/network/src/tarcap/stats/packets_stats.cpp
```diff
@@ -1,35 +1,56 @@
...
-void PacketsStats::addReceivedPacket(const dev::p2p::NodeID& node, const std::string& packet_type,
-                                     const SinglePacketStats& packet) {
+void PacketsStats::addReceivedPacket(const std::string& packet_type, const SinglePacketStats& packet) {
   received_packets_stats_.addPacket(packet_type, packet);
 
-  LOG(log_dg_) << "(\"" << node << "\") received " << packet_type << " packet from (\"" << packet.node_
-               << "\"). Stats: " << packet;
+  auto packet_json = packet.getStatsJson();
+  packet_json["type"] = packet_type;
+  LOG(log_tr_) << "Received packet: " << jsonToUnstyledString(packet_json);
```
mfrankovi, line 24: If I understand correctly addReceivedPacket and addSentPacket are called for every received/sent packet, maybe we should check if this change hurts performance when receiving/sending large number of packets. Even if this log channel is off these json methods are called which might affect performance

JakubFornadel: this is called inside packet handlers, which are running & processing packets on separate threads, which means that printing out json representation of packet is basically negligible when compared to all the code that is processed inside specific packet handlers....

libraries/cli/include/cli/default_config.hpp
```diff
@@ -90,6 +90,10 @@ const char *default_json = R"foo({
           {
             "name": "SUMMARY",
             "verbosity": "INFO"
+          },
+          {
```
MatusKysel, line 94: it should not be part of default as it is just for testing

JakubFornadel: it is not testing - it is statistics collection, which would be definitely worth to see also on mainnet and compare it with other networks

MatusKysel (approved):

libraries/cli/include/cli/default_config.hpp
MatusKysel: ok, then this can not be part of console output at all, but as we discussed just file output, even for testnet. And maybe we should increase logging interval to few minutes, because 30s does not show anything

kstdl: I see it as a separate log file just with that data. So we can easily get it from users. If we will add new configuration(`logging.configurations`) with different output it should work I believe

JakubFornadel: hmm now when I look at it maybe it should not be logged at all and we will use simply rpc :) Just so you understand how this works. For example the node is running for 1 day, during this day I collect packets stats every 30 sec and I keep track of max values(max count/max size) for each packet per these 30 sec time periods. So if the node is running 1 day and I call rpc, I get the max from this 1 day, if it is running for 1 week and call rpc, I get max from this week. When I think about it we dont need to log this regularly as for example if we collect max stats (in terms of max size) at the node startup when it starts syncing, this max stats will be printed in every log after that and it wont change

JakubFornadel: @MatusKysel not sure if it is more clear now, but it does not matter that much if you use 30 sec, 5 sec or 2 minute period.... you simply need some period which will be then used also for bandwidth limits

MatusKysel: Ok then, I am voting for RPC call

libraries/core_libs/network/src/tarcap/stats/packets_stats.cpp
mfrankovi: I still think that there is a possibility that some minor impact on performance is possible. Considering the number of packets we might be receiving I would go with some binary struct and not using json internally to log this. There is json processing inside of packet.getStatsJson and then call to jsonToUnstyledString.

JakubFornadel: Ok I did spend some time on it and no you are wrong, json creation has negligible impact on performance. Simple string creation takes about 5 us, json string creation takes around 25 us. Then  just logging any string takes 50 us, so it is 75 us for logging json string vs 55 us for logging simple packet string.

mfrankovi: I wouldn't count in the 50us for logging. I assume we would not log this always. If we wanted a node with max performance this log would not be on but still this code would run for every packet even with the log channel off and then 5us vs 25us if we might have thousands of packets per second in my opinion is not that negligible. I don't think this will cause some serious performance issues but I think it would be more optimal to go with some native struct and not json for these logs.

JakubFornadel: json is removed

kstdl (approved):

mfrankovi (approved):

PR #1506, opened 2022-02-03
chore: pbft transaction ordering

### LINKED ISSUE #1482: Pbft block txs ordering
## Task Description

Even if we fix txs ordering based on fee & sender & nonce, dag could still shuffle dag blocks inside single pbft blocks so txs could be again ordered in a wrong way. 

##### Solution
Once the pbft block is created and dag blocks ordered, take the final ordering of txs in it and swap those based on sender & nonce or sender & fee ? Not sure which algorithm to choose but from implementation point of view it does not matter. 

## Epic Parent

MatusKysel:
fixes #1482

### DIFF
--- a/libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
+++ b/libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
@@ -40,7 +40,7 @@ bytes SortitionParamsChange::rlp() const {
   s << interval_efficiency;
   s << actual_correction_per_percent;
 
-  return s.out();
+  return s.invalidate();
 }
 
 SortitionParamsChange SortitionParamsChange::from_rlp(const dev::RLP& rlp) {
--- a/libraries/core_libs/consensus/src/final_chain/final_chain.cpp
+++ b/libraries/core_libs/consensus/src/final_chain/final_chain.cpp
@@ -14,9 +14,8 @@ enum DBMetaKeys { LAST_NUMBER = 1 };
 
 class FinalChainImpl final : public FinalChain {
   shared_ptr<DB> db_;
-  StateAPI state_api_;
-  // TODO re-enable it
   unique_ptr<ReplayProtectionService> replay_protection_service_;
+  StateAPI state_api_;
 
   mutable shared_mutex last_block_mu_;
   mutable shared_ptr<BlockHeader> last_block_;
@@ -38,6 +37,7 @@ class FinalChainImpl final : public FinalChain {
  public:
   FinalChainImpl(shared_ptr<DB> const& db, Config const& config, addr_t const& node_addr)
       : db_(db),
+        replay_protection_service_(NewReplayProtectionService({10}, db)),
         state_api_([this](auto n) { return block_hash(n).value_or(ZeroHash()); },  //
                    config.state,
                    {
--- a/libraries/core_libs/consensus/src/final_chain/final_chain.cpp
+++ b/libraries/core_libs/consensus/src/final_chain/final_chain.cpp
@@ -37,7 +37,7 @@ class FinalChainImpl final : public FinalChain {
  public:
   FinalChainImpl(shared_ptr<DB> const& db, Config const& config, addr_t const& node_addr)
       : db_(db),
-        replay_protection_service_(NewReplayProtectionService({10}, db)),
+        // replay_protection_service_(NewReplayProtectionService({}, db)),
         state_api_([this](auto n) { return block_hash(n).value_or(ZeroHash()); },  //
                    config.state,
                    {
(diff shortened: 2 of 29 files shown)


libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -40,7 +40,7 @@ bytes SortitionParamsChange::rlp() const {
   s << interval_efficiency;
   s << actual_correction_per_percent;
 
-  return s.out();
+  return s.invalidate();
```
gaoqi8649, line 43: invalidate() return std::move(m_out). What's the benefits using invalidate? I think that's same here. I have no objection here, just want to learn.

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -229,6 +230,27 @@ size_t TransactionManager::getNonfinalizedTrxSize() const {
...
+    ret.push_back(nonfinalized_transactions_in_dag_[hash]);
+  }
+  if (sorted) {
+    std::stable_sort(ret.begin(), ret.end(), [](const auto &t1, const auto &t2) {
+      if (t1->getSender() == t2->getSender()) {
+        return t1->getNonce() < t2->getNonce() ||
+               (t1->getNonce() == t2->getNonce() && t1->getGasPrice() > t2->getGasPrice());
+      }
+      return true;
+    });
```
gaoqi8649, line 249: I remember you have a similar transactions sorting method in another place in another PR. I think that's better to make the transactions sorting to a unique method. If people change the transactions sorting in a later time, that will  be very possible only change in one place and forget another one.

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -229,6 +230,27 @@ size_t TransactionManager::getNonfinalizedTrxSize() const {
   return nonfinalized_transactions_in_dag_.size();
 }
 
+std::vector<std::shared_ptr<Transaction>> TransactionManager::getNonfinalizedTrx(const std::vector<trx_hash_t> &hashes,
+                                                                                 bool sorted) {
+  std::vector<std::shared_ptr<Transaction>> ret;
+  ret.reserve(hashes.size());
+  std::shared_lock transactions_lock(transactions_mutex_);
+  for (const auto &hash : hashes) {
+    assert(nonfinalized_transactions_in_dag_.contains(hash));
```
JakubFornadel, line 239: this assert is suspicious, what if you dont have the transaction (we used to have missing tx)... You never know what happens on network so if you received pbft block, and you dont have all of the transactions - this will kill the node ?

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
MatusKysel: `out()` is just sending the reference to `s.m_out`  witch will probably creates copy on return. `invalidate()` should "steal" by moving `m_out`and return this data

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
MatusKysel: they are not exactly same and the other could change anytime (we could improve something), but this one is unchangeable as this is requirement to have correctly ordered transaction, where gasprice is important only if the nonce is same. For transaction dag ordering some validators, could include transaction with expensive gas, as they want to earn more on rewards

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
MatusKysel: If you do not have trxs, then dag block and it's transactions can not be insert. If this assert is hit, either I have missed some flow or we have a bug. We could change it to error, but I am not checking logs on devnet, so this way I will now sooner. So, if you want I can create task to change it to error log when we want to deploy it on testnet

libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -38,6 +37,7 @@ class FinalChainImpl final : public FinalChain {
  public:
   FinalChainImpl(shared_ptr<DB> const& db, Config const& config, addr_t const& node_addr)
       : db_(db),
+        // replay_protection_service_(NewReplayProtectionService({}, db)),
```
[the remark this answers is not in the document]
MatusKysel: Because it is still not working correctly, I have another task to enable it

gaoqi8649 (approved):

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
JakubFornadel: so did you go through the flow and make sure it is not possible or not ?

JakubFornadel: ok make sure it does not get to testnet(I am sure we forget about it so take pls as your responsibility to monitor this) before we test it properly on devnet then.

kstdl (approved):

PR #1529, opened 2022-02-10
Modify identify leader selection to match stakes

mfrankovi:


MatusKysel:
This looks good, maybe we should write some edge cases, e.g. if the token distribution is flat, extreme (one has like 50% of stake) ....

PR #1532, opened 2022-02-10
fix(sortition): fix sortition params oscillations by decreasing computation_interval

kstdl:
https://github.com/Taraxa-project/taraxa-node/issues/1519

It has a few changes after that we will be able to set computation interval less than changing_interval. 

[testnet_efficiencies.json](https://github.com/Taraxa-project/taraxa-node/compare/testnet...fix_oscillations?expand=1#diff-089f3c851f5a62bdb159fe3a6b36062b9df2da9e8e3bcbbd7b203d3bfc8fc705) file has data from the testnet with oscillations. I created a test that shows how lowering of computation interval will affect calculations. In the result(logs of the test and its check) you can see that sum of all changes with smaller computation interval is also much smaller because of more precise data


mfrankovi (commented):
Is it a plan to reset the testnet without making a hard for on the testnet?
Should we maybe have a PR for develop first and test it there before merging to testnet?

JakubFornadel:
> Is it a plan to reset the testnet without making a hard for on the testnet? Should we maybe have a PR for develop first and test it there before merging to testnet?

yeah not a good idea to merge it to the testnet just like that... I would say we definitely have to run it on develop first, together with @MatusKysel  changes related to producer selection and once it works fine. I would just reset testnet with new version.... no hardfork

JakubFornadel (approved):

JakubFornadel (changes requested):
base branch still not changed to develop

mfrankovi (approved):
I believe we agreed that we will put this on testnet first. Once we do that we will need a new PR to merge the change on develop as well

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -133,13 +135,14 @@ uint16_t SortitionParamsManager::averageCorrectionPerPercent() const {
...
-    params_changes_.pop_front();
+  auto efficiencies_to_leave = std::max(config_.computation_interval - config_.changing_interval, 0);
+  {
+    auto excess_count = std::max(static_cast<int32_t>(dag_efficiencies_.size()) - efficiencies_to_leave, 0);
+    dag_efficiencies_.erase(dag_efficiencies_.begin(), dag_efficiencies_.begin() + excess_count);
   }
-  while (dag_efficiencies_.size() > efficiencies_to_leave) {
-    dag_efficiencies_.pop_front();
+  {
+    auto excess_count = std::max(static_cast<int32_t>(params_changes_.size()) - config_.changes_count_for_average, 0);
```
MatusKysel, line 144: several variables could be `const`

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -122,6 +122,8 @@ uint16_t SortitionParamsManager::calculateDagEfficiency(const SyncBlock& block)
 }
 
 uint16_t SortitionParamsManager::averageDagEfficiency() {
+  auto excess_count = std::max(static_cast<int32_t>(dag_efficiencies_.size()) - config_.computation_interval, 0);
```
MatusKysel, line 125: const

tests/sortition_test.cpp
```diff
@@ -486,4 +487,75 @@ TEST_F(SortitionTest, get_params_from_period) {
   EXPECT_EQ(db->getParamsChangeForPeriod(95)->interval_efficiency, 90 * kOnePercent);
 }
 
+class SortitionParamsManagerTest : public SortitionParamsManager {
+ public:
+  SortitionParamsManagerTest(SortitionConfig sort_conf, std::shared_ptr<DbStorage> db)
+      : SortitionParamsManager(addr_t::random(), sort_conf, db) {}
```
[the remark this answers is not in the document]
MatusKysel: and db

JakubFornadel (approved):

PR #1521, opened 2022-02-08
simple test to simulate testnet block proposal

MatusKysel:
## Purpose
Introduce test to simulate what is happening on testnet

Results for 1000 rounds:
```
Node 0 produced 163 blocks
Node 1 produced 181 blocks
Node 2 produced 168 blocks
Node 3 produced 153 blocks
Node 4 produced 157 blocks
Node 5 produced 156 blocks
Node 19 produced 1 blocks
Node 31 produced 1 blocks
Node 32 produced 1 blocks
Node 34 produced 1 blocks
Node 55 produced 1 blocks
Node 62 produced 1 blocks
Node 69 produced 1 blocks
Node 92 produced 1 blocks
Node 94 produced 1 blocks
Node 95 produced 1 blocks
Node 135 produced 1 blocks
Node 163 produced 1 blocks
Node 190 produced 1 blocks
Node 203 produced 1 blocks
Node 253 produced 1 blocks
Node 271 produced 1 blocks
Node 278 produced 1 blocks
Node 289 produced 1 blocks
Node 315 produced 1 blocks
Node 360 produced 1 blocks
Node 372 produced 1 blocks
Node 404 produced 1 blocks
```

## How to run

`crypto_test --gtest_filter=*testnet*`

PR #1533, opened 2022-02-11
new leader selection

MatusKysel:
This PR introduces 2 things:

1. Protection against attack(comment from their forum)
```
Here we do something subtle but important: Instead of using the VRF output directly, we hash it with the voter’s address. 
This defends against the following attack: If I split my money into a large number of small accounts, I can register the same VRF
key on all the accounts. Then the accounts will always have the same VRF output – if one account is selected, they’ll all be 
selected, which will give me a huge number of votes. Binomial sampling isn’t enough to save us here – as Ignacio mentions in 
the “Sybil Attacks” section, we need each account to have independent VRF outputs. So we hash the address in with the VRF 
output, and the result is guaranteed to be independent for different accounts even if they have the same VRF key.
```

2. Leader selection based on Algorand(code comment from their code)
```
// lowestOutput is used for breaking ties when there are multiple proposals.
// People will vote for the proposal whose credential has the lowest lowestOutput.
//
// We hash the credential and interpret the output as a bigint.
// For credentials with weight w > 1, we hash the credential w times (with
// different counter values) and use the lowest output.
//
// This is because a weight w credential is simulating being selected to be on the
// leader committee w times, so each of the w proposals would have a different hash,
// and the lowest would win.
```


JakubFornadel:
Once you have it ready, it would nice if we have some sort of the test that selection is random ? Something like in [this PR](https://github.com/Taraxa-project/taraxa-node/pull/1521) ?

justinsnapp:
// We hash the credential and interpret the output as a bigint.
// For credentials with weight w > 1, we hash the credential w times (with
// different counter values) and use the lowest output.
//
// This is because a weight w credential is simulating being selected to be on the
// leader committee w times, so each of the w proposals would have a different hash,
// and the lowest would win.

This is a crucial part, makes good sense

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1094,6 +1096,18 @@ std::shared_ptr<Vote> PbftManager::generateVote(blk_hash_t const &blockhash, Pbf
...
+uint64_t PbftManager::getThreshold(PbftVoteTypes vote_type) const {
+  switch (vote_type) {
+    case propose_vote_type:
+      return std::min<uint64_t>(NUMBER_OF_PROPOSERS, getDposTotalVotesCount());
+    case soft_vote_type:
+    case cert_vote_type:
+    case next_vote_type:
+    default:
+      return sortition_threshold_;
+  }
```
gaoqi8649, line 1108: if (vote_type == propose_vote_type) {
  return std::min<uint64_t>(NUMBER_OF_PROPOSERS, getDposTotalVotesCount());
} else {
  return sortition_threshold_;
}

I think if else will be better

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -319,7 +320,8 @@ size_t PbftManager::dposEligibleVoteCount_(addr_t const &addr) {
 // Only used by RPC call
 uint64_t PbftManager::getVoteWeight(PbftVoteTypes type, uint64_t round, size_t step) const {
   VrfPbftSortition vrf_sortition(vrf_sk_, {type, round, step});
-  return vrf_sortition.getBinominalDistribution(weighted_votes_count_, getDposTotalVotesCount(), sortition_threshold_);
+  return vrf_sortition.calculateWeight(weighted_votes_count_, getDposTotalVotesCount(), getThreshold(type),
```
gaoqi8649, line 323: Why reset weight here? RPC call should not reset weight value

libraries/types/vote/include/vote/vote.hpp
```diff
@@ -46,6 +46,24 @@ struct VrfPbftMsg {
   size_t step;
 };
 
+struct HashableVrf {
+  HashableVrf(const vrf_wrapper::vrf_output_t& vrf, const public_t& addr, uint64_t i = 0)
```
gaoqi8649, line 50: i is stake, should never be 0

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
justinsnapp: @gaoqi8649  I disagree because I think the way @MatusKysel  has it here makes it more clear that we will in the future set different thresholds for different steps.   Were that not the plan I might agree with you @gaoqi8649

MatusKysel: Yes that's correct @justinsnapp. This is for later, because I have noticed that algorand has committee per type, so maybe we could also introduce something like that, but so far I have no knowledge why it is better than what we have right now

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1293,6 +1308,19 @@ std::vector<std::vector<uint>> PbftManager::createMockTrxSchedule(
...
 }
 
+h256 PbftManager::getProposal(const std::shared_ptr<Vote> &vote) const {
+  HashableVrf vrf_hash(vote->getCredential(), vote->getVoter(), 1);
+  auto lowest_hash = vrf_hash.getHash();
+  for (uint64_t i = 2; i <= vote->getWeight(); ++i) {
+    vrf_hash.iter = i;
+    const auto tmp_hash = vrf_hash.getHash();
+    if (lowest_hash > tmp_hash) {
+      lowest_hash = std::move(tmp_hash);
```
JakubFornadel, line 1318: std::move does not work with const. This will make a copy

libraries/cli/include/cli/default_config.hpp
```diff
@@ -166,6 +166,7 @@ const char *default_json = R"foo({
     },
     "pbft": {
       "committee_size": "0x3e8",
+      "number_of_proposers": "0x14",
```
JakubFornadel, line 169: from my perspective as someone who is just reviewing this PR, I did not know that `number_of_proposers` is basically a specific committee size for proposal round. Wouldn't it be more intuitive if we call it something like `propose_round_committee_size`, `proposing_committee_size` or some name that would also include committe_size in it ?

libraries/types/vote/src/vote.cpp
```diff
@@ -58,9 +58,12 @@ uint64_t VrfPbftSortition::getBinominalDistribution(uint64_t stake, double dpos_
...
 }
 
-uint64_t VrfPbftSortition::getBinominalDistribution(uint64_t stake, double dpos_total_votes_count,
-                                                    double threshold) const {
-  return VrfPbftSortition::getBinominalDistribution(stake, dpos_total_votes_count, threshold, output);
+uint64_t VrfPbftSortition::calculateWeight(uint64_t stake, double dpos_total_votes_count, double threshold,
+                                           const public_t& address) const {
+  // Also hash in the address. This is necessary to decorrelate the selection of different accounts that have the same
+  // VRF key.
+  HashableVrf hash_vrf(output, address);
```
JakubFornadel, line 65: How does this work when output is not declared here nor passed as argument ?

tests/crypto_test.cpp
```diff
@@ -337,14 +337,90 @@ TEST_F(CryptoTest, binomial_distribution) {
...
+    for (const auto& n : high_stake_nodes) {
+      VrfPbftSortition sortition(n.second, msg);
+      HashableVrf hash(sortition.output, n.second, 0);
+      if (auto stake = VrfPbftSortition::getBinominalDistribution(
+              high_stake_nodes_power, valid_sortition_players, committee_size, hash.getHash())) {
+        hash.iter = 1;
+        auto lowest = hash.getHash();
+        for (uint64_t j = 2; j <= stake; j++) {
+          hash.iter = j;
+          const auto tmp = hash.getHash();
```
JakubFornadel, line 381: remove const so it works with std::move

JakubFornadel (approved):

libraries/types/vote/src/vote.cpp
MatusKysel: output is member of class/struct

JakubFornadel: we could fix things like this if we touch it ? VrfPbftSortition should be class, not struct and all members should have `_`

mfrankovi: This output initially always confused me as well specially since it is a member from the base class. We should definitely add "_". Maybe also put VrfPbftSortition class into a separate file?

mfrankovi (approved):

libraries/types/vote/src/vote.cpp
MatusKysel: sure, I will refactor the while thing

Issue #1525, opened 2022-02-09
Remove "de2b..." hardcoded account from config


JakubFornadel:
There is hardocded account `de2b1203d72d3549ee2f733b00b2789414c7cea5` in  predefined config `ChainConfig::predefined_`. As a result, this account is now on every network, even mainnet.... Remove that:

- does it need to be done in hf ?

PR #1534, opened 2022-02-11
fix(config): remove excess hardcoded account

kstdl:
#1525

### DIFF
--- a/libraries/config/src/chain_config.cpp
+++ b/libraries/config/src/chain_config.cpp
@@ -29,6 +29,13 @@ void dec_json(Json::Value const& json, ChainConfig& obj) {
   dec_json(json["final_chain"], obj.final_chain);
 }
 
+const ChainConfig& ChainConfig::predefined(std::string const& name) {
+  if (auto i = predefined_->find(name); i != predefined_->end()) {
+    return i->second;
+  }
+  throw std::runtime_error("unknown chain config: " + name);
+}
+
 decltype(ChainConfig::predefined_) const ChainConfig::predefined_([] {
   decltype(ChainConfig::predefined_)::val_t cfgs;
   cfgs["default"] = [] {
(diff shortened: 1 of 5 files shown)


MatusKysel (approved):

libraries/config/src/chain_config.cpp
```diff
@@ -48,12 +48,9 @@ decltype(ChainConfig::predefined_) const ChainConfig::predefined_([] {
     cfg.final_chain.state.eth_chain_config.dao_fork_block = state_api::BlockNumberNIL;
     cfg.final_chain.state.execution_options.disable_nonce_check = true;
     cfg.final_chain.state.execution_options.disable_gas_fee = true;
-    addr_t root_node_addr("de2b1203d72d3549ee2f733b00b2789414c7cea5");
```
JakubFornadel, line 51: dont we need to change also genesis dag block sender from few lines above ? `"sender": "de2b1203d72d3549ee2f733b00b2789414c7cea5"`

kstdl: There could be any random address. And it is changing nothing as I see

JakubFornadel: I would say lets remove that particular account from everywhere, where it is not needed so it is not confusing

tests/util_test/util.hpp
```diff
@@ -120,6 +120,10 @@ inline auto make_node_cfgs(uint count) {
       return ret;
     }
     for (auto& cfg : ret) {
+      addr_t root_node_addr("de2b1203d72d3549ee2f733b00b2789414c7cea5");
+      cfg.chain.final_chain.state.genesis_balances[root_node_addr] = 9007199254740991;
+      auto& dpos = *cfg.chain.final_chain.state.dpos;
+      dpos.genesis_state[root_node_addr][root_node_addr] = dpos.eligibility_balance_threshold;
```
JakubFornadel, line 126: A better place to add this this is probably in chain_config.cpp:
```
cfgs["test"] = [&] {
    auto cfg = cfgs["default"];
    cfg.chain_id = 12345;
    cfg.final_chain.state.genesis_balances[addr_t("de2b1203d72d3549ee2f733b00b2789414c7cea5")] =
        u256(7200999050) * 10000000000000000;  // https://ethereum.stackexchange.com/a/74832
    return cfg;
  }();
```

mfrankovi (approved):

libraries/config/src/chain_config.cpp
kstdl: I replaced it with new random address

JakubFornadel (approved):

PR #1539, opened 2022-02-14
chore: increase test threshold because it failed sometimes

### LINKED ISSUE #1498: CryptoTest.new_sortition_rate failing on develop
When running CryptoTest.new_sortition_rate in a loop, sooner or later it always fails with:

```
/home/fornadel/workspace/taraxa/taraxa-node/tests/crypto_test.cpp:281: Failure
Expected: (d) <= (maxd), actual: 267 vs 250
wanted 10000 selections but got 9733, d=267, maxd=250
[  FAILED  ] CryptoTest.new_sortition_rate (169 ms)
[----------] 1 test from CryptoTest (169 ms total)
```

This Task will...

MatusKysel:
## Purpose
Fixes #1498

### DIFF
--- a/tests/crypto_test.cpp
+++ b/tests/crypto_test.cpp
@@ -352,8 +352,8 @@ TEST_F(CryptoTest, leader_selection) {
 
   std::srand(std::time(nullptr));
   const uint64_t committee_size = 20;
-  const uint64_t rounds = 500;
-  const uint64_t low_stake_nodes_num = 100;
+  const uint64_t rounds = 1000;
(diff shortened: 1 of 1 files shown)


MatusKysel:
> I wonder how do you set the 4% as the difference? what's depend on?

I think with change of calculating distribution based on output (old way) that was 512bit and now hash of the output + address that is 256bit. So this change affected this specific test, but of course if we increase number of rounds we will get better numbers

mfrankovi (approved):

PR #1540, opened 2022-02-14
feat: new generic architecture for processing packets rlp that would …

JakubFornadel:
## Purpose

This is a new generic architecture that would allow us to parse packet rlp. I have created POC and adjusted TransactionPacketHandler according to this new design, but it would take multple days to transform all other packet handlers...

Pros of such desing:

- no need to parse packets manually
- automatic ddos protection - there is a struct representation of each packet rlp and when it is parsed, number of rlp items (and types) must be equal to number of this struct variables (and types)  
- rlp parsing at 1 place. Currently almost all types that we use (e.g. Transaction, DagBlock, etc...) must support constructors with rlp as argument because we create them from rlp -> this would be eliminated as each packet would be automatically parsed to its structure representation before it is even sent to the specific packet handler...  

Cons (all the cons are basically related tot he amount of work that would be required for such change)

- each packet handler must be a separate class (now we have multiple packets handlers in 1 class)
- some packets rlp's would have to be changed so it can be parsed automatically. For example if we would like to parse struct

```
struct DagsPacket {
  struct DagBlock {
    hash_type hash;
    std::vector<Trsansation> transactions;
  }

  std::vector<DagBlock> dags;
}
```

each struct must be represented by a list inside rlp, which we dont enforce at the moment... Rlp for this example would look like this:
`[  [dag1_hash, [dag1_tx1, dag1_tx2]],  [dag2_hash, [dag2_tx1, dag2_tx2]] ]`

- rlp numeric fields would be almost always copied. Our internal types (e.g. Transaction, DagBlock, etc...) would not support constructors with rlp bytes anymore, but they would have to support constructors with packet rlp struct representation (e.g. DagsPacket from above) because these constructors usually dont just set parsed rlp data, they also process them somehow and then set other class member based on these values and this is simply impossible to do automatically (see for example Transaction ctor). 

Here is simplified new design:
```
// IPacketHandler.hpp
class IPacketHandler {
public:
    virtual ~IPacketHandler() = default;
    virtual void processPacket() = 0;
};


// PacketHandler.hpp
template<typename PacketData>
class PacketHandler : public IPacketHandler {
public:
    virtual ~PacketHandler() = default;

    void processPacket() override {
        auto packet_data = parsePacketRlp();
        process(std::move(packet_data));
    }

    virtual PacketData parsePacketRlp(const RawPacketData& raw_packet_data) {
      // here will be simply called 
      // return util::encoding_rlp::rlp_dec<PacketData>(util::RLPDecoderRef(raw_packet_data.rlp_, true));
    };

    virtual void process(PacketData&& packet_data) = 0;
};



// StatusPacketHandler.hpp
class StatusPacketData {
public:
    std::string msg_;
    
   RLP_FIELDS_DEFINE_INPLACE(msg_) // this is enough for automatic parsing
};

class StatusPacketHandler : public PacketHandler<StatusPacketData> {
public:
    ~StatusPacketHandler() = default;

    StatusPacketData parsePacketRlp(const RawPacketData& raw_packet_data) override {
        // In case it is needed, automatic parsing can be overridden here
    }

    void process(StatusPacketData&& packet_data) override {
        std::cout << "StatusPacketHandler processing: " << packet_data.msg_ << std::endl;
    }
};


main.cpp

std::shared_ptr<IPacketHandler> handler = std::make_shared<StatusPacketHandler>();
handler.processPacket(rlp);

```

### DIFF
--- a/libraries/common/include/common/encoding_rlp.hpp
+++ b/libraries/common/include/common/encoding_rlp.hpp
@@ -139,6 +139,15 @@ auto rlp(RLPDecoderRef encoding, Map& target) -> decltype(target[target.begin()-
   }
 }
 
+struct InvalidEncodingSize : std::invalid_argument {
+  uint expected, actual;
+
+  InvalidEncodingSize(uint expected, uint actual)
+      : invalid_argument(fmt("Invalid rlp list size; expected: %s, actual: %s", expected, actual)),
+        expected(expected),
+        actual(actual) {}
+};
+
 template <typename Param, typename... Params>
 void __dec_rlp_tuple_body__(RLP::iterator& i, RLP::iterator const& end, RLP::Strictness strictness, Param& target,
                             Params&... rest) {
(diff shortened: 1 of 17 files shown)


MatusKysel:
I do not understand second point in Cons: 
```
struct DagsPacket {
  struct DagBlock {
    hash_type hash;
    std::vector<Trsansation> transactions;
  }

  std::vector<DagBlock> dags;
}
```
Why do we need such a complicated structure? why it can not be like it is right now
```
struct DagsPacket {
 //some period and other things .... 
  std::vector<Trsansation> transactions;
  std::vector<DagBlock> dags;
}
```

Anyway except that, I believe we should enforce this as soon as possible. And also @JakubFornadel why do you think this needs to be a breaking change? I do not see why this can not work with current way of sending blocks

MatusKysel (approved):

JakubFornadel:
> I like this idea, but I see that could over complicated our code base. Maybe we should just leave it the backlog and lower the prio

I agree, I think I will create a simplified version tomorrow where I just check the rlp format in each packet process() function and thats it

Issue #1323, opened 2021-12-03
Strict gossip rules - validation in handlers 


JakubFornadel:
## Is your feature request related to a problem? Please describe.

We are accepting any data(that can be parsed from rlp) and saving it into the unverified_queue. 
## Describe the solution you'd like.

Verify block as soon as received inside packet handler and push only verified blocks into the verified queue.  Whole threadpool verifying blocks can be delete, unverified_queue can be deleted, ....

gaoqi8649:
The current design that use 4 threads for verify DAG blocks in unverified_queue. If move DAG blocks verification into packet handler, I think that will effect network packet handler performance

JakubFornadel:
> The current design that use 4 threads for verify DAG blocks in unverified_queue. If move DAG blocks verification into packet handler, I think that will effect network packet handler performance

packets handlers are inside network threadpool as well. Big advantage of such design is that you dont need other trheadpools like you just mentioned anymore...

PR #1547, opened 2022-02-15
chore: move dag validation to packet handler

MatusKysel:
## Purpose
fixes: #1323

### DIFF
--- a/libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
@@ -5,12 +5,10 @@
 namespace taraxa {
 
 DagBlockManager::DagBlockManager(addr_t node_addr, SortitionConfig const &sortition_config,
-                                 std::optional<state_api::DPOSConfig> dpos_config, unsigned num_verifiers,
                                  std::shared_ptr<DbStorage> db, std::shared_ptr<TransactionManager> trx_mgr,
                                  std::shared_ptr<FinalChain> final_chain, std::shared_ptr<PbftChain> pbft_chain,
                                  logger::Logger log_time, uint32_t queue_limit)
-    : num_verifiers_(num_verifiers),
-      db_(db),
+    : db_(db),
       trx_mgr_(trx_mgr),
       final_chain_(final_chain),
       pbft_chain_(pbft_chain),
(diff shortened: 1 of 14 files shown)


libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
```diff
@@ -192,7 +151,7 @@ std::optional<DagBlock> DagBlockManager::popVerifiedBlock(bool level_limit, uint
       cond_for_verified_qu_.wait(lock);
     }
   }
-  if (stopped_) return std::nullopt;
+  if (stopped_) return {};
```
mfrankovi, line 154: Since there are no more threads in this class maybe we should remove stopped_ variable completely. This is the only place it is used and it could be replaced with the check if the queue is empty. I think it would be cleaner if we actually do always check if the queue is not empty before popping from the queue

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
```diff
@@ -206,97 +165,59 @@ void DagBlockManager::pushVerifiedBlock(DagBlock const &blk) {
...
-                   << " level failed on VDF verification with pivot hash " << blk.getPivot() << " reason " << e.what();
-      LOG(log_er_) << "period from map: " << propose_period.first << " current: " << pbft_chain_->getPbftChainSize();
-      markBlockInvalid(block_hash);
-      continue;
-    }
+  auto propose_period = getProposalPeriod(blk.getLevel());
+  // Verify DPOS
+  if (!propose_period.second) {
+    // Cannot find the proposal period in DB yet. The slow node gets an ahead block, remove from seen_blocks
+    LOG(log_er_) << "Cannot find proposal period " << propose_period.first << " in DB for DAG block " << blk.getHash();
```
mfrankovi, line 189: I would not use error log here, this is something that can happen sometimes and you could have a lot of these logs all at once when node just joins the network and it is way behind but has not started syncing yet and then community might complain about it

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
MatusKysel: we can not remove `stopped_` variable, check the code above. `popVerifiedBlock` is called from dag_manager in threads and it is waiting for new block, so this has to stay there until we refactor design of dag_manager. FYI: I wanted to remove it

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
MatusKysel: so warn level?

kstdl (approved):

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
mfrankovi: I would go with info, I see it was info before your change but warn is fine as well

PR #1549, opened 2022-02-16
feat: light node

mfrankovi:
Implements light node functionality by deleting old data from period_data column which takes most of the disk space in the database.

light-node-history parameter in the config file defines how much period_data history is kept in the database. The rest is deleted.

As part of this a new feature of limiting inclusion of old/expired dag blocks into pbft is implemented.

### DIFF
--- a/libraries/cli/include/cli/default_config.hpp
+++ b/libraries/cli/include/cli/default_config.hpp
@@ -16,6 +16,7 @@ const char *default_json = R"foo({
   "network_sync_level_size": 25,
   "network_packets_processing_threads": 14,
   "network_peer_blacklist_timeout" : 600,
+  "is_light_node" : false,
   "deep_syncing_threshold" : 10,
   "network_boot_nodes": [
   ],
--- a/libraries/config/include/config/config.hpp
+++ b/libraries/config/include/config/config.hpp
@@ -75,6 +75,8 @@ struct TestParamsConfig {
 };
 
 struct FullNodeConfig {
+  static const uint32_t kDefaultLightNodeHistoryDays = 7;
+
   FullNodeConfig() = default;
   // The reason of using Json::Value as a union is that in the tests
   // there are attempts to pass char const* to this constructor, which
@@ -96,6 +98,8 @@ struct FullNodeConfig {
   ChainConfig chain = ChainConfig::predefined();
   state_api::Opts opts_final_chain;
   std::vector<logger::Config> log_configs;
+  bool is_light_node = false;       // Is light node
+  uint64_t light_node_history = 0;  // Number of periods to keep in history for a light node
 
   auto net_file_path() const { return data_path / "net"; }
 
--- a/libraries/config/src/config.cpp
+++ b/libraries/config/src/config.cpp
@@ -238,6 +238,16 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value 
     chain = ChainConfig::predefined();
   }
 
+  is_light_node = getConfigDataAsBoolean(root, {"is_light_node"}, true, false);
+  if (is_light_node) {
+    uint64_t min_light_node_history =
+        (uint64_t)kDefaultLightNodeHistoryDays * 24 * 60 * 60000 / (chain.pbft.lambda_ms_min * 6);
+    light_node_history = getConfigDataAsUInt(root, {"light_node_history"}, true, min_light_node_history);
+    if (light_node_history < min_light_node_history) {
+      light_node_history = min_light_node_history;
+    }
+  }
+
   node_secret = wallet["node_secret"].asString();
   vrf_secret = vrf_wrapper::vrf_sk_t(wallet["vrf_secret"].asString());
 
--- a/libraries/core_libs/consensus/src/dag/dag.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag.cpp
@@ -420,6 +420,16 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
           LOG(log_dg_) << "Block already in DB: " << blk.getHash();
           return;
         }
+        const auto proposal_period = db_->getProposalPeriodForDagLevel(blk.getLevel());
+        const auto expiry_period = pbft_chain_->getDagExpiryPeriod();
+        assert(proposal_period);
+        if (*proposal_period < expiry_period) {
+          LOG(log_nf_) << "Dropping old block: " << blk.getHash() << ". Proposal period: " << *proposal_period
+                       << ". Current period: " << period_ << " Expiry period: " << expiry_period
+                       << ". Block level: " << blk.getLevel();
+          return;
+        }
+
         // Saves transactions and remove them from memory pool
         trx_mgr_->saveTransactionsFromDagBlock(trxs);
         // Save the dag block
@@ -434,12 +444,7 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
 
       addToDag(blk_hash, pivot_hash, tips, blk.getLevel());
 
-      auto [p, ts] = getFrontier();
-      frontier_.pivot = p;
-      frontier_.tips.clear();
-      for (auto const &t : ts) {
-        frontier_.tips.push_back(t);
-      }
+      updateFrontier();
     }
     if (save) {
       block_verified_.emit(blk);
@@ -493,6 +498,15 @@ std::pair<blk_hash_t, std::vector<blk_hash_t>> DagManager::getFrontier() const {
     tips.erase(end, tips.end());
   }
   return {pivot, tips};
+}
+
+void DagManager::updateFrontier() {
+  auto [p, ts] = getFrontier();
+  frontier_.pivot = p;
+  frontier_.tips.clear();
+  for (auto const &t : ts) {
+    frontier_.tips.push_back(t);
+  }
 }
 
 void DagManager::getGhostPath(blk_hash_t const &source, std::vector<blk_hash_t> &ghost) const {
@@ -584,23 +598,91 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
   assert(dag_order_set.count(new_anchor));
   addToDag(new_anchor, blk_hash_t(), vec_blk_t(), 0, true);
 
+  std::unordered_map<blk_hash_t, std::shared_ptr<DagBlock>> expired_dag_blocks_to_remove;
+  std::vector<trx_hash_t> expired_dag_blocks_transactions;
+
   for (auto &v : non_finalized_blocks) {
-    for (auto &blk : v.second) {
-      if (dag_order_set.count(blk) == 0) {
-        auto dag_block = dag_blk_mgr_->getDagBlock(blk);
+    for (auto &blk_hash : v.second) {
+      if (dag_order_set.count(blk_hash) == 0) {
+        auto dag_block = dag_blk_mgr_->getDagBlock(blk_hash);
         auto pivot_hash = dag_block->getPivot();
-        addToDag(blk, pivot_hash, dag_block->getTips(), dag_block->getLevel(), false);
+
+        if (validateBlockNotExpired(dag_block, expired_dag_blocks_to_remove)) {
+          addToDag(blk_hash, pivot_hash, dag_block->getTips(), dag_block->getLevel(), false);
+        } else {
+          for (const auto &trx : dag_block->getTrxs()) expired_dag_blocks_transactions.emplace_back(trx);
+        }
       }
     }
+  }
+
+  // Remove any transactions from expired dag blocks if not already finalized or included in another dag block
+  if (expired_dag_blocks_transactions.size()) {
+    handleExpiredDagBlocksTransactions(expired_dag_blocks_transactions);
   }
 
   old_anchor_ = anchor_;
   anchor_ = new_anchor;
   period_ = period;
+  updateFrontier();
 
   LOG(log_nf_) << "Set new period " << period << " with anchor " << new_anchor;
 
   return dag_order_set.size();
+}
+
+void DagManager::handleExpiredDagBlocksTransactions(
+    const std::vector<trx_hash_t> &transactions_from_expired_dag_blocks) const {
+  std::unordered_set<trx_hash_t> transactions_from_expired_dag_blocks_to_remove;
+  // Only transactions that need to be moved to memory pool are the ones that are not yet finalized and are not part of
+  // another valid DAG block
+  auto trxs_finalized = db_->transactionsFinalized(transactions_from_expired_dag_blocks);
+  for (uint32_t i = 0; i < trxs_finalized.size(); i++) {
+    if (!trxs_finalized[i]) {
+      transactions_from_expired_dag_blocks_to_remove.emplace(transactions_from_expired_dag_blocks[i]);
+    }
+  }
+  for (auto const &level : non_finalized_blks_) {
+    for (auto const &blk : level.second) {
+      auto dag_block = dag_blk_mgr_->getDagBlock(blk);
+      for (auto const &trx : dag_block->getTrxs()) {
+        transactions_from_expired_dag_blocks_to_remove.erase(trx);
+      }
+    }
+  }
+  if (transactions_from_expired_dag_blocks_to_remove.size() > 0) {
+    trx_mgr_->moveNonFinalizedTransactionsToTransactionsPool(std::move(transactions_from_expired_dag_blocks_to_remove));
+  }
+}
+
+bool DagManager::validateBlockNotExpired(
+    const std::shared_ptr<DagBlock> &dag_block,
+    std::unordered_map<blk_hash_t, std::shared_ptr<DagBlock>> &expired_dag_blocks_to_remove) {
+  // Check for expired dag blocks, in practice this should happen very rarely if some node is cut off from the rest of
+  // the network. In normal cases DAG blocks will be finalized or an old dag block will not enter the DAG at all
+  const auto &blk_hash = dag_block->getHash();
+  bool block_points_to_expired_block = expired_dag_blocks_to_remove.contains(dag_block->getPivot());
+  if (!block_points_to_expired_block) {
+    for (const auto &tip : dag_block->getTips()) {
+      if (expired_dag_blocks_to_remove.contains(tip)) {
+        block_points_to_expired_block = true;
+        break;
+      }
+    }
+  }
+
+  const auto proposal_period = db_->getProposalPeriodForDagLevel(dag_block->getLevel());
+  assert(proposal_period);
+  const auto expiry_period = pbft_chain_->getDagExpiryPeriod();
+  if (block_points_to_expired_block || *proposal_period < expiry_period) {
+    LOG(log_nf_) << "Dropping expired block in setDagBlockOrder: " << blk_hash
+                 << ". Proposal period: " << *proposal_period << ". Current period: " << period_
+                 << ". Expiry period: " << expiry_period << ". Block level: " << dag_block->getLevel();
+    expired_dag_blocks_to_remove[blk_hash] = dag_block;
+    assert(blk_hash != frontier_.pivot);
+    return false;
+  }
+  return true;
 }
 
 void DagManager::recoverDag() {
--- a/libraries/core_libs/consensus/src/pbft/pbft_chain.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_chain.cpp
@@ -14,6 +14,7 @@ PbftChain::PbftChain(blk_hash_t const& dag_genesis_hash, addr_t node_addr, std::
       dag_genesis_hash_(dag_genesis_hash),
       size_(0),
       non_empty_size_(0),
+      dag_expiry_period_(0),
       last_pbft_block_hash_(blk_hash_t(0)),
       db_(move(db)) {
   LOG_OBJECTS_CREATE("PBFT_CHAIN");
@@ -31,6 +32,7 @@ PbftChain::PbftChain(blk_hash_t const& dag_genesis_hash, addr_t node_addr, std::
   head_hash_ = blk_hash_t(doc["head_hash"].asString());
   size_ = doc["size"].asUInt64();
   non_empty_size_ = doc["non_empty_size"].asUInt64();
+  dag_expiry_period_ = doc["dag_expiry_period"].asUInt64();
   last_pbft_block_hash_ = blk_hash_t(doc["last_pbft_block_hash"].asString());
   auto dag_genesis_hash_db = blk_hash_t(doc["dag_genesis_hash"].asString());
   assert(dag_genesis_hash_ == dag_genesis_hash_db);
@@ -50,6 +52,11 @@ uint64_t PbftChain::getPbftChainSize() const {
 uint64_t PbftChain::getPbftChainSizeExcludingEmptyPbftBlocks() const {
   SharedLock lock(chain_head_access_);
   return non_empty_size_;
+}
+
+uint64_t PbftChain::getDagExpiryPeriod() const {
+  SharedLock lock(chain_head_access_);
+  return dag_expiry_period_;
 }
 
 blk_hash_t PbftChain::getLastPbftBlockHash() const {
@@ -107,6 +114,9 @@ void PbftChain::updatePbftChain(blk_hash_t const& pbft_block_hash, bool null_anc
   size_++;
   if (!null_anchor) {
     non_empty_size_++;
+    if ((dag_expiry_period_ == 0 && non_empty_size_ > kDagExpiryPeriodLimit) || dag_expiry_period_ > 0) {
+      dag_expiry_period_++;
+    }
   }
   last_pbft_block_hash_ = pbft_block_hash;
 }
@@ -182,6 +192,7 @@ std::string PbftChain::getJsonStr() const {
   json["dag_genesis_hash"] = dag_genesis_hash_.toString();
   json["size"] = (Json::Value::UInt64)size_;
   json["non_empty_size"] = (Json::Value::UInt64)non_empty_size_;
+  json["dag_expiry_period"] = (Json::Value::UInt64)dag_expiry_period_;
   json["last_pbft_block_hash"] = last_pbft_block_hash_.toString();
   return json.toStyledString();
 }
@@ -197,6 +208,7 @@ std::string PbftChain::getJsonStrForBlock(blk_hash_t const& block_hash, bool nul
     non_empty_size++;
   }
   json["non_empty_size"] = (Json::Value::UInt64)non_empty_size;
+  json["dag_expiry_period"] = (Json::Value::UInt64)dag_expiry_period_;
   json["last_pbft_block_hash"] = block_hash.toString();
   return json.toStyledString();
 }
--- a/libraries/core_libs/network/rpc/Test.cpp
+++ b/libraries/core_libs/network/rpc/Test.cpp
@@ -47,11 +47,17 @@ Json::Value Test::get_dag_block(const Json::Value &param1) {
   try {
     if (auto node = full_node_.lock()) {
       blk_hash_t hash = blk_hash_t(param1["hash"].asString());
-      auto blk = node->getDagBlockManager()->getDagBlock(hash);
-      if (!blk) {
-        res = "Block not available \n";
+      const auto &dag_blk_mgr = node->getDagBlockManager();
+      auto known_block = dag_blk_mgr->isDagBlockKnown(hash);
+      if (known_block) {
+        auto blk = dag_blk_mgr->getDagBlock(hash);
+        if (!blk) {
+          res["status"] = "Light node - Block could not be retrieved";
+        } else {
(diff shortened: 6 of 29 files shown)


libraries/cli/include/cli/default_config.hpp
```diff
@@ -16,6 +16,8 @@ const char *default_json = R"foo({
   "network_sync_level_size": 10,
   "network_packets_processing_threads": 14,
   "network_peer_blacklist_timeout" : 600,
+  "network_peer_light_node_timeout" : 600,
+  "light-node-history" : 0,
```
MatusKysel, line 20: naming

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -418,6 +418,14 @@ void DagManager::addDagBlock(DagBlock const &blk, SharedTransactions &&trxs, boo
           LOG(log_dg_) << "Block already in DB: " << blk.getHash();
           return;
         }
+        const auto proposal_period = dag_blk_mgr_->getProposalPeriod(blk.getLevel());
+        assert(proposal_period.second);
+        if (proposal_period.first + FullNodeConfig::kDagExpiryPeriodLimit < period_) {
```
MatusKysel, line 423: same check could be part of `verifyBlock` as we do not need to verify block if it is too old

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -576,25 +588,89 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
...
-        auto dag_block = dag_blk_mgr_->getDagBlock(blk);
+    for (auto &blk_hash : v.second) {
+      if (dag_order_set.count(blk_hash) == 0) {
+        auto dag_block = dag_blk_mgr_->getDagBlock(blk_hash);
         auto pivot_hash = dag_block->getPivot();
-        addToDag(blk, pivot_hash, dag_block->getTips(), dag_block->getLevel(), false);
+
+        if (validateBlockNotExpired(dag_block, expired_dag_blocks_to_remove)) {
+          addToDag(blk_hash, pivot_hash, dag_block->getTips(), dag_block->getLevel(), false);
+        }
```
MatusKysel, line 601: I guess this should be only called when we are light node

libraries/core_libs/network/rpc/Test.cpp
```diff
@@ -47,11 +47,17 @@ Json::Value Test::get_dag_block(const Json::Value &param1) {
...
     if (auto node = full_node_.lock()) {
       blk_hash_t hash = blk_hash_t(param1["hash"].asString());
-      auto blk = node->getDagBlockManager()->getDagBlock(hash);
-      if (!blk) {
-        res = "Block not available \n";
+      const auto &dag_blk_mgr = node->getDagBlockManager();
+      auto known_block = dag_blk_mgr->isDagBlockKnown(hash);
+      if (known_block) {
+        auto blk = dag_blk_mgr->getDagBlock(hash);
+        if (!blk) {
```
MatusKysel, line 54: I do not understand this change, why not call `getDagBlock` as `isDagBlockKnown` calls inside `getDagBlock` so you basically calling it 2x

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -51,6 +52,17 @@ void StatusPacketHandler::process(const PacketData& packet_data, const std::shar
     auto const node_major_version = (*it++).toInt<unsigned>();
     auto const node_minor_version = (*it++).toInt<unsigned>();
     auto const node_patch_version = (*it++).toInt<unsigned>();
+    uint64_t node_history = (*it++).toInt<uint64_t>();
```
MatusKysel, line 55: const

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -51,6 +52,17 @@ void StatusPacketHandler::process(const PacketData& packet_data, const std::shar
...
+
+    // If this is a light node and it cannot serve our sync request disconnect from it
+    if (node_history > 0) {
+      if (pbft_synced_period + node_history < peer_pbft_chain_size) {
+        LOG(log_nf_) << "Light node is not able to serve our syncing request. " << packet_data.from_node_id_.abridged()
+                     << " peer will be disconnected";
+        peers_state_->set_peer_light_node(peer->getId());
+        disconnect(peer->getId(), dev::p2p::UserReason);
```
MatusKysel, line 63: do we need to disconnect? I guess it is useless for us, but maybe we could still gossip transactions via RPC back to the network. I do not know what's right

libraries/core_libs/storage/include/storage/storage.hpp
```diff
@@ -148,6 +148,7 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
   bool snapshot_enable_ = true;
   uint32_t db_max_snapshots_ = 0;
   std::set<uint64_t> snapshots_;
+  uint64_t light_node_history_ = 0;
```
MatusKysel, line 151: const

tests/full_node_test.cpp
```diff
@@ -955,12 +955,14 @@ TEST_F(FullNodeTest, sync_two_nodes1) {
   auto num_vertices1 = nodes[0]->getDagManager()->getNumVerticesInDag();
   auto num_vertices2 = nodes[1]->getDagManager()->getNumVerticesInDag();
   for (unsigned i = 0; i < SYNC_TIMEOUT; i++) {
-    if (num_vertices1.first > 3 && num_vertices2.first > 3 && num_vertices1 == num_vertices2) break;
+    if (nodes[0]->getTransactionManager()->getTransactionPoolSize() == 0 &&
+        nodes[1]->getTransactionManager()->getTransactionPoolSize() == 0 && num_vertices1 == num_vertices2)
+      break;
```
MatusKysel, line 960: is this related to your changes?

MatusKysel (approved):
Maybe you could also create some simple test suite, like we have fullnode -> lightnode? where we could check if everything works consensus wise and deleting old data

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: I think it is better that it is here under the DagManager mutex. Otherwise there could be a race condition that when we verify block it is not old but when it is inserted to DAG period changed and it could be old.

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: No, this is not related to light node functionality. This is about expiring old DAG blocks. It applies both for light node and full node

libraries/core_libs/network/rpc/Test.cpp
mfrankovi: I am not sure how to handle the rpc calls for light node. In current implementation if in a light node if dag block is in the part of period_data that we deleted isDagBlockKnown will return true but getDagBlock will return a null value because the period_data is missing. I am not sure if this is the best way to go.

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
mfrankovi: If we are out of sync I think we need to disconnect which will trigger us connecting to other nodes. I want to avoid the state where we are connected only to light nodes

libraries/core_libs/storage/include/storage/storage.hpp
mfrankovi: Fixed

tests/full_node_test.cpp
mfrankovi: I am not sure what triggered this tests to not pass but I got intermittent failures here where vertices count was 3. I don't understand why we had it to be over count of 3. I replaced this with checking that the transactions pool is empty which means that all the inserted transactions made it to DAG.

libraries/config/include/config/config.hpp
```diff
@@ -71,6 +73,10 @@ struct TestParamsConfig {
 };
 
 struct FullNodeConfig {
+  static const uint32_t kDagExpiryPeriodLimit = 1000;  // Any non finalized dag block with a propose period smaller by
```
gaoqi8649, line 76: I have a concern on this design. After introduce PBFT blocks with NULL anchor, we may have lots of empty PBFT blocks when there is no transactions generated. The proposal period only increase when finalizing DAG blocks. When generate empty PBFT blocks, proposal period will not increase but finalized period will keep increase. So the proposal period will be much behind the finalized period

libraries/cli/include/cli/default_config.hpp
gaoqi8649: I think "light node history" should be a const number and is not changeable. 
Maybe you want add in config for testing to adjust a good number, and remove it later.

libraries/config/include/config/config.hpp
mfrankovi: Yes, true, I will need to make changes not to count empty pbft blocks.

libraries/core_libs/consensus/src/dag/dag.cpp
MatusKysel: but it could be in both. This way we could save some CPU if we receive old blocks

libraries/config/src/config.cpp
```diff
@@ -144,6 +146,14 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value
     network.network_boot_nodes.push_back(node);
   }
 
+  light_node_history = getConfigDataAsUInt(root, {"light_node_history"}, true, 0);
+
+  // Min history for light node is 110% of kDagExpiryPeriodLimit to avoid any race condition around the
+  // kDagExpiryPeriodLimit
+  if (light_node_history > 0 && light_node_history < kDagExpiryPeriodLimit * 1.1) {
+    light_node_history = kDagExpiryPeriodLimit * 1.1;
+  }
```
gaoqi8649, line 155: I think it's better to make kDagExpiryPeriodLimit to 1100.
Then:
if (light_node_history != 0) {
  light_node_history = std::min(light_node_history, kDagExpiryPeriodLimit);
}

libraries/core_libs/network/src/tarcap/shared_states/peers_state.cpp
```diff
@@ -134,4 +134,31 @@ bool PeersState::is_peer_malicious(const dev::p2p::NodeID& peer_id) {
   return false;
 }
 
+void PeersState::set_peer_light_node(const dev::p2p::NodeID& peer_id) {
+  light_node_peers_.emplace(peer_id, std::chrono::steady_clock::now());
+}
+
+bool PeersState::is_peer_light_node(const dev::p2p::NodeID& peer_id) {
+  // Peers are marked light nodes for the time defined in conf_.network_peer_light_node_timeout
```
gaoqi8649, line 142: I don't understand why define light node peers in timing. 
In this way, some light nodes will be treat as light nodes, and some will be not. 
What is the purpose for that?

libraries/core_libs/storage/include/storage/storage.hpp
```diff
@@ -94,7 +94,7 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
     // do not change/move
     COLUMN(default_column);
     // Contains full data for an executed PBFT block including PBFT block, cert votes, dag blocks and transactions
-    COLUMN(period_data);
+    COLUMN_W_COMP(period_data, getIntComparator<uint64_t>());
```
gaoqi8649, line 97: I guess the COLUMN_W_COMP will do saving in order in DB. Am i right?

libraries/core_libs/network/src/tarcap/shared_states/peers_state.cpp
mfrankovi: it is possible for a node to change its state and go from light node back to full node

libraries/core_libs/storage/include/storage/storage.hpp
mfrankovi: Yes. This is needed for DeleteRange method in the db to work correctly.

libraries/cli/include/cli/default_config.hpp
```diff
@@ -16,6 +16,8 @@ const char *default_json = R"foo({
   "network_sync_level_size": 25,
```
JakubFornadel, line 16: we have now also mainnet_config.hpp, which should be propably adjusted too

libraries/config/src/config.cpp
```diff
@@ -144,6 +146,14 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value
     network.network_boot_nodes.push_back(node);
   }
 
+  light_node_history = getConfigDataAsUInt(root, {"light_node_history"}, true, 0);
+
+  // Min history for light node is 110% of kDagExpiryPeriodLimit to avoid any race condition around the
+  // kDagExpiryPeriodLimit
+  if (light_node_history > 0 && light_node_history < kDagExpiryPeriodLimit * 1.1) {
```
JakubFornadel, line 153: We have a void `FullNodeConfig::validate()` function in this file, which checks values limits, so this check should be probably there so we keep it on the same place + it should throw ConfigException rather than overriding the value so the node does not even start up and enforce user to setup this value properly in case he put there invalid number

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -584,25 +596,89 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
...
+  updateFrontier();
 
   LOG(log_nf_) << "Set new period " << period << " with anchor " << new_anchor;
 
   return dag_order_set.size();
 }
 
+void DagManager::handleExpiredDagBlocksTransactions(std::vector<trx_hash_t> &transactions_from_expired_dag_blocks) {
+  std::unordered_set<trx_hash_t> transactions_from_expired_dag_blocks_to_remove;
+  // Only transactions that need to be moved to memory pool are the one that are not yet finalized and are not part of
```
JakubFornadel, line 631: are the ones

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -584,25 +596,89 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
...
   anchor_ = new_anchor;
   period_ = period;
+  updateFrontier();
 
   LOG(log_nf_) << "Set new period " << period << " with anchor " << new_anchor;
 
   return dag_order_set.size();
 }
 
+void DagManager::handleExpiredDagBlocksTransactions(std::vector<trx_hash_t> &transactions_from_expired_dag_blocks) {
```
JakubFornadel, line 629: missing two const ? 
`void DagManager::handleExpiredDagBlocksTransactions(const std::vector<trx_hash_t> &transactions_from_expired_dag_blocks)  const `

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -584,25 +596,89 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
   assert(dag_order_set.count(new_anchor));
   addToDag(new_anchor, blk_hash_t(), vec_blk_t(), 0, true);
 
+  std::pair<std::unordered_map<blk_hash_t, std::shared_ptr<DagBlock>>, std::vector<trx_hash_t>>
+      expired_dag_blocks_to_remove;
```
JakubFornadel, line 600: This whole std::pair seems overcomplicated and could be simplified, which would also improved readability:

Instead of 1 pair, you could declre two separate containers with proper names:
```
std::unordered_map<blk_hash_t, std::shared_ptr<DagBlock>> ct1;
std::vector<trx_hash_t> ct2;
```

Read comments below, which are related

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -584,25 +596,89 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
...
-        auto dag_block = dag_blk_mgr_->getDagBlock(blk);
+    for (auto &blk_hash : v.second) {
+      if (dag_order_set.count(blk_hash) == 0) {
+        auto dag_block = dag_blk_mgr_->getDagBlock(blk_hash);
         auto pivot_hash = dag_block->getPivot();
-        addToDag(blk, pivot_hash, dag_block->getTips(), dag_block->getLevel(), false);
+
+        if (validateBlockNotExpired(dag_block, expired_dag_blocks_to_remove)) {
+          addToDag(blk_hash, pivot_hash, dag_block->getTips(), dag_block->getLevel(), false);
+        }
```
JakubFornadel, line 609: this could be refactored to:

if (validateBlockNotExpired(dag_block, ct1)) {
  addToDag(blk_hash, pivot_hash, dag_block->getTips(), dag_block->getLevel(), false);
} else {
 //  add dag_block->getTrxs() to ct2
}

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -278,6 +278,20 @@ void TransactionManager::updateFinalizedTransactionsStatus(SyncBlock const &sync
   db_->saveStatusField(StatusDbField::TrxCount, trx_count_);
 }
 
+void TransactionManager::moveNonFinalizedTransactionsToTransactionsPool(std::unordered_set<trx_hash_t> &&transactions) {
+  // !!! There is no lock because it is called under std::unique_lock trx_lock(trx_mgr_->getTransactionsMutex());
+  auto write_batch = db_->createWriteBatch();
+  for (auto const &trx_hash : transactions) {
+    db_->removeTransactionToBatch(trx_hash, write_batch);
```
JakubFornadel, line 285: should be this inside `if (trx != nonfinalized_transactions_in_dag_.end()) {` ?

libraries/core_libs/network/src/tarcap/shared_states/peers_state.cpp
JakubFornadel: to go from light node to full node, user have to change config and restart the node, in which case the node will create new connections as full node, so `network_peer_light_node_timeout` seems like we could delete it ?

libraries/cli/include/cli/default_config.hpp
```diff
@@ -16,6 +16,8 @@ const char *default_json = R"foo({
...
   "network_packets_processing_threads": 14,
   "network_peer_blacklist_timeout" : 600,
+  "network_peer_light_node_timeout" : 600,
+  "light_node_history" : 0,
```
JakubFornadel, line 20: This is one of the very few config parameters that users might actually touch. I would suggest to make it more user friendly and have here flag `"is_light_node" : true/false` + `"light_node_history"`, which would be set to some default value and users would need only change is_light_node param

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -51,6 +52,17 @@ void StatusPacketHandler::process(const PacketData& packet_data, const std::shar
     auto const node_major_version = (*it++).toInt<unsigned>();
     auto const node_minor_version = (*it++).toInt<unsigned>();
     auto const node_patch_version = (*it++).toInt<unsigned>();
+    auto const node_history = (*it++).toInt<uint64_t>();
+
+    // If this is a light node and it cannot serve our sync request disconnect from it
+    if (node_history > 0) {
+      if (pbft_synced_period + node_history < peer_pbft_chain_size) {
```
JakubFornadel, line 59: I dont think that this s enough. What if I am light node and when I connected to you this condition did not pass so everything is ok but for some reason, my node get out of sync(for any reason - connection problems, etc...). You would try to get sync data from me but I dont have them. I think we have to handle this + we should probably also save  node_history for each light_node peer so we actually know if and how much data can we ask from him

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -51,6 +52,17 @@ void StatusPacketHandler::process(const PacketData& packet_data, const std::shar
     auto const node_major_version = (*it++).toInt<unsigned>();
     auto const node_minor_version = (*it++).toInt<unsigned>();
     auto const node_patch_version = (*it++).toInt<unsigned>();
+    auto const node_history = (*it++).toInt<uint64_t>();
```
JakubFornadel, line 55: if we are going to use "is_light_node" flag in config, I would use it also here + light_node_history

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: I have put it now in both

libraries/config/include/config/config.hpp
mfrankovi: Implemented

libraries/config/src/config.cpp
mfrankovi: this part of code changed, please review again

libraries/cli/include/cli/default_config.hpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: Refactored

libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
mfrankovi: Yes, good catch. Fixed

libraries/cli/include/cli/default_config.hpp
mfrankovi: Done

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
mfrankovi: I believe that any scenario like this is already handled and we would disconnect from any node that is not able to serve us data we need. Maybe we need additional ways to detect that we are not syncing and try to disconnect from other nodes but I do not think this is related to light node functionality.

libraries/config/src/config.cpp
```diff
@@ -144,6 +146,11 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value
     network.network_boot_nodes.push_back(node);
   }
 
+  is_light_node = getConfigDataAsBoolean(root, {"is_light_node"}, true, false);
+  if (is_light_node) {
+    light_node_history = getConfigDataAsUInt(root, {"light_node_history"}, true, kDefaultLightNodeHistory);
```
gaoqi8649, line 151: You removed light_node_history in config file

libraries/core_libs/network/src/tarcap/shared_states/peers_state.cpp
JakubFornadel: exactly, this `light_node` flag should be inside Peer class as once the new node connects it is either light or full node. It cannot change it's state without restarting the node. So let's put it into the Peer class and then you dont need no timeout either

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
JakubFornadel: hmm I think the solution is pretty easy to implement. Check my other comment first. I would do this:

- add light_node flag to Peer class
- add optional<light_node_history> to Peer class
- any time we ask for sync data, we check if the peer is light node(we will node as it will be in Peer class). If it is light node we will check also num of pbft blocks he keeps and compare it to what we are asking for. If he does not have such long history, we will skip for another peer

This check will stay here + the functionality from above will make it robust so in any scenario (which we might not even know about) we will not ask for sync data from a light node, which he wont ba able to serve

libraries/core_libs/consensus/src/pbft/pbft_chain.cpp
```diff
@@ -107,6 +114,9 @@ void PbftChain::updatePbftChain(blk_hash_t const& pbft_block_hash, bool null_anc
   size_++;
   if (!null_anchor) {
     non_empty_size_++;
+    if ((dag_expiry_period_ == 0 && non_empty_size_ > kDagExpiryPeriodLimit) || dag_expiry_period_ > 0) {
```
gaoqi8649, line 117: This should not right. Like non_empty_size_ is 2000 and faucet down at the moment. There may still some DAG blocks in the last DAG level as sibling for non finalized. If no new DAG block generate, these DAG blocks will be treat as expired.

libraries/config/src/config.cpp
JakubFornadel: yeah I think it was because of my comment with flag. So the idea with flag is that most people, who will run light node wont really configure light_node_history so it is much more user friendly to have the light_node flag, which they can set to true/false and use default value for light_node_history.

But we still can and probably should have also this parameter for setting the light_node_history -> Some advanced users might run some kind of application where they need full data for the past N days for example so they might set this history to some bigger number as the default value we use...

mfrankovi: this is added as an optional parameter which is not present in the default config files but it is supported, I think this kind of more advanced params do not need to be in the config file but they can be added manually if advanced user wants to add them.

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
mfrankovi: I can try to implement this but I still believe that we should disconnect from light nodes if we are out of sync and light nodes cannot serve our sync requests. There is always a chance that most/all of the connected nodes are light nodes and I do not see any point in being connected with light node if we are out of sync. If we do not have some timeout/state outside of the Peer class than we might get into some state where we keep connecting over and over again to the same peer so I do not see how this can work with keeping this only in Peer class

gaoqi8649 (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
JakubFornadel: true, to solve this without any complicated timers: once we get into the `restartPbftSyncing` function we can go through the connected peers and if we find out that none of the connected peers are able to serve us the data(because they are light nodes), we would simply disconnect all of them and we would not be able to create a new  connection with them due the condition in initial status packet

libraries/cli/include/cli/default_config.hpp
```diff
@@ -16,6 +16,8 @@ const char *default_json = R"foo({
   "network_sync_level_size": 25,
   "network_packets_processing_threads": 14,
   "network_peer_blacklist_timeout" : 600,
+  "network_peer_light_node_timeout" : 600,
+  "is_light_node" : false,
```
JakubFornadel, line 20: are we going to use also `light_node_history` config parameter together with `is_light_node` flag ?

- is_light_node is just easy way to say for most of the users if they wanna run light/full node through true/false flag
- light_node_history would be set to some default value that most of the users dont need to touch at all, but if someone builds an app on top of our blockchain and needs to run local node with lets say 1 week of history, he would have a possibility to set custom history for his light node

libraries/core_libs/network/src/tarcap/shared_states/peers_state.cpp
mfrankovi: Implemented

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -39,6 +42,13 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
     throw MaliciousPeerException(err_msg.str());
   }
 
+  if (is_light_node_ && height_to_sync + light_node_history_ <= my_chain_size) {
+    std::ostringstream err_msg;
+    err_msg << "Peer " << packet_data.from_node_id_ << " request syncing period start at " << height_to_sync
+            << ". Light node does not have the data " << my_chain_size;
+    throw MaliciousPeerException(err_msg.str());
```
gaoqi8649, line 49: I think sender should not know the peer is a light node and don't have the block before sending request. Do we need to mark sender as malicous here?

mfrankovi: Sender should know this since it is being sent as part of status message

Issue #1516, opened 2022-02-07
Enable replay protection service


MatusKysel:
## Epic Parent

[Transaction ordering & delegation process](https://github.com/Taraxa-project/taraxa-node/issues/1480).

PR #1553, opened 2022-02-18
chore: enable strict ordering of transactions

MatusKysel:
## Purpose
Fixes: #1516

### DIFF
--- a/tests/util_test/util.hpp
+++ b/tests/util_test/util.hpp
@@ -149,6 +149,9 @@ inline auto make_node_cfgs(uint count) {
       cfg.chain.final_chain.state.genesis_balances[root_node_addr] = 9007199254740991;
       auto& dpos = *cfg.chain.final_chain.state.dpos;
       dpos.genesis_state[root_node_addr][root_node_addr] = dpos.eligibility_balance_threshold;
+      // As test are badly written let's disable it for now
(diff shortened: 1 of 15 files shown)


JakubFornadel (approved):

tests/util_test/util.hpp
```diff
@@ -124,6 +124,9 @@ inline auto make_node_cfgs(uint count) {
       cfg.chain.final_chain.state.genesis_balances[root_node_addr] = 9007199254740991;
       auto& dpos = *cfg.chain.final_chain.state.dpos;
       dpos.genesis_state[root_node_addr][root_node_addr] = dpos.eligibility_balance_threshold;
+      // As test are badly written let's disable it for now
+      cfg.chain.final_chain.state.execution_options.disable_nonce_check = true;
```
gaoqi8649, line 128: This will cause lots of unit tests disable nonce check. Which unit tests will test nonce checking?

MatusKysel: With this change we have 2 levels of nonce checking, first one when we verifying transaction and second one when we execute transaction. This flag is disabling second one, that is part of EVM and not this PR. Anyway I am not disabling anything new, just moving this flag only for tests. If we want to enable it by default, we need to rework a lot of tests and that's for separate task

gaoqi8649: I see. Thanks!

Issue #1554, opened 2022-02-18
CryptoTest.leader_selection failing


kstdl:
CryptoTest.leader_selection sometimes failing failing on develop 
https://app.circleci.com/pipelines/github/Taraxa-project/taraxa-node/5064/workflows/fff0db1a-9393-45ad-b96c-03fdf1b23647/jobs/5491

JakubFornadel:
needs to be revisited and finetuned so the test never fails...

PR #1555, opened 2022-02-18
chore: fix selection unit test in probability

gaoqi8649:
Fix https://github.com/Taraxa-project/taraxa-node/issues/1554 

Fix the PBFT proposed blocks leader selection unit test with specific VRF keys and nodes voting power. Also change testing rounds from 1000 to 100, reduce the unit test running time to 2s.

### DIFF
--- a/tests/crypto_test.cpp
+++ b/tests/crypto_test.cpp
@@ -354,8 +354,8 @@ TEST_F(CryptoTest, leader_selection) {
   const uint64_t committee_size = 20;
   const uint64_t rounds = 1000;
   const uint64_t low_stake_nodes_num = 50;
-  const uint64_t high_stake_nodes_num = 2;
-  const uint64_t high_stake_nodes_power = std::rand() % 2000;
+  const uint64_t high_stake_nodes_num = 6;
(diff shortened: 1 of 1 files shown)


MatusKysel:
did you run it at least 100x ?

gaoqi8649:
> 



> did you run it at least 100x ?

I modify the unit test with specific VRF keys and nodes voting power. So every time running result should be exactly the same.

JakubFornadel (approved):

MatusKysel:
> > 
> 
> > did you run it at least 100x ?
> 
> I modify the unit test with specific VRF keys and nodes voting power. So every time running result should be exactly the same.

What is the meaning of this test then? we everything is fixed what we are testing?

mfrankovi (approved):

MatusKysel (changes requested):
What is the meaning of this test then? we everything is fixed what we are testing?

gaoqi8649:
> What is the meaning of this test then? we everything is fixed what we are testing?

I think the purpose of unit tests are fixed inputs and test exactly output. 
If PBFT selection changed(include binomial distribution, hashing, leader selection), that will effect the output. Then we will known if the change is correct or not

Issue #1562, opened 2022-02-21
Remove TestPacket


JakubFornadel:
TestPacket is used 1 single test just for sending some dummy data to see if it was received by the other node. To test this send/receive functionality we could use one of the other packet types that we support and actually use. 

- Just because of this 1 test we have TestPacket handler and a lot of useless code
- It is also part of node, which means malicious node could just send us TestPacket and try to do some sort of an attack

PR #1567, opened 2022-02-22
chore: remove testpacket that is unnecessary

MatusKysel:
## Purpose
Fixed: #1562


kstdl (approved):

mfrankovi (commented):
I think this is a breaking change since by removing the packet you are changing internal enum values of other packets

JakubFornadel (approved):

Issue #1563, opened 2022-02-21
Broken local-net


JakubFornadel:
Local net does not work anymore. I wont start:
```

state_api.cpp(63): Throw in function auto taraxa::state_api::ErrorHandler::(anonymous class)::operator()(void *, taraxa_evm_Bytes)::(anonymous class)::operator()() const
[rpc-node-0]     Dynamic exception type: boost::wrapexcept<taraxa::state_api::TaraxaEVMError>
[rpc-node-0]     std::exception::what: insufficient balance for the deposits
```

Issue #1471, opened 2022-01-21
Default config should NOT be mainnet


MatusKysel:
## Task Description
I think it is wrong to have default config as mainnet, as it is used for local/pr-nets ..., we should keep it separated

PR #1570, opened 2022-02-22
chore: re-introduce default config

MatusKysel:
## Purpose
Fixes: #1563 & #1471


MatusKysel:
@rattrap please check the new default config, if we have all needed accounts there. Now it should match old values

JakubFornadel (approved):

mfrankovi (approved):

PR #1589, opened 2022-03-01
fix: reading packet size by checking size of the actual vector of bytes

JakubFornadel:


mfrankovi (approved):

MatusKysel (approved):

gaoqi8649 (commented):
rlp_.size() should be same with rlp_.data().size()

  /// @returns the number of bytes in the data, or zero if it isn't data.
  size_t size() const { return isData() ? length() : 0; }

JakubFornadel:
> rlp_.size() should be same with rlp_.data().size()
> 
> /// @returns the number of bytes in the data, or zero if it isn't data. size_t size() const { return isData() ? length() : 0; }

I thought so too but no it is not the same, `rlp_.size()` actually checks if the rlp is data object, if not it returns 0. All of our packets are lists and not data objects so it returns 0 for every packet...

PR #1590, opened 2022-03-01
test: added script for running tests in infinite loop

JakubFornadel:

### DIFF
--- a/for_devs/run_test_in_loop.sh
+++ b/for_devs/run_test_in_loop.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+# This simple script runs a specified test in an infinite loop until it fails. It prints run number before each new run
+# + when it fails, script ends and it prints the error + number of runs until it failed.
+# It is used to find bugs like rare race conditions that happen only occasionally, etc...
+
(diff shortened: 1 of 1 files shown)


for_devs/run_test_in_loop.sh
```diff
@@ -0,0 +1,26 @@
...
+# It is used to find bugs like rare race conditions that happen only occasionally, etc...
+
+# Usage: ./for_devs/run_test_in_loop.sh
+
+COUNTER=1
+
+while true; do 
+    printf "\n\n******************************************** TEST RUN %d  ************************************\n\n" $COUNTER
+
+    ./cmake-build-debug/bin/full_node_test --gtest_filter=FullNodeTest.sync_five_nodes:FullNodeTest/*.sync_five_nodes:FullNodeTest.sync_five_nodes/*:*/FullNodeTest.sync_five_nodes/*:*/FullNodeTest/*.sync_five_nodes --gtest_color=no
```
MatusKysel, line 14: I think this script should take this line as argument

gaoqi8649 (approved):

for_devs/run_test_in_loop.sh
JakubFornadel: true, I fixed that

Issue #1588, opened 2022-03-01
sync_five_nodes test failure


JakubFornadel:
While running tests on [this PR](https://github.com/Taraxa-project/taraxa-node/pull/1564), I randomly run into the test failure:

https://app.circleci.com/pipelines/github/Taraxa-project/taraxa-node/5188/workflows/2f81165e-5787-4ea8-9092-00653863a979/jobs/5625

```
0x00007f39cfff7700 4fae949a… TRXMGR [2022-03-01 02:51:04.504088] ERROR: Block #94b3bccc… has missing transaction #72f34e0d…
0x00007f39cfff7700 4fae949a… TRXMGR [2022-03-01 02:51:04.504088] ERROR: Block #94b3bccc… has missing transaction #72f34e0d…
0x00007f39cfff7700 4fae949a… DAGBLKMGR [2022-03-01 02:51:04.504269] ERROR: Ignore block #94b3bccc… since it has missing transactions
0x00007f39cfff7700 4fae949a… DAGBLKMGR [2022-03-01 02:51:04.504269] ERROR: Ignore block #94b3bccc… since it has missing transactions
0x00007f39cfff7700 4fae949a… DAGBLKMGR [2022-03-01 02:51:04.504269] ERROR: Ignore block #94b3bccc… since it has missing transactions
0x00007f39cfff7700 4fae949a… DAGBLKMGR [2022-03-01 02:51:04.504269] ERROR: Ignore block #94b3bccc… since it has missing transactions
0x00007f39cfff7700 4fae949a… DAGBLKMGR [2022-03-01 02:51:04.504269] ERROR: Ignore block #94b3bccc… since it has missing transactions
Initial coin transfers from node 0 issued ... 
/opt/taraxa/tests/full_node_test.cpp:501: Failure
Expected equality of these values:
  result.stage
    Which is: 4-byte object <00-00 00-00>
  TransactionClient::TransactionStage::inserted
    Which is: 4-byte object <01-00 00-00>
/opt/taraxa/tests/full_node_test.cpp:501: Failure
Expected equality of these values:
  result.stage
    Which is: 4-byte object <00-00 00-00>
  TransactionClient::TransactionStage::inserted
    Which is: 4-byte object <01-00 00-00>

....

[...]

Issue #1568, opened 2022-02-22
Instability of NetworkTest.node_sync_with_transactions


MatusKysel:
Recent changes on develop made this test unstable so it is crashing from time to time

PR #1597, opened 2022-03-02
 stabilize  tests

MatusKysel:
## Purpose
Fixes: #1588
Fixes: #1568

- sync_five_nodes -> now we are waiting until initial transaction are executed before we continue with the test
- DISABLED_transfer_lot_of_blocks -> @mfrankovi PR should fix that
- node_full_sync -> now not checking for syncing due to empty PBFT blocks but checking non empty blocks
- pbft_manager_run_multi_nodes -> checking for non-empty blocks and increase timeouts because of execution can takes more time

### DIFF
--- a/tests/network_test.cpp
+++ b/tests/network_test.cpp
@@ -25,8 +25,6 @@ auto g_secret = Lazy([] {
                      dev::Secret::ConstructFromStringType::FromHex);
 });
 auto node_key = dev::KeyPair(g_secret);
-
-auto g_trx_samples = Lazy([] { return samples::createMockTrxSamples(0, NUM_TRX); });
 auto g_signed_trx_samples = Lazy([] { return samples::createSignedTrxSamples(0, NUM_TRX, g_secret); });
 
 const unsigned NUM_TRX2 = 200;
@@ -34,7 +32,6 @@ auto g_secret2 = Lazy([] {
   return dev::Secret("3800b2875669d9b2053c1aff9224ecfdc411423aac5b5a73d7a45ced1c3b9dcd",
                      dev::Secret::ConstructFromStringType::FromHex);
 });
-auto g_trx_samples2 = Lazy([] { return samples::createMockTrxSamples(0, NUM_TRX2); });
 auto g_signed_trx_samples2 = Lazy([] { return samples::createSignedTrxSamples(0, NUM_TRX2, g_secret2); });
 auto three_default_configs = Lazy([] { return make_node_cfgs(3); });
 auto g_conf1 = Lazy([] { return three_default_configs[0]; });
@@ -1233,7 +1230,7 @@ TEST_F(NetworkTest, node_transaction_sync) {
 // intervals on randomly selected nodes It verifies that the blocks created from
 // these transactions which get created on random nodes are synced and the
 // resulting DAG is the same on all nodes
-TEST_F(NetworkTest, DISABLED_node_full_sync) {
+TEST_F(NetworkTest, node_full_sync) {
   constexpr auto numberOfNodes = 5;
   auto node_cfgs = make_node_cfgs<20>(numberOfNodes);
   auto nodes = launch_nodes(slice(node_cfgs, 0, numberOfNodes - 1));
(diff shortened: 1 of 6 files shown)


tests/network_test.cpp
```diff
@@ -80,7 +77,8 @@ TEST_F(NetworkTest, transfer_block) {
 }
 
 // Test creates two Network setup and verifies sending blocks between is successfull
-TEST_F(NetworkTest, transfer_lot_of_blocks) {
+// This test can not work anymore as we are marking other nodes as malicous becasue of invalid dag blocks
```
JakubFornadel, line 80: solution is to send valid blocks and not disabling the test

JakubFornadel (changes requested):
I did not check the other changes, but the fix for sync_five_nodes from this PR probably did not do anything -  at least on my PC under Debug it did not help at all. dummy_transaction() that you deleted was not causing any issues and the errors we have for failing txs were not caused by dummy txs. 

At the beginning we sent 5 txs that would transfer some tokens from node0 to all other nodes so they can start sending txs too but for some reason these were not always finalized on all nodes and then when they tried to send tokens, those txs would start failing because of insufficient balance. 

I fixed it here https://github.com/Taraxa-project/taraxa-node/pull/1600

Let's keep the other changes you made and remove the ones from sync_five_nodes test

tests/network_test.cpp
MatusKysel: @mfrankovi  has a fix in his PR, so this will work for now

MatusKysel:
> I did not check the other changes, but the fix for sync_five_nodes from this PR probably did not help - at least on my PC under Debug it did not help and I experienced the same errors as before. dummy_transaction() that you deleted was not causing any issues and the errors we have for failing txs were not caused by dummy txs.
> 
> At the beginning we sent 5 txs that would transfer some tokens from node0 to all other nodes so they can start sending txs too but for some reason these were not always finalized on all nodes and then when they tried to send tokens, those txs would start failing because of insufficient balance.
> 
> I fixed it here #1600
> 
> Let's keep the other changes you made and remove the ones from sync_five_nodes test

- you did not fixed the test, just change it's behavior 
- please stop editing my PRs and task assigned to me !!!

tests/network_test.cpp
```diff
@@ -80,7 +77,8 @@ TEST_F(NetworkTest, transfer_block) {
...
+TEST_F(NetworkTest, DISABLED_transfer_lot_of_blocks) {
```
gaoqi8649, line 81: Let's enable the transfer_lot_of_blocks unit test, Miho's PR has merged in develop branch

gaoqi8649 (approved):

MatusKysel:
> 

let's talk about that over the meeting, it will be easier.

MatusKysel:
> > > I did not check the other changes, but the fix for sync_five_nodes from this PR probably did not help - at least on my PC under Debug it did not help and I experienced the same errors as before. dummy_transaction() that you deleted was not causing any issues and the errors we have for failing txs were not caused by dummy txs.
> > > At the beginning we sent 5 txs that would transfer some tokens from node0 to all other nodes so they can start sending txs too but for some reason these were not always finalized on all nodes and then when they tried to send tokens, those txs would start failing because of insufficient balance.
> > > I fixed it here #1600
> > > Let's keep the other changes you made and remove the ones from sync_five_nodes test
> > 
> > 
> > 
> > * you did not fixed the test, just change it's behavior
> > * please stop editing my PRs and task assigned to me !!!
> 
> This is such waste of time, we could have had it merged yesterday as both of our solutions fix it, but ok I am going to follow your `"please stop editing my PRs and task assigned to me !!!"` and provide you here a few facts why I edited your PR and unassigned it from you:
> 
>     * when you proposed your first solution (removal of dummy_transaction) I knew for 100% it is not enough so while you were sleeping I debugged it and found out what exactly is the problem - first 5 txs would not be applied on all nodes. [The fix for this is here - this single line](https://github.com/Taraxa-project/taraxa-node/pull/1600/files#diff-cb712a3395b4b23506c8277c7056adf0f3cf679e3e7f04bb3de00f3fb1a68dd4R496). Just this line was enough for the test stop failing (you fixed it differently and achieved the same thing [here](https://github.com/Taraxa-project/taraxa-node/pull/1597/files#diff-cb712a3395b4b23506c8277c7056adf0f3cf679e3e7f04bb3de00f3fb1a68dd4R551))
> 
>     * So the 1 line change was basically enough, but then I realized that the test is not about those first 5 txs anyway(in the next for loop we send approx 50 txs so if there is 50 or 55 it does not really matter) so I simply removed it and made the test little bit faster as we can skip this initial tokens distribution and just put it into the config
> 
> 
> So thats it, I basically did try to help. I fixed the test, edited your PR and and assigned it to me as at that time your solution was not working and I did expect that we simply merge this and move on.

let's talk about that over the meeting, it will be easier.

PR #1600, opened 2022-03-02
chore: fixed sync five nodes test

JakubFornadel:
## Purpose

At the beginning we used to send 5 txs that would transfer some tokens from node0 to all other nodes so they can start sending txs too but for some reason these were not always finalized on all nodes and then when they tried to send tokens, those txs would start failing because of insufficient balance. 

1. I am setting some initial balance through config for all five nodes instead of sending tokens from node 0 to all other nodes. It is faster and we dont need to wait until those txs are finalized as this test is not about that...
2. I am also checking if `wait_all_transactions_known` was actually successful as we did not check it before

### DIFF
--- a/tests/full_node_test.cpp
+++ b/tests/full_node_test.cpp
@@ -450,10 +450,20 @@ TEST_F(FullNodeTest, db_test) {
   EXPECT_EQ(period_1_levels_from_db.levels_interval.second, 110);
 }
 
-TEST_F(FullNodeTest, DISABLED_sync_five_nodes) {
+TEST_F(FullNodeTest, sync_five_nodes) {
   using namespace std;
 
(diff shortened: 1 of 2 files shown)


gaoqi8649 (approved):

MatusKysel (changes requested):
I do not think this is solution, you are just hiding the real problem. We should figure out why one node does not receive transaction

JakubFornadel:
> I do not think this is solution, you are just hiding the real problem. We should figure out why one node does not receive transaction

this fix is similar to yours, I am not hiding the problem. Explanation here: https://github.com/Taraxa-project/taraxa-node/pull/1597#issuecomment-1058630236

Let's merge either this one or your it does not really matter and move on...

Issue #1561, opened 2022-02-21
Rework PbftSyncPacket


JakubFornadel:
The way we handle PbftSyncPacket must be changed as we are using empty rlp as valid data -> it means pbft syncing is complete. 

This is bad - someone could simply send empty PbftSyncPacket over and over again

We could use some bool flag to indicate that syncing is complete

PR #1607, opened 2022-03-04
fix: add pbft chain synced flag

gaoqi8649:
Fix https://github.com/Taraxa-project/taraxa-node/issues/1561

### DIFF
--- a/libraries/config/src/config.cpp
+++ b/libraries/config/src/config.cpp
@@ -130,7 +130,6 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value 
   network.network_ideal_peer_count = getConfigDataAsUInt(root, {"network_ideal_peer_count"});
   network.network_max_peer_count = getConfigDataAsUInt(root, {"network_max_peer_count"});
   network.network_sync_level_size = getConfigDataAsUInt(root, {"network_sync_level_size"});
-  assert(network.network_sync_level_size);  // network_sync_level_size cannot be 0
   network.network_packets_processing_threads = getConfigDataAsUInt(root, {"network_packets_processing_threads"});
   network.network_peer_blacklist_timeout = getConfigDataAsUInt(root, {"network_peer_blacklist_timeout"}, true,
                                                                NetworkConfig::kBlacklistTimeoutDefaultInSeconds);
@@ -249,6 +248,10 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value 
 }
 
 void FullNodeConfig::validate() {
+  if (network.network_sync_level_size == 0) {
+    throw ConfigException(std::string("network_sync_level_size cannot be 0"));
+  }
+
   // Max enabled number of threads for processing rpc requests
   constexpr uint16_t MAX_PACKETS_PROCESSING_THREADS_NUM = 30;
   if (network.network_packets_processing_threads < 3 ||
--- a/libraries/config/src/config.cpp
+++ b/libraries/config/src/config.cpp
@@ -130,6 +130,7 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value 
   network.network_ideal_peer_count = getConfigDataAsUInt(root, {"network_ideal_peer_count"});
   network.network_max_peer_count = getConfigDataAsUInt(root, {"network_max_peer_count"});
   network.network_sync_level_size = getConfigDataAsUInt(root, {"network_sync_level_size"});
+  assert(network.network_sync_level_size);  // network_sync_level_size cannot be 0
   network.network_packets_processing_threads = getConfigDataAsUInt(root, {"network_packets_processing_threads"});
   network.network_peer_blacklist_timeout = getConfigDataAsUInt(root, {"network_peer_blacklist_timeout"}, true,
                                                                NetworkConfig::kBlacklistTimeoutDefaultInSeconds);
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
@@ -40,43 +40,41 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
   }
 
   size_t blocks_to_transfer = 0;
-  auto will_be_synced = false;
+  auto pbft_chain_synced = false;
   const auto total_sync_blocks_size = my_chain_size - height_to_sync + 1;
   if (total_sync_blocks_size <= network_sync_level_size_) {
     blocks_to_transfer = total_sync_blocks_size;
-    will_be_synced = true;
+    pbft_chain_synced = true;
   } else {
     blocks_to_transfer = network_sync_level_size_;
   }
   LOG(log_tr_) << "Will send " << blocks_to_transfer << " PBFT blocks to " << packet_data.from_node_id_;
 
-  sendPbftBlocks(packet_data.from_node_id_, height_to_sync, blocks_to_transfer, will_be_synced);
+  sendPbftBlocks(packet_data.from_node_id_, height_to_sync, blocks_to_transfer, pbft_chain_synced);
 }
 
 // api for pbft syncing
 void GetPbftSyncPacketHandler::sendPbftBlocks(dev::p2p::NodeID const &peer_id, size_t height_to_sync,
-                                              size_t blocks_to_transfer, bool will_be_synced) {
+                                              size_t blocks_to_transfer, bool pbft_chain_synced) {
   LOG(log_tr_) << "sendPbftBlocks: peer want to sync from pbft chain height " << height_to_sync
                << ", will send at most " << blocks_to_transfer << " pbft blocks to " << peer_id;
 
-  uint64_t current_period = height_to_sync;
-  while (current_period < height_to_sync + blocks_to_transfer) {
-    bool last_block = (current_period == height_to_sync + blocks_to_transfer - 1);
-    auto data = db_->getPeriodDataRaw(current_period);
+  for (auto block_period = height_to_sync; block_period < height_to_sync + blocks_to_transfer; block_period++) {
+    bool last_block = (block_period == height_to_sync + blocks_to_transfer - 1);
+    auto data = db_->getPeriodDataRaw(block_period);
 
     if (data.size() == 0) {
-      LOG(log_er_) << "DB corrupted. Cannot find period " << current_period << " PBFT block in db";
+      LOG(log_er_) << "DB corrupted. Cannot find period " << block_period << " PBFT block in db";
       assert(false);
     }
 
     dev::RLPStream s;
     s.appendList(3);
-    s << (will_be_synced && last_block);
+    s << (pbft_chain_synced && last_block);
     s << last_block;
     s.appendRaw(data);
-    LOG(log_dg_) << "Sending PbftSyncPacket period " << current_period << " to " << peer_id;
+    LOG(log_dg_) << "Sending PbftSyncPacket period " << block_period << " to " << peer_id;
     sealAndSend(peer_id, SubprotocolPacketType::PbftSyncPacket, std::move(s));
-    current_period++;
   }
 }
 
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
@@ -32,38 +32,58 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
   // Here need PBFT chain size, not synced period since synced blocks has not verified yet.
   const size_t my_chain_size = pbft_chain_->getPbftChainSize();
   size_t blocks_to_transfer = 0;
+  auto will_be_synced = false;
   if (my_chain_size >= height_to_sync) {
-    blocks_to_transfer = std::min(network_sync_level_size_, (my_chain_size - (height_to_sync - 1)));
+    auto total_sync_blocks_size = my_chain_size - height_to_sync + 1;
+    if (total_sync_blocks_size <= network_sync_level_size_) {
+      blocks_to_transfer = total_sync_blocks_size;
+      will_be_synced = true;
+    } else {
+      blocks_to_transfer = network_sync_level_size_;
+    }
+    LOG(log_tr_) << "Will send " << blocks_to_transfer << " PBFT blocks to " << packet_data.from_node_id_;
+  } else {
+    LOG(log_wr_) << "Request syncing period is ahead of own PBFT chain. Peer " << packet_data.from_node_id_
+                 << " want to sync PBFT blocks period from " << height_to_sync << ". But own PBFT chain size is "
+                 << my_chain_size;
+    will_be_synced = true;
   }
-  LOG(log_tr_) << "Will send " << blocks_to_transfer << " PBFT blocks to " << packet_data.from_node_id_;
 
-  sendPbftBlocks(packet_data.from_node_id_, height_to_sync, blocks_to_transfer);
+  sendPbftBlocks(packet_data.from_node_id_, height_to_sync, blocks_to_transfer, will_be_synced);
 }
 
 // api for pbft syncing
 void GetPbftSyncPacketHandler::sendPbftBlocks(dev::p2p::NodeID const &peer_id, size_t height_to_sync,
-                                              size_t blocks_to_transfer) {
+                                              size_t blocks_to_transfer, bool will_be_synced) {
   LOG(log_tr_) << "sendPbftBlocks: peer want to sync from pbft chain height " << height_to_sync
                << ", will send at most " << blocks_to_transfer << " pbft blocks to " << peer_id;
-  uint64_t current_period = height_to_sync;
+  dev::RLPStream s;
+  auto pbft_chain_synced = false;
   if (blocks_to_transfer == 0) {
-    LOG(log_dg_) << "Sending empty PbftSyncPacket to " << peer_id;
-    sealAndSend(peer_id, SubprotocolPacketType::PbftSyncPacket, dev::RLPStream(0));
+    pbft_chain_synced = true;
+    s << pbft_chain_synced;
+    sealAndSend(peer_id, SubprotocolPacketType::PbftSyncPacket, std::move(s));
     return;
   }
 
+  uint64_t current_period = height_to_sync;
   while (current_period < height_to_sync + blocks_to_transfer) {
     bool last_block = (current_period == height_to_sync + blocks_to_transfer - 1);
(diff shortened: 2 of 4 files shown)


libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -32,38 +32,58 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
   // Here need PBFT chain size, not synced period since synced blocks has not verified yet.
   const size_t my_chain_size = pbft_chain_->getPbftChainSize();
   size_t blocks_to_transfer = 0;
+  auto will_be_synced = false;
   if (my_chain_size >= height_to_sync) {
-    blocks_to_transfer = std::min(network_sync_level_size_, (my_chain_size - (height_to_sync - 1)));
+    auto total_sync_blocks_size = my_chain_size - height_to_sync + 1;
```
MatusKysel, line 37: const

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -55,19 +52,25 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
   }
 
   const size_t item_count = packet_data.rlp_.itemCount();
+  bool pbft_chain_synced = packet_data.rlp_[0].toInt<bool>();
```
MatusKysel, line 55: const

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -32,38 +32,58 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
...
   LOG(log_tr_) << "sendPbftBlocks: peer want to sync from pbft chain height " << height_to_sync
                << ", will send at most " << blocks_to_transfer << " pbft blocks to " << peer_id;
-  uint64_t current_period = height_to_sync;
+  dev::RLPStream s;
+  auto pbft_chain_synced = false;
   if (blocks_to_transfer == 0) {
-    LOG(log_dg_) << "Sending empty PbftSyncPacket to " << peer_id;
-    sealAndSend(peer_id, SubprotocolPacketType::PbftSyncPacket, dev::RLPStream(0));
+    pbft_chain_synced = true;
+    s << pbft_chain_synced;
```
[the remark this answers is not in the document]
MatusKysel: it could be even something like this `sealAndSend(peer_id, SubprotocolPacketType::PbftSyncPacket, std::move(dev::RLPStream(1) << true))`

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -32,38 +32,58 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
...
-    dev::RLPStream s;
-    s.appendList(2);
+    if (will_be_synced && last_block) {
+      pbft_chain_synced = true;
+    }
+    s.appendList(3);
+    s << pbft_chain_synced;
     s << last_block;
     s.appendRaw(data);
     LOG(log_dg_) << "Sending PbftSyncPacket period " << current_period << " to " << peer_id;
```
MatusKysel, line 89: line under this is huge BUG, you can not use `sealAndSend(peer_id, SubprotocolPacketType::PbftSyncPacket, std::move(s));` because `s` is defined outside of this loop and you can not move it several times

MatusKysel (changes requested):
Overall I do not see fix to this part `This is bad - someone could simply send empty PbftSyncPacket over and over again`. This line is copied from the task and I do not know how this is fixing it. Now I could send several packets with `true` for syncing flag and nothing changed. Maybe ask @JakubFornadel what was the idea behind the task

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -25,20 +25,17 @@ PbftSyncPacketHandler::PbftSyncPacketHandler(std::shared_ptr<PeersState> peers_s
       delayed_sync_events_tp_(1, true) {}
 
 void PbftSyncPacketHandler::validatePacketRlpFormat(const PacketData &packet_data) const {
-  // TODO[1561]: the way we handle PbftSyncPacket must be changed as we are using empty rlp as valid data -> it means
-  // pbft
-  //             syncing is complete. This is bad - someone could simply send empty PbftSyncPacket over and over again
-  if (packet_data.rlp_.itemCount() == 0) {
+  if (packet_data.rlp_.itemCount() == 1) {
```
JakubFornadel, line 28: This requires comment. After few weeks/months no one will understand why we return here in case itemCount == 1

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -32,38 +32,58 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
   // Here need PBFT chain size, not synced period since synced blocks has not verified yet.
   const size_t my_chain_size = pbft_chain_->getPbftChainSize();
   size_t blocks_to_transfer = 0;
+  auto will_be_synced = false;
   if (my_chain_size >= height_to_sync) {
```
JakubFornadel, line 36: I believe that any node, which is requesting period higher than our current period should be considered malicious because we provide info about our actual  chain size so other node should never request period higher than what we provided to it(please double check this), so here could be something like this:

```
if (my_chain_size < height_to_sync) {
    throw MaliciousPeerException("err message");
}
```

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -32,38 +32,58 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
...
 // api for pbft syncing
 void GetPbftSyncPacketHandler::sendPbftBlocks(dev::p2p::NodeID const &peer_id, size_t height_to_sync,
-                                              size_t blocks_to_transfer) {
+                                              size_t blocks_to_transfer, bool will_be_synced) {
   LOG(log_tr_) << "sendPbftBlocks: peer want to sync from pbft chain height " << height_to_sync
                << ", will send at most " << blocks_to_transfer << " pbft blocks to " << peer_id;
-  uint64_t current_period = height_to_sync;
+  dev::RLPStream s;
+  auto pbft_chain_synced = false;
   if (blocks_to_transfer == 0) {
```
JakubFornadel, line 62: In case there are no blocks to be sent, we should probably never even get into this function... See my other comments for better understanding

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -32,38 +32,58 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
...
+    sealAndSend(peer_id, SubprotocolPacketType::PbftSyncPacket, std::move(s));
     return;
   }
 
+  uint64_t current_period = height_to_sync;
   while (current_period < height_to_sync + blocks_to_transfer) {
     bool last_block = (current_period == height_to_sync + blocks_to_transfer - 1);
     auto data = db_->getPeriodDataRaw(current_period);
 
     if (data.size() == 0) {
```
JakubFornadel, line 74: I think this should not be possible at all so it could be assert ? Each node node broadcasts its chain size through StatusPacket -> based on that no node should ask for a period that is higher of what the other node has... We should validate this in `GetPbftSyncPacketHandler::process`. If the packet gets through that condition, the only situation when this(we dont have the requested block in db) could happen if there is some serious bug in our code -> we can use assert

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -32,38 +32,58 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
   // Here need PBFT chain size, not synced period since synced blocks has not verified yet.
   const size_t my_chain_size = pbft_chain_->getPbftChainSize();
   size_t blocks_to_transfer = 0;
+  auto will_be_synced = false;
```
JakubFornadel, line 35: this flag seems like a redundant info to what we are already sending -> last_block fklag

JakubFornadel (changes requested):
I might be missing something, but it seems to me that we could do these changes here:

- if a node requests sync period higher that we is our actual chain size, we can disconnect him a mark him as malicious (this should never happen during normal operation as each node provides its chain_size through StatusPacket) 
- once we send PbftSyncPacket as response to GetPbftSyncPacket, it should always contain at least 1 block as the requesting node should never ask for a syncing data if it is not behind. PbftSyncPacket could then consists of `SyncingBlock` + `pbft_synced` flag and that is all we need.

@mfrankovi could you please check this idea if you see any problems in it ?

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
gaoqi8649: fixed, thanks

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
gaoqi8649: fixed, thanks

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -32,38 +32,58 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
...
+  uint64_t current_period = height_to_sync;
   while (current_period < height_to_sync + blocks_to_transfer) {
     bool last_block = (current_period == height_to_sync + blocks_to_transfer - 1);
     auto data = db_->getPeriodDataRaw(current_period);
 
     if (data.size() == 0) {
-      sealAndSend(peer_id, SubprotocolPacketType::PbftSyncPacket, dev::RLPStream(0));
+      pbft_chain_synced = true;
+      s << pbft_chain_synced;
+      sealAndSend(peer_id, SubprotocolPacketType::PbftSyncPacket, std::move(s));
```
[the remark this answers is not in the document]
gaoqi8649: This is DB corrupted, add assertion here

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
gaoqi8649: fixed, thanks

gaoqi8649:
> Overal I do not see fix to this part `This is bad - someone could simply send empty PbftSyncPacket over and over again`. This line is copied from the task and I do not know how this is fixing it. Now I could send several packets with `true` for syncing flag and nothing changed. Maybe ask @JakubFornadel what was the idea behind the task

Fixed with Jakub's review comments

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
gaoqi8649: No itemCount == 1 anymore, removed

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
gaoqi8649: I agree. Fixed, thanks

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
gaoqi8649: I agree. FIxed, thanks

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
gaoqi8649: I agree. Fixed, thanks

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
gaoqi8649: They are different.
last_block flag is the last block in each syncing round
will_be_synced is the flag to tell peer its PBFT chain has synced with mine
For example:
Node1 has PBFT chain size 7, node2 has pbft chain size 0, and network_sync_level_size_ is 5
Round 1:
node1 send 5 PBFT blocks [1, 2, 3, 4, 5] to node2
when send block5, the last_block is true, but will_be_synced is false
Round 2:
node1 send 2 PBFT blocks [6, 7] to node2
when send block7, the last_block is true and will_be_synced is also true

gaoqi8649:
> I might be missing something, but it seems to me that we could do these changes here:
> 
> * if a node requests sync period higher that we is our actual chain size, we can disconnect him a mark him as malicious (this should never happen during normal operation as each node provides its chain_size through StatusPacket)
> * once we send PbftSyncPacket as response to GetPbftSyncPacket, it should always contain at least 1 block as the requesting node should never ask for a syncing data if it is not behind. PbftSyncPacket could then consists of `SyncingBlock` + `pbft_synced` flag(saying this is the last block) and that is all we need.
> 
> @mfrankovi could you please check this idea if you see any problems in it ?

I agree with that. Fixed code, please review

MatusKysel (approved):

libraries/config/src/config.cpp
```diff
@@ -130,6 +130,7 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value
   network.network_ideal_peer_count = getConfigDataAsUInt(root, {"network_ideal_peer_count"});
   network.network_max_peer_count = getConfigDataAsUInt(root, {"network_max_peer_count"});
   network.network_sync_level_size = getConfigDataAsUInt(root, {"network_sync_level_size"});
+  assert(network.network_sync_level_size);  // network_sync_level_size cannot be 0
```
JakubFornadel, line 133: This should be exception in FullNodeConfig::validate() rather than assert here

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -31,39 +31,47 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
...
                << ", will send at most " << blocks_to_transfer << " pbft blocks to " << peer_id;
-  uint64_t current_period = height_to_sync;
-  if (blocks_to_transfer == 0) {
-    LOG(log_dg_) << "Sending empty PbftSyncPacket to " << peer_id;
-    sealAndSend(peer_id, SubprotocolPacketType::PbftSyncPacket, dev::RLPStream(0));
-    return;
-  }
 
+  uint64_t current_period = height_to_sync;
   while (current_period < height_to_sync + blocks_to_transfer) {
```
JakubFornadel, line 63: this would be easier to read:
for (auto block_period = height_to_sync; block_period < height_to_sync + blocks_to_transfer; block_period++)

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
JakubFornadel: hmm ok I understand now. Maybe better name instead of `will_be_synced` is `pbft_synced`. Once you receive packet with this flag set to true, sender tells you that with this packet you are pbft_synced with him...

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -54,20 +47,12 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
...
-  if (item_count == 0) {
-    pbftSyncComplete();
-    return;
-  }
-
   // Process received pbft blocks
-  bool last_block = packet_data.rlp_[0].toInt<bool>();
-
+  const bool pbft_chain_synced = packet_data.rlp_[0].toInt<bool>();
+  const bool last_block = packet_data.rlp_[1].toInt<bool>();
```
JakubFornadel, line 52: pls add here comments with short explanation of what these 2 flags (pbft_chain_synced / last_block) mean and how they differ so other devs dont really need to go through code to understand it if they need to work with this part of code

JakubFornadel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
gaoqi8649: Rename will_be_synced to pbft_chain_synced

libraries/config/src/config.cpp
gaoqi8649: Fixed, thanks

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
gaoqi8649: Fixed, thanks

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -54,20 +47,12 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
...
-  // No blocks received - syncing is complete
-  if (item_count == 0) {
-    pbftSyncComplete();
-    return;
-  }
-
   // Process received pbft blocks
-  bool last_block = packet_data.rlp_[0].toInt<bool>();
-
+  const bool pbft_chain_synced = packet_data.rlp_[0].toInt<bool>();
```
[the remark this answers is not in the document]
gaoqi8649: Unite name in pbft_chain_synced

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
gaoqi8649: Fix adding comments

Issue #1464, opened 2022-01-20
Votes inefficient calculations


JakubFornadel:
## Task Description

During each call to:
PbftManager::stateOperations_() -> resetRound_() -> vote_mgr_->roundDeterminedFromVotes(TWO_T_PLUS_ONE) -> getVotesBundleByRoundAndStep() we iterate thtrough all next votes(in all steps >= 3) over and over again, which is inefficient and not needed. 

We know total number of processed votes in each step so with new vote we just have to add this value to the existing sum... and not iterate through all votes...

PR #1609, opened 2022-03-07
chore: improve round calculation from votes

MatusKysel:
## Purpose
Fixes: #1464

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -692,12 +692,12 @@ bool PbftManager::stateOperations_() {
   // ONLY CHECK IF HAVE *NOT* YET EXECUTED THIS ROUND...
   if (state_ == certify_state && !have_executed_this_round_) {
     auto voted_block_hash_with_cert_votes = vote_mgr_->getVotesBundleByRoundAndStep(round, 3, TWO_T_PLUS_ONE);
-    if (voted_block_hash_with_cert_votes.enough) {
-      LOG(log_dg_) << "PBFT block " << voted_block_hash_with_cert_votes.voted_block_hash << " has enough certed votes";
+    if (voted_block_hash_with_cert_votes) {
+      LOG(log_dg_) << "PBFT block " << voted_block_hash_with_cert_votes->voted_block_hash << " has enough certed votes";
       // put pbft block into chain
-      const auto vote_size = voted_block_hash_with_cert_votes.votes.size();
-      if (pushCertVotedPbftBlockIntoChain_(voted_block_hash_with_cert_votes.voted_block_hash,
-                                           std::move(voted_block_hash_with_cert_votes.votes))) {
+      const auto vote_size = voted_block_hash_with_cert_votes->votes.size();
+      if (pushCertVotedPbftBlockIntoChain_(voted_block_hash_with_cert_votes->voted_block_hash,
+                                           std::move(voted_block_hash_with_cert_votes->votes))) {
         db_->savePbftMgrStatus(PbftMgrStatus::ExecutedInRound, true);
         have_executed_this_round_ = true;
         LOG(log_nf_) << "Write " << vote_size << " cert votes ... in round " << round;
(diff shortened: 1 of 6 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -718,20 +718,22 @@ bool PbftManager::updateSoftVotedBlockForThisRound_() {
...
+                                       voted_block_hash_with_soft_votes->voted_block_hash, batch);
+      db_->addPbftMgrStatusToBatch(PbftMgrStatus::SoftVotedBlockInRound, true, batch);
+      if (voted_block_hash_with_soft_votes->voted_block_hash != NULL_BLOCK_HASH) {
+        db_->addSoftVotesToBatch(round, voted_block_hash_with_soft_votes->votes, batch);
+      }
+      soft_voted_block_for_this_round_ = {voted_block_hash_with_soft_votes->voted_block_hash, true};
+    } else {
+      db_->addPbftMgrVotedValueToBatch(PbftMgrVotedValue::SoftVotedBlockHashInRound, NULL_BLOCK_HASH, batch);
+      db_->addPbftMgrStatusToBatch(PbftMgrStatus::SoftVotedBlockInRound, false, batch);
+      soft_voted_block_for_this_round_ = {NULL_BLOCK_HASH, false};
```
gaoqi8649, line 733: I think that should be:
if (auto voted_block_hash_with_soft_votes = vote_mgr_->getVotesBundleByRoundAndStep(round, 2, TWO_T_PLUS_ONE)) {
      auto batch = db_->createWriteBatch();
      db_->addPbftMgrVotedValueToBatch(PbftMgrVotedValue::SoftVotedBlockHashInRound,
                                       voted_block_hash_with_soft_votes->voted_block_hash, batch);
      db_->addPbftMgrStatusToBatch(PbftMgrStatus::SoftVotedBlockInRound, true, batch);
      db_->addSoftVotesToBatch(round, voted_block_hash_with_soft_votes->votes, batch);
      db_->commitWriteBatch(batch);
      soft_voted_block_for_this_round_ = {voted_block_hash_with_soft_votes->voted_block_hash, true};
}

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -428,34 +426,43 @@ VotesBundle VoteManager::getVotesBundleByRoundAndStep(uint64_t round, size_t ste
...
+          while (count < two_t_plus_one) {
+            votes.emplace_back(it->second);
+            count += it->second->getWeight().value();
+            it++;
+          }
+          LOG(log_nf_) << "Found enough " << count << " votes at voted value " << voted_value.first << " for round "
+                       << round_rit->first << " step " << step_rit->first;
+          next_votes_manager_->updateNextVotes(votes, two_t_plus_one);
+          return round_rit->first + 1;
+        }
```
gaoqi8649, line 462: I don't understand the part. That's very similar with using getVotesBundleByRoundAndStep. What's benefits about the change?

gaoqi8649:
I don't understand issuehttps://github.com/Taraxa-project/taraxa-node/issues/1464. In data structure verified_votes_, we have table store each step for each voted value, and for each voted value has total votes weight.

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
MatusKysel: I think main idea is to not iterate from step 1,2 ... but reverse that, as the highest step has higher probability to have enough votes.

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
MatusKysel: but before we did changes based on return value of `getVotesBundleByRoundAndStep`. So I do no need to set anything in case we do not have enough votes to specify any round?

gaoqi8649: When PBFT go to a new round(in resetRound_), that will reset soft voted value to NULL. 
https://github.com/Taraxa-project/taraxa-node/blob/develop/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp#L399
https://github.com/Taraxa-project/taraxa-node/blob/develop/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp#L417

 getVotesBundleByRoundAndStep return values or null pointer. If no value return, don't need set NULL

gaoqi8649 (approved):

mfrankovi (approved):

Issue #1603, opened 2022-03-03
Fix rlp call to get actual predicted gas fee


JakubFornadel:

MatusKysel:
So I did some research, we need to have some kind of service, that will based on the previous transactions gas prices calculate price for current transacions

Issue #1473, opened 2022-01-21
Get rid of stupid hex-string values in config files


MatusKysel:
## Task Description

For to me unknown reason we use hax-string values for integers in our configs. it makes reading the config impossible and even parsing those values is more difficult then plain number. We should remove it

PR #1611, opened 2022-03-08
feat: implement basic gas pricer service

MatusKysel:
## Purpose
Fixes: #1603


MatusKysel:
Failing on some-tests, so we need to adjust scripts, but otherwise build should be fine

libraries/core_libs/consensus/include/transaction/gas_pricer.hpp
```diff
@@ -0,0 +1,29 @@
+#pragma once
+
+#include <boost/circular_buffer.hpp>
+
+#include "final_chain/final_chain.hpp"
+
+namespace taraxa {
+
+class GasPricer {
```
JakubFornadel, line 9: maybe a short description of class - what it does, where and how it is used would be nice

libraries/core_libs/consensus/src/transaction/gas_pricer.cpp
```diff
@@ -0,0 +1,87 @@
...
+  auto block_num = *last_blk_num;
+
+  std::unique_lock lock(mutex_);
+
+  while (price_list_.capacity() != price_list_.size() && block_num) {
+    auto trxs = db->getPeriodTransactions(block_num);
+    block_num--;
+
+    // Light node
+    if (!trxs) {
```
JakubFornadel, line 34: not sure if this is ideal, what if you are not light node and getPeriodTransactions does not return the txs for some reason ? In case of full node we could require these txs to exist by assert and in case of light node (we will know it is a light node from config) we could check this

libraries/cli/include/cli/default_config.hpp
```diff
@@ -189,6 +189,10 @@ const char *default_json = R"foo({
         }
       }
     },
+    "gas_price" : {
+      "blocks" : "0xc8",
+      "percentile" : "0x3c"
```
gaoqi8649, line 194: I think using integer will be better

MatusKysel: there is separate task for that https://github.com/Taraxa-project/taraxa-node/issues/1473, so I would keep as it is for now (to have it uniform) and we should change multiple values to make it human readable

libraries/core_libs/consensus/src/transaction/gas_pricer.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+GasPricer::GasPricer(uint64_t percentile, uint64_t number_of_blocks, std::shared_ptr<DbStorage> db)
+    : percentile_(percentile), price_list_(number_of_blocks) {
+  if (db) {
+    init_daemon_ = std::make_unique<std::thread>([this, db_ = std::move(db)]() { init(std::move(db_)); });
+  }
+}
+
+GasPricer::~GasPricer() {
+  if (init_daemon_) init_daemon_->join();
```
JakubFornadel, line 13: if (init_daemon_ && init_daemon_->joinable())

MatusKysel: ? this will not compile

libraries/core_libs/consensus/src/transaction/gas_pricer.cpp
```diff
@@ -0,0 +1,91 @@
...
+    // Empty block
+    if (!trxs->size()) {
+      continue;
+    }
+
+    if (const auto min_trx =
+            *std::min_element(trxs->begin(), trxs->end(),
+                              [](const auto& t1, const auto& t2) { return t1.getGasPrice() < t2.getGasPrice(); });
+        min_trx.getGasPrice()) {
+      price_list_.push_front(std::move(min_trx.getGasPrice()));
```
JakubFornadel, line 50: moving const reference makes a copy so this move should not be here

libraries/core_libs/consensus/src/transaction/gas_pricer.cpp
```diff
@@ -0,0 +1,91 @@
...
+    std::unique_lock lock(mutex_);
+
+    price_list_.push_back(std::move(minimum));
+
+    std::vector<u256> sorted_prices;
+    sorted_prices.reserve(price_list_.size());
+    std::copy(price_list_.begin(), price_list_.end(), std::back_inserter(sorted_prices));
+    std::sort(sorted_prices.begin(), sorted_prices.end());
+
+    if (auto new_price = sorted_prices[(sorted_prices.size() - 1) * percentile_ / 100]) {
```
JakubFornadel, line 85: in case percentile_ is not properly configured, you might try to access non existing index in array here:

- percentile_ should be checked in ctor 
- percentile value should be also validated when loaded from config, do it in void FullNodeConfig::validate()  and throw config exception in case the value is out of boundaries

libraries/core_libs/node/src/node.cpp
```diff
@@ -226,6 +206,39 @@ void FullNode::start() {
         },
         *rpc_thread_pool_);
   }
+  subscription_pool_ = std::make_unique<util::ThreadPool>(1);
```
JakubFornadel, line 209: does this need to be a unique_ptr ?

libraries/core_libs/node/src/node.cpp
```diff
@@ -226,6 +206,39 @@ void FullNode::start() {
...
+
+  // GasPricer updater
+  final_chain_->block_finalized_.subscribe(
+      [gas_pricer = as_weak(gas_pricer_)](auto const &res) {
+        if (auto _gas_pricer = gas_pricer.lock()) {
+          _gas_pricer->update(res->trxs);
+        }
+      },
+      *subscription_pool_);
```
JakubFornadel, line 218: why does block_finalized_ use subscription_pool_ and block_applying_  does not ?

tests/gas_pricer_test.cpp
```diff
@@ -0,0 +1,59 @@
...
+#include <gtest/gtest.h>
+
+namespace taraxa::core_tests {
+
+// Do not use BaseTest from "util_test/gtest.hpp" as its functionality is not needed in this test
+struct BaseTest : virtual testing::Test {
+  testing::UnitTest* current_test = ::testing::UnitTest::GetInstance();
+  testing::TestInfo const* current_test_info = current_test->current_test_info();
+
+  virtual ~BaseTest() {}
```
JakubFornadel, line 12: always use = default instead of {}

libraries/core_libs/node/src/node.cpp
```diff
@@ -226,6 +206,39 @@ void FullNode::start() {
         },
         *rpc_thread_pool_);
   }
+  subscription_pool_ = std::make_unique<util::ThreadPool>(1);
+
+  // GasPricer updater
+  final_chain_->block_finalized_.subscribe(
+      [gas_pricer = as_weak(gas_pricer_)](auto const &res) {
+        if (auto _gas_pricer = gas_pricer.lock()) {
```
JakubFornadel, line 214: I would stick to the naming conventions so custom variables start with letter and not with undescore

libraries/core_libs/node/src/node.cpp
MatusKysel: because block_applying_ is wrongly written

libraries/core_libs/consensus/src/transaction/gas_pricer.cpp
```diff
@@ -0,0 +1,91 @@
...
+void GasPricer::update(const Transactions& trxs) {
+  if (trxs.empty()) return;
+
+  std::vector<u256> prices;
+  prices.reserve(trxs.size());
+
+  std::transform(trxs.begin(), trxs.end(), std::back_inserter(prices),
+                 [](const auto& trx) { return trx.getGasPrice(); });
+
+  if (u256 minimum = *std::min_element(prices.begin(), prices.end())) {
```
JakubFornadel, line 75: isnt it possible to use min_element with lambda so you could skip whole prices vector creation ?

libraries/core_libs/consensus/include/transaction/gas_pricer.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+ private:
+  /**
+   * @brief initialize values on start from blocks in DB
+   *
+   * @param db
+   */
+  void init(std::shared_ptr<DbStorage> db);
+
+  const uint64_t percentile_;
```
JakubFornadel, line 40: k prefix

libraries/core_libs/node/src/node.cpp
MatusKysel: not until we add number of threads for this pool to config. I will change it and add comment there

libraries/core_libs/consensus/include/transaction/gas_pricer.hpp
```diff
@@ -0,0 +1,49 @@
...
+
+  mutable std::shared_mutex mutex_;
+  u256 latest_price_ = 1;
+  boost::circular_buffer<u256> price_list_;
```
JakubFornadel, line 44: question is: do we want to use boost in general or do we want to minimize its usage ? I believe that except the dag we could get rid of boost pretty much everywhere if we wanted to, which would be cool

JakubFornadel (approved):

libraries/core_libs/consensus/include/transaction/gas_pricer.hpp
MatusKysel: ? we use boost almost everywhere. Our logging is boost, threads (std does not have upgradable lock), asio, networking ...

libraries/core_libs/node/src/node.cpp
JakubFornadel: is it too difficult to adjust it so all event use this subscription_pool_ ?

MatusKysel: probably not, but I do not want to touch HF as it is sensitive, I would rather create separate task for it. I think it needs a lot of testing that is out of the scope of this task

libraries/core_libs/consensus/src/transaction/gas_pricer.cpp
MatusKysel: this only applies if you want to join current thread (that's not joinable), that's in this case impossible

libraries/core_libs/consensus/include/transaction/gas_pricer.hpp
JakubFornadel: ok nevermind

libraries/core_libs/consensus/src/transaction/gas_pricer.cpp
JakubFornadel: not true, thread is not joinable also if you moved it, detached or joined before, which is possible in case someone else changes the code and forgets about this

libraries/core_libs/node/src/node.cpp
MatusKysel: https://github.com/Taraxa-project/taraxa-node/issues/1644

PR #1615, opened 2022-03-09
Release/v1.1.0 candidate new

MatusKysel:
## Purpose
Fixes: #1613 

Mainnet HF is prepared for Monday (14th of March) ~ 12:00 CET time. Calculated based on block every 6s


rattrap (approved):
LGTM

PR #1633, opened 2022-03-17
Missing transaction fix

mfrankovi:
Removing seen_txs_ cache because it was causing the issue where we would mark transaction as known before it was actually part of the transaction pool or database, so transaction would be marked as known but it was actually missing causing a race condition which led to the missing transaction error and incorrectly marking peer as malicious.

I believe that this cache is not really needed but if we find some performance issues we might think of some smarter was to create some cache of this type. For now this does fix the issue.

The actual race condition causing the error:
Node 1 sends transaction T to Node 3
Node 2 sends the same transaction T to node 3
Node 2 sends dag block D that depends on transaction T to node 3
Node 3 receives message from Node 1 and marks transaction as known before inserting it in the pool
Node 3 receives transactions from Node 2 and since transactions is "known" discards it
Node 3 receives dag block D before transaction is actually in the memory pool and reports this error

### DIFF
--- a/libraries/core_libs/consensus/include/transaction_manager/transaction_manager.hpp
+++ b/libraries/core_libs/consensus/include/transaction_manager/transaction_manager.hpp
@@ -52,14 +52,6 @@ class TransactionManager : public std::enable_shared_from_this<TransactionManage
    */
   uint32_t insertValidatedTransactions(const SharedTransactions &txs);
 
-  /**
-   * @brief Marks transaction as seen and returns if was seen before
-   *
-   * @param trx_hash transaction hash
-   * @return true if seen
-   */
-  bool markTransactionSeen(const trx_hash_t &trx_hash);
-
   size_t getTransactionPoolSize() const;
 
   size_t getNonfinalizedTrxSize() const;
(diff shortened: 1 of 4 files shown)


libraries/core_libs/consensus/src/transaction_manager/transaction_manager.cpp
```diff
@@ -169,6 +164,15 @@ std::shared_ptr<Transaction> TransactionManager::getTransaction(trx_hash_t const
   return db_->getTransaction(hash);
 }
 
+bool TransactionManager::isTransactionKnown(trx_hash_t const &hash) const {
+  std::shared_lock transactions_lock(transactions_mutex_);
+  auto trx = transactions_pool_.get(hash);
+  if (trx) {
+    return true;
+  }
+  return db_->transactionInDb(hash);
```
JakubFornadel, line 173: you could use this code instead:
```
std::shared_lock transactions_lock(transactions_mutex_);
if (transactions_pool_.contains(hash)) {
  return true;
}

return db_->transactionInDb(hash);

auto trx = transactions_pool_.get(hash);
if (trx) {
  return true;
}
return db_->transactionInDb(hash);
```

libraries/core_libs/consensus/include/transaction_manager/transaction_manager.hpp
```diff
@@ -108,6 +100,15 @@ class TransactionManager : public std::enable_shared_from_this<TransactionManage
...
 
   std::shared_ptr<Transaction> getTransaction(trx_hash_t const &hash) const;
+
+  /**
+   * @brief returns if transaction seen and verified
+   *
+   * @param trx_hash transaction hash
+   * @return true if seen
+   */
+  bool isTransactionKnown(trx_hash_t const &hash) const;
```
JakubFornadel, line 110: const trx_hash_t

gaoqi8649 (approved):

kstdl (approved):

PR #1634, opened 2022-03-17
fix(tx_pool): mark tx as known only after is is verified and inserted…

JakubFornadel:
… in the tx pool or db


mfrankovi (approved):

gaoqi8649 (approved):

MatusKysel (commented):
Actually this is not good fix, you have created bug, when you do not mark non-valid transaction as know. Even transaction that fails validation should be mark as know

MatusKysel:
> Actually this is not good fix, you have created bug, when you do not mark non-valid transaction as know. Even transaction that fails validation should be mark as know

Never mind, this is not needed

PR #1637, opened 2022-03-18
Merge v1.0.1 to develop

MatusKysel:


libraries/cli/include/cli/default_config.hpp
```diff
@@ -207,7 +208,7 @@ const char *default_json = R"foo({
           "111f91441efc8c6c0edf6534970cc887e2fabaa8": "0x24048ce3d"
         },
         "hardforks": {
-          "fix_genesis_fork_block": "0x0"
+          "fix_genesis_fork_block": "0x102CA0"
```
gaoqi8649, line 211: Do we need change this? Since we have mainnet config already, I think we only need change there

libraries/cli/include/cli/devnet_config.hpp
```diff
@@ -1,6 +1,7 @@
 namespace taraxa::cli {
 
 const char *devnet_json = R"foo({
+  "version" : 1,
```
JakubFornadel, line 4: I believe we removed this

PR #1638, opened 2022-03-18
V1.0.1 conflict

MatusKysel:
## Purpose
Intermediary branch to resolve conflict between master v1.0.1 and develop after that we could merge https://github.com/Taraxa-project/taraxa-node/pull/1637


gaoqi8649 (approved):

JakubFornadel (approved):

JakubFornadel (changes requested):
Why are there 2 PR's with the same functionality ? https://github.com/Taraxa-project/taraxa-node/pull/1637

MatusKysel:
> Why are there 2 PR's with the same functionality ? #1637

as we can not change master branch, I had to create this branch and resolve conflicts. But before that I have had created branch 1637 as I did not know where are that many conflict there. If we merge this one 1637 should automatically close itself as the  code will be the same as from this one. I think this is the flow we should use as we can track it back

mfrankovi (approved):

JakubFornadel (approved):

PR #1649, opened 2022-03-23
feat: add network id to discovery packet

MatusKysel:
## Purpose
Fixes: #1646


JakubFornadel (approved):

libraries/cli/include/cli/devnet_config.hpp
```diff
@@ -32,21 +32,6 @@ const char *devnet_json = R"foo({
       "id": "8a919928e2ee8908d3ef7cf49f245b79d228a86fd2c1c0b8c687007e62b5777aff11e748d0b29fcebd321780ce9e0544717a4c221271252a66bd50b3db6eb532",
       "ip": "boot-node-2.devnet.taraxa.io",
       "udp_port": 10002
-    },
```
mfrankovi, line 35: so the expectation is that devnet should work even without specifying taraxa nodes?

mfrankovi (approved):

libraries/cli/include/cli/devnet_config.hpp
MatusKysel: Yes, as we now propagate nodes over the discovery only if the chain id is same

JakubFornadel (approved):

PR #1656, opened 2022-03-24
fix: efficiency cache

mfrankovi:
Efficiency was not correctly populated because ignored_efficiency_counter_ was not getting reset after each changing_interval. This made the calculation be based on full changing_interval and not just computation_interval. On restart it was reset which made nodes that restarted have different efficiency data which corrupted difficulty calculation.

The fix is only reseting the ignored_efficiency_counter_ to 0 but it also contains added tests to make sure that both efficiencies and params_changes_ that are kept in memory have same values before and after the restart.

### DIFF
--- a/libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
+++ b/libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
@@ -49,6 +49,12 @@ class SortitionParamsManager {
    */
   uint16_t averageDagEfficiency();
 
+  /**
+   * Get memory params changes, used only for unit tests
+   * @returns params changes
+   */
+  const std::deque<SortitionParamsChange>& getParamsChanges() const { return params_changes_; }
+
  protected:
   SortitionConfig config_;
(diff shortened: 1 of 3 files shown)


libraries/core_libs/consensus/include/dag/sortition_params_manager.hpp
```diff
@@ -49,6 +49,12 @@ class SortitionParamsManager {
    */
   uint16_t averageDagEfficiency();
 
+  /**
+   * Get memory params changes, used only for unit tests
+   * @returns params changes
+   */
+  std::deque<SortitionParamsChange> getParamsChanges() const { return params_changes_; }
```
JakubFornadel, line 56: this could maybe return `const std::deque<SortitionParamsChange>&`

JakubFornadel (approved):

MatusKysel (approved):

MatusKysel:
let's deploy it ASAP on devnet and make some restarts to prove, it is ok

PR #1658, opened 2022-03-25
efficiency target adjust to 80%

mfrankovi:
Efficiency target adjust to 80%

Adjusted the algorithm to avoid large changes when close to the target efficiency to use 1% of the current difficulty value and not 1% of the full range.

Adjusted calculating deviation from the target efficiency to account for 80% target being close to 100% to the base value is the range between the target and 0%-100%. New calculation will give exactly same result for 50% target.


libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -125,11 +125,24 @@ void SortitionParamsManager::pbftBlockPushed(const SyncBlock& block, DbStorage::
   }
 }
 
-int32_t getThresholdChange(uint16_t efficiency, uint16_t target_efficiency) {
-  uint16_t deviation = std::abs(efficiency - target_efficiency) * 100 / target_efficiency;
-  // If goal is 50 % return 1% change for 40%-60%, 2% change for 30%-70% and 5% change for over that
+int32_t getThresholdChange(uint16_t efficiency, uint16_t target_efficiency, const int32_t current_threshold) {
+  uint16_t deviation = 0;
+  // Deviation is measured in percentage of the range betwwen target efficiency to 0% or 100% depending if the
```
MatusKysel, line 130: typo `betwwen`

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
```diff
@@ -125,11 +125,24 @@ void SortitionParamsManager::pbftBlockPushed(const SyncBlock& block, DbStorage::
   }
 }
 
-int32_t getThresholdChange(uint16_t efficiency, uint16_t target_efficiency) {
-  uint16_t deviation = std::abs(efficiency - target_efficiency) * 100 / target_efficiency;
-  // If goal is 50 % return 1% change for 40%-60%, 2% change for 30%-70% and 5% change for over that
+int32_t getThresholdChange(uint16_t efficiency, uint16_t target_efficiency, const int32_t current_threshold) {
```
MatusKysel, line 128: no need for const

MatusKysel (approved):

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/dag/sortition_params_manager.cpp
mfrankovi: Fixed

kstdl (approved):

Issue #1659, opened 2022-03-28
Segmentation fault - latest testnet


MatusKysel:
I have noticed core-dumps on latest testnet branch 

```
node_1        | 0x00007fdafb7fe700 SUMMARY [2022-03-24 16:22:05.734887] INFO: ##################################
node_1        | Segmentation fault (core dumped)
node_1        | 0x00007f5060e32300 FULLND [2022-03-24 16:23:40.986706] SILENT: Node public key: a080dd9a388c3f8ff146f20b32fd3b7e7a8d7a200bec1c5efa6fabba46c225761f21a31172600b9e64fef9c2548a31b8eff96267c45952b0f79dbbe189df5512        
```

PR #1660, opened 2022-03-28
fix: improve async cancellation

MatusKysel:
## Purpose
Fixes: #1659 
Please check also https://github.com/Taraxa-project/taraxa-vdf/pull/9

I can not say for sure if this was the case, but I know exact spot. So I think flow was like this:

1. We execute std::sync
2. async is still not executed ...
3. some of the condition applied
4. we want to cancel execution but pointer is empty
5. Segfault

### DIFF
--- a/libraries/vdf/include/vdf/sortition.hpp
+++ b/libraries/vdf/include/vdf/sortition.hpp
@@ -27,8 +27,12 @@ class VdfSortition : public vrf_wrapper::VrfSortitionBase {
   explicit VdfSortition(bytes const& b);
   explicit VdfSortition(Json::Value const& json);
 
-  void computeVdfSolution(SortitionParams const& config, dev::bytes const& msg);
-  void cancelCompute();
+  // Only for testing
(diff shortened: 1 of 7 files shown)


kstdl:
I think we should also have a PR in the taraxa-vdf repo. As I see develop branch right now is not pointing to master branch of this repo

kstdl (approved):

libraries/vdf/include/vdf/sortition.hpp
```diff
@@ -27,8 +27,12 @@ class VdfSortition : public vrf_wrapper::VrfSortitionBase {
   explicit VdfSortition(bytes const& b);
   explicit VdfSortition(Json::Value const& json);
 
-  void computeVdfSolution(SortitionParams const& config, dev::bytes const& msg);
-  void cancelCompute();
+  // Only for testing
+  void computeVdfSolution(const SortitionParams& config, const bytes& msg);
```
JakubFornadel, line 31: What is this function good for ? Why is not computeVdfSolutionCancellable used in tests ?

Issue #1619, opened 2022-03-10
Protect PBFT blocks syncing


gaoqi8649:
Protect PBFT blocks syncing from malicious players in both side:
1. Peers keep asking for same period block again and again 
2. Peers send same period blocks again and again

PR #1666, opened 2022-03-29
feature: protect syncing PBFT blocks

gaoqi8649:
Fix https://github.com/Taraxa-project/taraxa-node/issues/1619

Protect PBFT blocks syncing from malicious players in both side:

Peers keep asking for same period block again and again
Peers send same period blocks again and again

### DIFF
--- a/libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
@@ -69,27 +69,18 @@ class PbftSyncingState {
    *
    * @param peer_id
    * @param period peer asking sync period
-   *
-   * @return true if table not include peer or asking sync period greater than saving value or has excessed the
-   * kPeerAskingPeriodTimeout for peer restart/rebuild DB situation. Otherwise return false.
    */
-  bool updatePeerAskingPeriod(const dev::p2p::NodeID& peer_id, uint64_t period);
+  void updatePeerAskingPeriod(const dev::p2p::NodeID& peer_id, uint64_t period);
 
   /**
    * @brief Update peers sending sync blocks period
    *
    * @param peer_id
    * @param period peer sending sync block period
-   *
-   * @return true if table not include peer or sending sync block period greater than saving value. Otherwise return
-   * false
    */
-  bool updatePeerSyncingPeriod(const dev::p2p::NodeID& peer_id, uint64_t period);
+  void updatePeerSyncingPeriod(const dev::p2p::NodeID& peer_id, uint64_t period);
 
  private:
-  using UpgradableLock = boost::upgrade_lock<boost::shared_mutex>;
-  using UpgradeLock = boost::upgrade_to_unique_lock<boost::shared_mutex>;
-
   std::atomic<bool> deep_pbft_syncing_{false};
   std::atomic<bool> pbft_syncing_{false};
 
--- a/libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
@@ -64,7 +64,32 @@ class PbftSyncingState {
    */
   const dev::p2p::NodeID syncingPeer() const;
 
+  /**
+   * @brief Update peers asking sync period table
+   *
+   * @param peer_id
+   * @param period peer asking sync period
+   *
+   * @return true if table not include peer or asking sync period greater than saving value or has excessed the
+   * kPeerAskingPeriodTimeout for peer restart/rebuild DB situation. Otherwise return false.
+   */
+  bool updatePeerAskingPeriod(const dev::p2p::NodeID& peer_id, uint64_t period);
+
+  /**
+   * @brief Update peers sending sync blocks period
+   *
+   * @param peer_id
+   * @param period peer sending sync block period
+   *
+   * @return true if table not include peer or sending sync block period greater than saving value. Otherwise return
+   * false
+   */
+  bool updatePeerSyncingPeriod(const dev::p2p::NodeID& peer_id, uint64_t period);
+
  private:
+  using UpgradableLock = boost::upgrade_lock<boost::shared_mutex>;
+  using UpgradeLock = boost::upgrade_to_unique_lock<boost::shared_mutex>;
+
   std::atomic<bool> deep_pbft_syncing_{false};
   std::atomic<bool> pbft_syncing_{false};
 
@@ -73,6 +98,10 @@ class PbftSyncingState {
   // Number of seconds needed for ongoing syncing to be declared as inactive
   static constexpr std::chrono::seconds kSyncingInactivityThreshold{60};
 
+  // Number of milliseconds for peer asking sync period timeout
+  // corner case restart/rebuild DB may ask previous periods again
+  const std::chrono::milliseconds kPeerAskingPeriodTimeout{1000};
+
   // What time was received last syncing packet
   std::chrono::steady_clock::time_point last_received_sync_packet_time_{std::chrono::steady_clock::now()};
   mutable std::shared_mutex time_mutex_;
@@ -80,6 +109,23 @@ class PbftSyncingState {
   // Peer that the node is syncing with
   std::shared_ptr<TaraxaPeer> peer_;
   mutable std::shared_mutex peer_mutex_;
+
+  /**
+   * @brief Table for peer asking sync period
+   *        Table key: peer node ID, value: a pair
+   *        Pair key: peer last asking request period, pair value: peer last asking request time point
+   */
+  std::unordered_map<dev::p2p::NodeID, std::pair<uint64_t, std::chrono::steady_clock::time_point>>
+      asking_period_peers_table_;
+  mutable boost::shared_mutex asking_period_mutex_;
+
+  /**
+   * @brief Table for incoming last PBFT blocks period for syncing per each peer.
+   *        Key: peer node ID
+   *        Value: last PBFT block period number
+   */
+  std::unordered_map<dev::p2p::NodeID, uint64_t> incoming_blocks_period_peers_table_;
+  mutable boost::shared_mutex incoming_blocks_mutex_;
 };
 
 }  // namespace taraxa::network::tarcap
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
@@ -32,6 +32,12 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
   LOG(log_tr_) << "Received GetPbftSyncPacket Block";
 
   const size_t height_to_sync = packet_data.rlp_[0].toInt();
+  if (!pbft_syncing_state_->updatePeerAskingPeriod(packet_data.from_node_id_, height_to_sync)) {
+    std::ostringstream err_msg;
+    err_msg << "Peer " << packet_data.from_node_id_ << " request syncing asked previoud period " << height_to_sync;
+    throw MaliciousPeerException(err_msg.str());
+  }
+
   // Here need PBFT chain size, not synced period since synced blocks has not verified yet.
   const size_t my_chain_size = pbft_chain_->getPbftChainSize();
   if (height_to_sync > my_chain_size) {
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
@@ -32,10 +32,10 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
   LOG(log_tr_) << "Received GetPbftSyncPacket Block";
 
   const size_t height_to_sync = packet_data.rlp_[0].toInt();
-  if (!pbft_syncing_state_->updatePeerAskingPeriod(packet_data.from_node_id_, height_to_sync)) {
-    std::ostringstream err_msg;
-    err_msg << "Peer " << packet_data.from_node_id_ << " request syncing asked previoud period " << height_to_sync;
-    throw MaliciousPeerException(err_msg.str());
+  try {
+    pbft_syncing_state_->updatePeerAskingPeriod(packet_data.from_node_id_, height_to_sync);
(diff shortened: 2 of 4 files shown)


libraries/core_libs/network/src/tarcap/shared_states/pbft_syncing_state.cpp
```diff
@@ -65,4 +65,32 @@ bool PbftSyncingState::isPbftSyncing() {
...
+  }
+
+  UpgradeLock locked(lock);
+  asking_period_peers_table_[peer_id] = std::make_pair(period, now);
+
+  return true;
+}
+
+bool PbftSyncingState::updatePeerSyncingPeriod(const dev::p2p::NodeID& peer_id, uint64_t period) {
+  UpgradableLock lock(incoming_blocks_mutex_);
```
MatusKysel, line 85: UpgradableLock + UpgradeLock design make sense only when there is another "reader" (shared_lock) otherwise there is no gain, as UpgradableLock can not be locked 2x. So I would get rid of it and keep it simple with unique_lock

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -32,6 +32,12 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
   LOG(log_tr_) << "Received GetPbftSyncPacket Block";
 
   const size_t height_to_sync = packet_data.rlp_[0].toInt();
+  if (!pbft_syncing_state_->updatePeerAskingPeriod(packet_data.from_node_id_, height_to_sync)) {
+    std::ostringstream err_msg;
+    err_msg << "Peer " << packet_data.from_node_id_ << " request syncing asked previoud period " << height_to_sync;
```
MatusKysel, line 37: `previoud` typo, but msg should be more clear like as it could be timeout or the height that caused this condition

MatusKysel: and anther thing, is this even error for this node? I would say that it is warning. MaliciousPeerException -> logs error

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -59,6 +53,21 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
     throw MaliciousPeerException("Unable to parse SyncBlock: " + std::string(e.what()));
   }
 
+  auto sync_period = sync_block.pbft_blk->getPeriod();
```
MatusKysel, line 56: const

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -59,6 +53,21 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
...
+
+  if (!pbft_syncing_state_->updatePeerSyncingPeriod(packet_data.from_node_id_, sync_period)) {
+    std::ostringstream err_msg;
+    err_msg << "PbftSyncPacket received from peer " << packet_data.from_node_id_ << " sending previous period "
+            << sync_period << " PBFT block";
+    throw MaliciousPeerException(err_msg.str());
+  }
+
+  if (pbft_syncing_state_->syncingPeer() != packet_data.from_node_id_) {
```
MatusKysel, line 65: this should be first check before `updatePeerSyncingPeriod` and maybe we should also mark peer as malicious?

libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
```diff
@@ -73,13 +98,34 @@ class PbftSyncingState {
...
+      asking_period_peers_table_;
+  mutable boost::shared_mutex asking_period_mutex_;
+
+  /**
+   * @brief Table for incoming last PBFT blocks period for syncing per each peer.
+   *        Key: peer node ID
+   *        Value: last PBFT block period number
+   */
+  std::unordered_map<dev::p2p::NodeID, uint64_t> incoming_blocks_period_peers_table_;
+  mutable boost::shared_mutex incoming_blocks_mutex_;
```
MatusKysel, line 128: I do not think we need this table, as you can sync only from one node = one node id

libraries/core_libs/network/src/tarcap/shared_states/pbft_syncing_state.cpp
```diff
@@ -65,4 +65,32 @@ bool PbftSyncingState::isPbftSyncing() {
...
+  if (asking_period_peers_table_.contains(peer_id) && asking_period_peers_table_[peer_id].first >= period &&
+      std::chrono::duration_cast<std::chrono::milliseconds>(now - asking_period_peers_table_[peer_id].second) <=
+          kPeerAskingPeriodTimeout) {
+    return false;
+  }
+
+  UpgradeLock locked(lock);
+  asking_period_peers_table_[peer_id] = std::make_pair(period, now);
+
+  return true;
```
MatusKysel, line 81: you should be cleaning this table on disconnect otherwise it will have a lot of old data. Or maybe when the syncing is finished

libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
MatusKysel: so you should use existing `std::shared_ptr<TaraxaPeer> peer_` and maybe just add period and use existing mutex too

MatusKysel (approved):

libraries/core_libs/network/src/tarcap/shared_states/pbft_syncing_state.cpp
JakubFornadel: I would say we should not use this pattern at all...

libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
```diff
@@ -64,7 +64,32 @@ class PbftSyncingState {
...
+   */
+  bool updatePeerAskingPeriod(const dev::p2p::NodeID& peer_id, uint64_t period);
+
+  /**
+   * @brief Update peers sending sync blocks period
+   *
+   * @param peer_id
+   * @param period peer sending sync block period
+   *
+   * @return true if table not include peer or sending sync block period greater than saving value. Otherwise return
```
JakubFornadel, line 84: true if the table does not include ..... greater than <period> value

libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
```diff
@@ -64,7 +64,32 @@ class PbftSyncingState {
    */
   const dev::p2p::NodeID syncingPeer() const;
 
+  /**
+   * @brief Update peers asking sync period table
+   *
+   * @param peer_id
+   * @param period peer asking sync period
+   *
+   * @return true if table not include peer or asking sync period greater than saving value or has excessed the
```
JakubFornadel, line 73: true if the table does not include ..... greater than value

libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
```diff
@@ -73,13 +98,34 @@ class PbftSyncingState {
...
   mutable std::shared_mutex peer_mutex_;
+
+  /**
+   * @brief Table for peer asking sync period
+   *        Table key: peer node ID, value: a pair
+   *        Pair key: peer last asking request period, pair value: peer last asking request time point
+   */
+  std::unordered_map<dev::p2p::NodeID, std::pair<uint64_t, std::chrono::steady_clock::time_point>>
+      asking_period_peers_table_;
+  mutable boost::shared_mutex asking_period_mutex_;
```
JakubFornadel, line 120: I agree with Matus, even this could be part of TaraxaPeer class - if you mark some peer as malicious due to this functionality, he wont be able to reconnect anyway so why would you keep this in some separate table that you have to clean separately, etc... ?

JakubFornadel (commented):
So if I remember correctly, this PR was created because you noticed we are sending too many duplicate pbft sync requests ? If so, this is not a solution as you did not touch part of code, which is actually triggering this behaviour ?

libraries/core_libs/network/src/tarcap/shared_states/pbft_syncing_state.cpp
MatusKysel: not sure about that, there are several places in our code where "heavy" computing is done in `UpgradableLock` and there is no need to completely lock it as it blocks all the "reader" meanwhile

gaoqi8649: Change to std::unique_lock

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
gaoqi8649: Fix typo and error message. Thanks
This should be error because peer sending previous asked period within kPeerAskingPeriodTimeout

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
gaoqi8649: Fixed, thanks

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
gaoqi8649: I think this should not before updatePeerSyncingPeriod

There is an example:
1. A node1 first sync with node2, node2 send PBFT blocks B1, B2, B3 to node1
2. Node1 received B1 and B2. But B3 take a long time to deliver(maybe because network delay or something else caused B3 cannot deliver on time)
3. Node1 switch to sync with node3, node3 send B3, B4 to node1
4. Node1 receive B3, B4 from node3
5. node1 receive B3 from node2
In the case, node3 from node2 is valid, node2 is not malicious.

libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
gaoqi8649: In the same example:
1. A node1 first sync with node2, node2 send PBFT blocks B1, B2, B3 to node1
2. Node1 received B1 and B2. But B3 take a long time to deliver(maybe because network delay or something else caused B3 cannot deliver on time)
3. Node1 switch to sync with node3, node3 send B3, B4 to node1
4. Node1 receive B3, B4 from node3
5. node1 receive B3 from node2

I think we need save all peers that the node has synced with from/to

libraries/core_libs/network/src/tarcap/shared_states/pbft_syncing_state.cpp
gaoqi8649: You are right. Nodes will not connect again after mark malicious. Erase the malicious peers from table

libraries/core_libs/network/src/tarcap/shared_states/pbft_syncing_state.cpp
```diff
@@ -65,4 +65,32 @@ bool PbftSyncingState::isPbftSyncing() {
   return pbft_syncing_;
 }
 
+bool PbftSyncingState::updatePeerAskingPeriod(const dev::p2p::NodeID& peer_id, uint64_t period) {
+  auto now = std::chrono::steady_clock::now();
+
+  UpgradableLock lock(asking_period_mutex_);
```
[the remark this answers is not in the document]
gaoqi8649: Fixed, thanks

libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
gaoqi8649: I put the tables in PbftSyncingState because I want to make pbft syncing components in the same module

gaoqi8649:
> So if I remember correctly, this PR was created because you noticed we are sending too many duplicate pbft sync requests ? If so, this is not a solution as you did not touch part of code, which is actually triggering this behaviour ?

The PR is not for sending too many duplicate PBFT syncing requests. I remember the sending too many duplicate PBFT syncing requests issue had been fixed by @mfrankovi 
The PR is only for protect PBFT syncing from malicious players

libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
MatusKysel: ok, then you need to clean this cache, or create expiration map here

gaoqi8649: clean the cache when mark as malicious

libraries/core_libs/network/include/network/tarcap/shared_states/pbft_syncing_state.hpp
JakubFornadel: I strongly disagree. It is peer's property and it should be in Peer class + we would not need to create any additional containers that have to be cleaned etc...

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -32,6 +32,12 @@ void GetPbftSyncPacketHandler::process(const PacketData &packet_data,
   LOG(log_tr_) << "Received GetPbftSyncPacket Block";
 
   const size_t height_to_sync = packet_data.rlp_[0].toInt();
+  try {
+    pbft_syncing_state_->updatePeerAskingPeriod(packet_data.from_node_id_, height_to_sync);
+  } catch (const std::logic_error &e) {
```
JakubFornadel, line 37: No need to use exception if your use case is basically checking return value of one specific function...  You could rarther check just returned bool flag

JakubFornadel (changes requested):
Remove both `incoming_blocks_mutex_` & `asking_period_mutex_` and their locks. Instead, create 2 atomic flags inside Peer class that will have exactly the same functionality without need to clean these containers, lock them, etc...

PR #1668, opened 2022-03-30
chore: mark transactions correctly

MatusKysel:
## Purpose
This should improve test stability, as right now we are not marking transactions as seen correclty
Now we mark them

- when transaction is validate
- when transaction is part of DAG
- when transaction is part of PBFT

### DIFF
--- a/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
@@ -194,6 +194,7 @@ void TransactionManager::saveTransactionsFromDagBlock(SharedTransactions const &
   for (uint64_t i = 0; i < trxs.size(); i++) {
     auto const &trx_hash = trx_hashes[i];
     // We only save transaction if it has not already been saved
+    markTransactionKnown(trx_hash);
     if (!trx_in_db[i]) {
       db_->addTransactionToBatch(*trxs[i], write_batch);
       nonfinalized_transactions_in_dag_.emplace(trx_hash, trxs[i]);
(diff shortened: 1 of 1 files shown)


libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -132,11 +132,9 @@ uint32_t TransactionManager::insertValidatedTransactions(SharedTransactions &&tx
   // Check the db with a multiquery if transactions are really new
   auto db_seen = db_->transactionsInDb(txs_hashes);
   for (uint32_t i = 0; i < txs_hashes.size(); i++) {
+    markTransactionKnown(txs[i]->getHash());
```
JakubFornadel, line 135: you cant do this, whole race condition that Miho found out was about marking tx as seen only after it is validated and inside tx_pool - at this point, tx is not inside the pool

MatusKysel: this is ok, it is under mutex so all those transactions are either in DB or will be in pool

gaoqi8649 (approved):

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -194,6 +194,7 @@ void TransactionManager::saveTransactionsFromDagBlock(SharedTransactions const &
   for (uint64_t i = 0; i < trxs.size(); i++) {
     auto const &trx_hash = trx_hashes[i];
     // We only save transaction if it has not already been saved
+    markTransactionKnown(trx_hash);
```
mfrankovi, line 197: I don't think this is needed because these transactions should have already been marked as known. these are transactions which we are pulling from our structures and not something we received over network

PR #1669, opened 2022-03-31
Improve tests

MatusKysel:
## Purpose

- Improved execution of one problematic network_test
- fix condition for nonce checking, after lot of debugging I found out that EVM does not return nonce, but number of transactions execute so basically lastNonce - 1


mfrankovi (approved):

MatusKysel:
there is still one test failing `node_sync2` but let's have this in develop, so we can reset devnet and meanwhile I will look at that test

Issue #1639, opened 2022-03-18
Marking peer maicious on failed transaction


MatusKysel:
Currently we are marking peer as malicious when transaction validation fails, but this could happened if one of the nodes (send or receiver) could be out of sync and in it's view state is different and transaction is valid. For example:

- insufficient funds
- nonce value

MatusKysel:
https://github.com/Taraxa-project/taraxa-node/pull/1714

PR #1671, opened 2022-04-01
feat: DAG blocks could end up with old transactions and we should let…

MatusKysel:
… it failed on execution

## Purpose
Fixes: #1639 but it also stabilize a lot of tests

### DIFF
--- a/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
@@ -21,41 +21,42 @@ TransactionManager::TransactionManager(FullNodeConfig const &conf, std::shared_p
   }
 }
 
-std::pair<bool, std::string> TransactionManager::verifyTransaction(const std::shared_ptr<Transaction> &trx) const {
+std::pair<TransactionStatus, std::string> TransactionManager::verifyTransaction(
+    const std::shared_ptr<Transaction> &trx) const {
   // ONLY FOR TESTING
   if (!final_chain_) [[unlikely]] {
-    return {true, ""};
+    return {TransactionStatus::Verified, ""};
   }
 
   if (trx->getChainID() != conf_.chain.chain_id) {
-    return {false, "chain_id mismatch"};
+    return {TransactionStatus::Invalid, "chain_id mismatch"};
   }
 
   // Ensure the transaction doesn't exceed the current block limit gas.
   if (FinalChain::GAS_LIMIT < trx->getGas()) {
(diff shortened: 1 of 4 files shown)


libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -21,41 +21,42 @@ TransactionManager::TransactionManager(FullNodeConfig const &conf, std::shared_p
...
-    return {false, "invalid signature"};
+    return {TransactionStatus::Invalid, "invalid signature"};
   }
 
   const auto account = final_chain_->get_account(trx->getSender()).value_or(taraxa::state_api::ZeroAccount);
 
   // Ensure the transaction adheres to nonce ordering
   if (account.nonce && account.nonce > trx->getNonce()) {
-    return {false, "nonce too low"};
+    return {TransactionStatus::Old, "nonce too low"};
```
gaoqi8649, line 50: In the case, malicious players could send lots of transactions with old nonce number, cause network threads pool out of memory.

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -21,41 +21,42 @@ TransactionManager::TransactionManager(FullNodeConfig const &conf, std::shared_p
...
   }
 
   // Transactor should have enough funds to cover the costs
   // cost == V + GP * GL
   if (account.balance < trx->getCost()) {
-    return {false, "insufficient balance"};
+    return {TransactionStatus::Old, "insufficient balance"};
```
gaoqi8649, line 56: Same case, I think malicious players could send lots of fake transactions with much high cost for DDOS attack.

gaoqi8649 (approved):

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
JakubFornadel, line 50: This tx is `state - invalid`, which could be simply because tx creator did send invalid tx or it is due to dag(tx is old). I would not call it old as it rules out the other scenario

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
JakubFornadel, line 56: So I am not sure if this is the only solution (could be due to dag). If it is, we should probably require at least have enough funds to pay for fees, otherwise an attacker might "attack" us through free txs ?

Issue #1496, opened 2022-01-31
Limit pbft & dag blocks size


JakubFornadel:
## Task Description

Limit the size of PBFT blocks to prevent massive blocks after network stalling / partitioning by:

- max total size
- max number of dag blocks in pbft block
- max total number of txs in pbft block(related to number of dag blocks)
- max used gas for included txs

Note: It is possible that we create such huge pbft block, which will fail inside evm due to "gas limit". If there is gas limit on blocks, others limits (txs number/size) might not be needed anymore ?

MatusKysel:
I still think about this crazy idea, that we would not calculate gas estimate before executing transaction, but simply use gas limit. In case you will have not enough trx will fail and when you set limit to really high number then we would move unused gas to reward pool or something like that. This way it would be much easier to handle transactions

kstdl:
My current idea is to estimate transaction gas limit during dag block creation. During it we will know dag level and we can get proposal period from the mapping. We will launch estimation(transaction execution) with proposal period as block number, but not finalized block number. Because of this we will be consistent in calculations and could verify it. But we will have at least one not big problem that is incorrect estimations of some calls. It could happen because proposal period may lag behind the finalized period. So could be a situation when contract was already created at block 100, but most recent call of that contract estimates on 80 proposal period and we try to call non-existing contract in that state. Another one thing is that values could change with the state and block. Because it could affect only that our limit mechanism and its estimation, but not the execution and other stuff I guess it is not a problem and it should not lead to big underestimates of transactions. 

Other option there is to calculate limit just by gas_limit from the transaction as @MatusKysel offers. But the issue with it is that you can set any gaslimit to transaction. We can't check that you used estimate method, but not just set it manually. And with default logic you will spend only used part, but not the estimated. So you will receive some odd money after execution. Also if transaction fails it will use all gaslimit. All allocated funds will be spent on the transaction. This leaves room for malicious behavior, including bandwidth throttling

MatusKysel:
I see your point @kstdl mine idea would be only working in case we do not need to be 100% same as ethereum, that we would not return any gas that won't be used. Anyway I do not see room for malicious player, as any attempts will punished by removing fees from his balance

kstdl:
[...]

PR #1672, opened 2022-04-04
feat: limit dag and pbft blocks size with gas estimations

kstdl:
#1496


CMakeModules/cppcheck.cmake
```diff
@@ -14,6 +14,7 @@ else ()
             --suppress=useStlAlgorithm
             --suppress=noExplicitConstructor
             --suppress=unknownMacro
+            --suppress=unsignedPositive
```
mfrankovi, line 17: why did you need to suppress this one?

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
```diff
@@ -221,6 +222,25 @@ DagBlockManager::InsertAndVerifyBlockReturnType DagBlockManager::verifyBlock(con
     markBlockInvalid(block_hash);
     return InsertAndVerifyBlockReturnType::NotEligible;
   }
+  {
```
mfrankovi, line 225: any special reason why this new part of code is under it's own block {...}?

libraries/core_libs/consensus/src/pbft/block_proposer.cpp
```diff
@@ -198,13 +198,26 @@ level_t BlockProposer::getProposeLevel(blk_hash_t const& pivot, vec_blk_t const&
...
 
   // When we propose block we know it is valid, no need for block verification with queue,
   // simply add the block to the DAG
-  DagBlock blk(frontier.pivot, level, std::move(frontier.tips), std::move(trx_hashes), std::move(vdf), node_sk_);
+  vec_trx_t trx_hashes;
+  estimations_vec_t estimations;
+  u256 block_weight = 0;
+  for (auto trx : trxs) {
+    const auto& trx_hash = trx->getHash();
+    auto weight = trx_mgr_->estimateTransactionByHash(trx_hash, proposal_period);
```
mfrankovi, line 210: maybe do an estimate directly on transaction by calling estimateTransaction and not on trx_hash which makes you call getTransaction inside of estimateTransactionByHash to get the transaction

libraries/core_libs/consensus/src/pbft/block_proposer.cpp
```diff
@@ -198,13 +198,26 @@ level_t BlockProposer::getProposeLevel(blk_hash_t const& pivot, vec_blk_t const&
...
-                 [](std::shared_ptr<Transaction> const& t) { return t->getHash(); });
+  auto proposal_period = db_->getProposalPeriodForDagLevel(level);
 
   // When we propose block we know it is valid, no need for block verification with queue,
   // simply add the block to the DAG
-  DagBlock blk(frontier.pivot, level, std::move(frontier.tips), std::move(trx_hashes), std::move(vdf), node_sk_);
+  vec_trx_t trx_hashes;
+  estimations_vec_t estimations;
+  u256 block_weight = 0;
+  for (auto trx : trxs) {
```
mfrankovi, line 208: for (const auto &trx : trxs)

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1164,6 +1165,20 @@ blk_hash_t PbftManager::calculateOrderHash(std::vector<DagBlock> const &dag_bloc
   return dev::sha3(order_stream.out());
 }
 
+std::optional<blk_hash_t> findClosestAnchor(const std::vector<blk_hash_t> &ghost, std::vector<blk_hash_t> &dag_order,
+                                            uint32_t included) {
+  for (uint32_t i = included - 1; i >= 0; i--) {
+    auto is_equal = [&](const auto &e) { return dag_order[i].toString() == e.toString(); };
```
mfrankovi, line 1171: I don't think you need to convert hashes to strings to compare them

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1494,14 +1547,21 @@ bool PbftManager::pushCertVotedPbftBlockIntoChain_(taraxa::blk_hash_t const &cer
...
+  auto [dag_blocks_order, ok] = comparePbftBlockScheduleWithDAGblocks_(pbft_block);
+  if (!ok) {
     LOG(log_nf_) << "DAG has not build up for PBFT block " << cert_voted_block_hash;
     return false;
   }
 
   cert_sync_block_.cert_votes = std::move(cert_votes_for_round);
-  if (!pushPbftBlock_(std::move(cert_sync_block_), std::move(dag_blocks_order.first))) {
+
+  if (!checkBlockWeight(cert_sync_block_)) {
```
mfrankovi, line 1559: is this check really needed? I mean in theory a block that is cert voted on should be guaranteed to have the correct weight.

mfrankovi (approved):

CMakeModules/cppcheck.cmake
kstdl: That was added because of one if in for. I changed it little bit and removed it

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
kstdl: No. I just wanted to group this code. Maybe it is better to move it to separate method, but we will need to have additional code to process it. Like try catch above, or by returning and processing some optional value for example

libraries/core_libs/consensus/include/dag/dag_block_manager.hpp
```diff
@@ -110,6 +115,8 @@ class DagBlockManager {
           // always current anchor level minus dag_expiry_limit_ of non empty pbft periods
 
   SortitionParamsManager sortition_params_manager_;
+  DagConfig dag_config_;
```
MatusKysel, line 118: const?

libraries/core_libs/consensus/src/pbft/block_proposer.cpp
```diff
@@ -198,13 +198,25 @@ level_t BlockProposer::getProposeLevel(blk_hash_t const& pivot, vec_blk_t const&
...
+  for (const auto& trx : trxs) {
+    auto weight = trx_mgr_->estimateTransaction(trx, proposal_period);
+    block_weight += weight;
+    if (block_weight > dag_blk_mgr_->getDagConfig().gas_limit) {
+      break;
+    }
+    trx_hashes.push_back(trx->getHash());
+    estimations.push_back(weight);
+  }
+  DagBlock blk(frontier.pivot, std::move(level), std::move(frontier.tips), std::move(trx_hashes), estimations,
```
MatusKysel, line 217: why not move on `estimations` ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1238,26 +1252,65 @@ blk_hash_t PbftManager::proposePbftBlock_() {
...
+
   if (dag_block_order.empty()) {
     LOG(log_er_) << "DAG anchor block hash " << dag_block_hash << " getDagBlockOrder failed in propose";
     assert(false);
   }
-  std::vector<trx_hash_t> non_finalized_transactions;
+
   std::unordered_set<trx_hash_t> trx_hashes_set;
   std::vector<trx_hash_t> trx_hashes;
-  for (auto const &blk_hash : dag_block_order) {
```
MatusKysel, line 1248: why did you change it to itr? I do not see where are you using iterator

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1238,26 +1252,65 @@ blk_hash_t PbftManager::proposePbftBlock_() {
...
+    if (total_weight + dag_block_weight > config_.gas_limit) {
+      break;
+    }
+    total_weight += dag_block_weight;
+    dag_blocks_included++;
+  }
+
+  if (dag_blocks_included != dag_block_order.size()) {
+    auto closest_anchor = findClosestAnchor(ghost, dag_block_order, dag_blocks_included);
+    assert(closest_anchor.has_value());
```
MatusKysel, line 1295: so what will happened in release? maybe we should log it and do something

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1238,26 +1252,65 @@ blk_hash_t PbftManager::proposePbftBlock_() {
...
+    u256 dag_block_weight = 0;
+    std::vector<trx_hash_t> hashes;
+    const auto &estimations = dag_blk->getEstimations();
+    hashes.reserve(estimations.size());
+
+    int32_t i = 0;
     for (auto const &trx_hash : dag_blk->getTrxs()) {
       if (trx_hashes_set.emplace(trx_hash).second) {
-        trx_hashes.emplace_back(trx_hash);
+        hashes.emplace_back(trx_hash);
```
MatusKysel, line 1281: when are you using this `hashes` ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1238,26 +1252,65 @@ blk_hash_t PbftManager::proposePbftBlock_() {
...
+    for (auto const &dag_blk_hash : dag_block_order) {
+      auto dag_block = dag_blk_mgr_->getDagBlock(dag_blk_hash);
+      assert(dag_block);
+      for (auto const &trx_hash : dag_block->getTrxs()) {
+        if (trx_set.insert(trx_hash).second) {
+          trx_hashes.emplace_back(trx_hash);
+        }
+      }
+    }
+  }
```
MatusKysel, line 1312: can this section be done in upper loop where you calculate size? I mean in case `dag_blocks_included == dag_block_order.size()` then it should be same

libraries/core_libs/node/include/node/node.hpp
```diff
@@ -114,6 +114,7 @@ class FullNode : public std::enable_shared_from_this<FullNode> {
   auto const &getNextVotesManager() const { return next_votes_mgr_; }
   auto const &getPbftChain() const { return pbft_chain_; }
   auto const &getFinalChain() const { return final_chain_; }
+  auto &getBlockProposer() { return blk_proposer_; }
```
MatusKysel, line 117: if we are not using it outside of tests, maybe you should add here a comment

libraries/types/dag_block/src/dag_block.cpp
```diff
@@ -14,19 +14,31 @@ namespace taraxa {
...
+DagBlock::DagBlock(blk_hash_t pivot, level_t level, vec_blk_t tips, vec_trx_t trxs, estimations_vec_t est, sig_t sig,
+                   blk_hash_t hash, addr_t sender)
+    : pivot_(pivot),
+      level_(level),
+      tips_(tips),
+      trxs_(trxs),
+      trx_estimations_(est),
+      sig_(sig),
+      hash_(hash),
+      cached_sender_(sender) {}
```
MatusKysel, line 26: move where it is possible?

libraries/types/dag_block/src/dag_block.cpp
```diff
@@ -144,6 +162,7 @@ void DagBlock::streamRLP(dev::RLPStream &s, bool include_sig) const {
   if (include_sig) {
     s << sig_;
   }
+  s.appendVector(trx_estimations_);
```
MatusKysel, line 165: are we sure about that? I mean this should be calculated per each node and we should not append it to RLP. We malicious player could lie about the gas estimations

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1867,4 +1926,28 @@ void PbftManager::syncBlockQueuePush(SyncBlock &&block, dev::p2p::NodeID const &
...
+  u256 total_weight = 0;
+  for (const auto &tx : blk.transactions) {
+    total_weight += trx_estimations.at(tx.getHash());
+  }
+  if (total_weight > config_.gas_limit) {
+    return false;
+  }
+  return true;
+}
+
```
MatusKysel, line 1952: why do not we keep estimation as part of trx object? I do not see any use-case keeping it part of Dag

MatusKysel (approved):

libraries/config/src/chain_config.cpp
```diff
@@ -89,7 +91,7 @@ decltype(ChainConfig::predefined_) const ChainConfig::predefined_([] {
     cfg.pbft.lambda_ms_min = 2000;
     cfg.pbft.committee_size = 5;
     cfg.pbft.dag_blocks_size = 100;
-    cfg.pbft.ghost_path_move_back = 1;
+    cfg.pbft.ghost_path_move_back = 0;
```
JakubFornadel, line 94: what does this param affect ?

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -67,6 +67,10 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
     }
   }
 
+  if (!pbft_mgr_->checkBlockWeight(sync_block)) {
+    throw MaliciousPeerException("Block summary gas amount is bigger then limit");
+  }
+
```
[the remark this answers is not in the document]
kstdl: Estimations wouldn't change. Because we are estimating it with proposal period from dag block, not on the latest block state(see [block_proposer.cpp:209](https://github.com/Taraxa-project/taraxa-node/pull/1672/files#diff-0240e8dd854ad98bf711fcd683da658fe26250cc6bf29d2c54b488b8a96dba60R209))

libraries/types/dag_block/src/dag_block.cpp
kstdl: If malicious peer would lie about the estimation it wouldn't pass verification on other nodes. Estimations should be the same for all nodes, because we are estimating transactions with dag block `proposal_period`, but not the latest block state(see [block_proposer.cpp:209](https://github.com/Taraxa-project/taraxa-node/pull/1672/files#diff-0240e8dd854ad98bf711fcd683da658fe26250cc6bf29d2c54b488b8a96dba60R209))

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
kstdl: By including it in dag block we are picking proposal period and state for which it would be estimated. Because from one block to other estimation could be different

libraries/config/src/chain_config.cpp
kstdl: It affects count of dag blocks that would be included in pbft block. I will return this back and change only for new tests if it will be needed

libraries/types/dag_block/src/dag_block.cpp
MatusKysel: I see, so another question does it need to be u256? because this has huge impact on DB and Dag size

MatusKysel: if it needs to be u256 do we need to store it as part of `SyncBlock`?

kstdl: Yeah, makes sense. I think we can make it `uint64_t` because I believe out block limit shouldn't be bigger then `uint64_t::max`

gaoqi8649 (commented):
There set PBFT "gas_limit": "0x3938700" that is 60000000, and DAG block "gas_limit": "0x989680" that is 10000000. 
For example, each DAG block is full (gas fee to 10000000), one PBFT block max include 6 DAG blocks.
If in DAG level 10, there are 12 DAG blocks {b1, ..., b12}, b1 is the anchor in level 10. And in level 11, b13 is the anchor. 
How to calculate gas limitation for proposal PBFT block with anchor b13?

libraries/common/include/common/util.hpp
```diff
@@ -50,11 +50,14 @@ auto int_pow(Int1 x, Int2 y) {
 template <typename T>
 std::vector<T> asVector(Json::Value const &json) {
   std::vector<T> v;
+  v.reserve(json.size());
   std::transform(json.begin(), json.end(), std::back_inserter(v),
                  [](const Json::Value &item) { return T(item.asString()); });
   return v;
 }
 
+std::vector<uint64_t> asUInt64Vector(Json::Value const &json);
```
gaoqi8649, line 59: const Json::Value& json

libraries/common/src/util.cpp
```diff
@@ -34,4 +34,11 @@ std::string getFormattedVersion(std::initializer_list<uint32_t> list) {
   return ret.substr(0, ret.size() - 1);
 }
 
+std::vector<uint64_t> asUInt64Vector(Json::Value const &json) {
```
gaoqi8649, line 37: const Json::Value& json

libraries/config/include/config/dag_config.hpp
```diff
@@ -0,0 +1,15 @@
...
+#include <json/json.h>
+
+#include "common/types.hpp"
+
+namespace taraxa {
+
+struct DagConfig {
+  uint64_t gas_limit = 0x989680;
+};
+Json::Value enc_json(DagConfig const& obj);
```
gaoqi8649, line 12: const DagConfig& obj

libraries/config/include/config/dag_config.hpp
```diff
@@ -0,0 +1,15 @@
...
+void dec_json(Json::Value const& json, DagConfig& obj);
```
gaoqi8649, line 13: const Json::Value& json

libraries/core_libs/consensus/include/dag/dag_block_manager.hpp
```diff
@@ -25,12 +25,15 @@ class DagBlockManager {
...
-    ExpiredBlock
+    ExpiredBlock,
+    IncorrectTransactionsEstimation,
+    BlockTooBig
   };
 
-  DagBlockManager(addr_t node_addr, SortitionConfig const &sortition_config, std::shared_ptr<DbStorage> db,
-                  std::shared_ptr<TransactionManager> trx_mgr, std::shared_ptr<FinalChain> final_chain,
-                  std::shared_ptr<PbftChain> pbft_chain, logger::Logger log_time_, uint32_t queue_limit = 0,
+  DagBlockManager(addr_t node_addr, SortitionConfig const &sortition_config, const DagConfig &dag_config,
```
gaoqi8649, line 33: const SortitionConfig& sortition_config

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
```diff
@@ -4,7 +4,7 @@
 
 namespace taraxa {
 
-DagBlockManager::DagBlockManager(addr_t node_addr, SortitionConfig const &sortition_config,
+DagBlockManager::DagBlockManager(addr_t node_addr, SortitionConfig const &sortition_config, const DagConfig &dag_config,
```
gaoqi8649, line 7: const SortitionConfig& sortition_config

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1164,6 +1165,19 @@ blk_hash_t PbftManager::calculateOrderHash(std::vector<DagBlock> const &dag_bloc
   return dev::sha3(order_stream.out());
 }
 
+std::optional<blk_hash_t> findClosestAnchor(const std::vector<blk_hash_t> &ghost, std::vector<blk_hash_t> &dag_order,
+                                            uint32_t included) {
+  for (uint32_t i = included; i > 0; i--) {
+    if (std::count(ghost.begin(), ghost.end(), dag_order[i - 1])) {
```
gaoqi8649, line 1171: I don't understand here. There must have dag_order[i - 1] and only have one. So here should always return 1

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1238,26 +1252,68 @@ blk_hash_t PbftManager::proposePbftBlock_() {
...
+    auto closest_anchor = findClosestAnchor(ghost, dag_block_order, dag_blocks_included);
+    if (!closest_anchor) {
+      LOG(log_er_) << "Can't find closest anchor after block clipping. Ghost: " << ghost << ". Clipped block_order: "
+                   << vec_blk_t(dag_block_order.begin(), dag_block_order.begin() + dag_blocks_included);
+      assert(false);
+    }
+
+    dag_block_hash = closest_anchor.value();
+    dag_block_order = dag_mgr_->getDagBlockOrder(dag_block_hash, propose_period);
+  }
```
gaoqi8649, line 1302: I don't think set PBFT gas limit will work. 

There set PBFT "gas_limit": "0x3938700" that is 60000000, and DAG block "gas_limit": "0x989680" that is 10000000.
For example, each DAG block is full (gas fee to 10000000), one PBFT block max include 6 DAG blocks.
If in DAG level 10, there are 8 DAG blocks {b1, ..., b8}, b1 is the anchor in level 10 and have been finalized. And in level 11, b9 is the anchor.
DAG:
...     Level 10       Level 11, ...
        b1                 b9
        b2
        b3 
        b4
        b5
        b6
        b7 
        b8
ghost will be {b9}
dag_block_order will be {b2, b3, b4, b5, b6, b7, b8, b9}
dag_blocks_included will be 6
findClosestAnchor will return {}
Then assertion happen.

So the approach cannot handle the case in each level DAG blocks that have more than PBFT gas limitation

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1476,6 +1531,11 @@ std::pair<vec_blk_t, bool> PbftManager::comparePbftBlockScheduleWithDAGblocks_(s
     return std::make_pair(std::move(dag_blocks_order), false);
   }
 
+  if (!checkBlockWeight(cert_sync_block_)) {
+    LOG(log_er_) << "PBFT block " << pbft_block->getBlockHash() << " is overweighted";
+    return std::make_pair(std::move(dag_blocks_order), false);
```
gaoqi8649, line 1536: return std::make_pair({}, false);

gaoqi8649:
Forget check checkBlockWeight for synced PBFT blocks from peers. 
I think should check in PbftManager::processSyncBlock()

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -21,6 +21,28 @@ TransactionManager::TransactionManager(FullNodeConfig const &conf, std::shared_p
...
+      state_api::EVMTransaction{
+          trx->getSender(),
+          trx->getGasPrice(),
+          trx->getReceiver(),
+          trx->getNonce(),
+          trx->getValue(),
+          trx->getGas(),
+          trx->getData(),
+      },
+      proposal_period);
```
gaoqi8649, line 42: Since PBFT chain is async with execution of final chain. When final chain execution much behind to PBFT chain. That should not work here.

libraries/config/include/config/pbft_config.hpp
```diff
@@ -13,6 +13,7 @@ struct PbftConfig {
   uint32_t dag_blocks_size = 0;
   uint32_t ghost_path_move_back = 0;
   bool run_count_votes = false;
+  uint64_t gas_limit = 0x3938700;
```
gaoqi8649, line 16: Should initial with 0, and will read from config

libraries/core_libs/consensus/src/pbft/block_proposer.cpp
```diff
@@ -198,13 +198,25 @@ level_t BlockProposer::getProposeLevel(blk_hash_t const& pivot, vec_blk_t const&
 void BlockProposer::proposeBlock(DagFrontier&& frontier, level_t level, SharedTransactions&& trxs, VdfSortition&& vdf) {
   if (stopped_) return;
 
-  vec_trx_t trx_hashes;
-  std::transform(trxs.begin(), trxs.end(), std::back_inserter(trx_hashes),
-                 [](std::shared_ptr<Transaction> const& t) { return t->getHash(); });
+  auto proposal_period = db_->getProposalPeriodForDagLevel(level);
```
gaoqi8649, line 201: proposal_period should pass from propose() instead of reading from DB again

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1238,26 +1252,68 @@ blk_hash_t PbftManager::proposePbftBlock_() {
...
+  if (dag_blocks_included != dag_block_order.size()) {
+    auto closest_anchor = findClosestAnchor(ghost, dag_block_order, dag_blocks_included);
+    if (!closest_anchor) {
+      LOG(log_er_) << "Can't find closest anchor after block clipping. Ghost: " << ghost << ". Clipped block_order: "
+                   << vec_blk_t(dag_block_order.begin(), dag_block_order.begin() + dag_blocks_included);
+      assert(false);
+    }
+
+    dag_block_hash = closest_anchor.value();
+    dag_block_order = dag_mgr_->getDagBlockOrder(dag_block_hash, propose_period);
```
gaoqi8649, line 1301: I think don't need call getDagBlockOrder again. In findClosestAnchor, you have erase extra DAG blocks already. 
I am not sure if topological sorting will return same ordering here. Please double check

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
gaoqi8649: In our experience, the proposal_period should always behind the finalized period. I think should be fine here.
Let's add "try" and "catch" for exception for final_chain_->call

libraries/aleth/libdevcore/RLP.h
```diff
@@ -565,6 +565,12 @@ class RLPStream {
     for (auto const& i : _s) append(i);
     return *this;
   }
+  template <class _T, class _T2>
+  RLPStream& appendMap(std::map<_T, _T2> const& _s) {
```
gaoqi8649, line 569: Do you need this? I don't find where using appendMap

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
kstdl: But we shouldn't have so much dag blocks in one level I believe. We can change the ratio between this limits to potentially include more dag blocks into pbft if you think thats not enough

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
kstdl: I didn't get what are you talking about in 1 message
Isn't it right that `proposal_period` should be always behind finalized period?

libraries/aleth/libdevcore/RLP.h
```diff
@@ -565,6 +565,12 @@ class RLPStream {
...
+    appendList(_s.size());
+    for (auto const& p : _s) append(p);
```
JakubFornadel, line 571: const auto

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
kstdl: We need to. I also thought that we can just clip it, but not. It actually could be different even in test cases I wrote

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
JakubFornadel: you can separate it by some comment and visually it would have the same effect:

```
// This part of code does this and that
u256 total_block_weight = 0;
const auto &trxs = blk.getTrxs();
const auto &estimations = blk.getEstimations();
if (trxs.size() != estimations.size()) {
  return InsertAndVerifyBlockReturnType::IncorrectTransactionsEstimation;
}
for (uint32_t i = 0; i < trxs.size(); ++i) {
  const auto &e = trx_mgr_->estimateTransactionByHash(trxs[i], propose_period);
  if (e != estimations[i]) {
    return InsertAndVerifyBlockReturnType::IncorrectTransactionsEstimation;
  }
  total_block_weight += estimations[i];
}
if (total_block_weight > getDagConfig().gas_limit) {
  return InsertAndVerifyBlockReturnType::BlockTooBig;
}

// And this does another thing....
```

libraries/types/dag_block/include/dag/dag_block.hpp
```diff
@@ -7,6 +7,7 @@ namespace taraxa {
 
 using std::string;
 using VdfSortition = vdf_sortition::VdfSortition;
+using estimations_vec_t = std::vector<uint64_t>;
```
JakubFornadel, line 10: what is the advantage of using this alias ? In terms of length it the same as `std::vector<uint64_t>`

libraries/types/dag_block/include/dag/dag_block.hpp
```diff
@@ -20,7 +21,9 @@ class DagBlock {
   level_t level_ = 0;
   vec_blk_t tips_;
   vec_trx_t trxs_;  // transactions
+  estimations_vec_t trx_estimations_;
```
JakubFornadel, line 24: I dont think I would automatically know that trx_estimation is trxs_gas_estimations if I was not reviewing this PR

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
```diff
@@ -221,6 +222,25 @@ DagBlockManager::InsertAndVerifyBlockReturnType DagBlockManager::verifyBlock(con
...
     return InsertAndVerifyBlockReturnType::NotEligible;
   }
+  {
+    u256 total_block_weight = 0;
+    const auto &trxs = blk.getTrxs();
+    const auto &estimations = blk.getEstimations();
+    if (trxs.size() != estimations.size()) {
+      return InsertAndVerifyBlockReturnType::IncorrectTransactionsEstimation;
+    }
+    for (uint32_t i = 0; i < trxs.size(); ++i) {
```
JakubFornadel, line 232: size_t

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
```diff
@@ -221,6 +222,25 @@ DagBlockManager::InsertAndVerifyBlockReturnType DagBlockManager::verifyBlock(con
...
+      const auto &e = trx_mgr_->estimateTransactionByHash(trxs[i], propose_period);
+      if (e != estimations[i]) {
+        return InsertAndVerifyBlockReturnType::IncorrectTransactionsEstimation;
+      }
+      total_block_weight += estimations[i];
```
JakubFornadel, line 237: total_block_weight += e;

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1238,26 +1252,68 @@ blk_hash_t PbftManager::proposePbftBlock_() {
...
   for (auto const &blk_hash : dag_block_order) {
     auto dag_blk = dag_blk_mgr_->getDagBlock(blk_hash);
     if (!dag_blk) {
       LOG(log_er_) << "DAG anchor block hash " << dag_block_hash << " getDagBlock failed in propose for block "
                    << blk_hash;
       assert(false);
     }
-    for (auto const &trx_hash : dag_blk->getTrxs()) {
+    u256 dag_block_weight = 0;
+    const auto &estimations = dag_blk->getEstimations();
```
JakubFornadel, line 1273: again, something like getTxsGasEstimations would make it much clearer of what you are actually getting here

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
JakubFornadel: so why should be proposal_period always behind finalized_period ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1164,6 +1165,19 @@ blk_hash_t PbftManager::calculateOrderHash(std::vector<DagBlock> const &dag_bloc
...
+                                            uint32_t included) {
+  for (uint32_t i = included; i > 0; i--) {
+    if (std::count(ghost.begin(), ghost.end(), dag_order[i - 1])) {
+      if (dag_order.size() > i) {
+        dag_order.erase(dag_order.begin() + i, dag_order.end());
+      }
+      return dag_order[i - 1];
+    }
+  }
+  return {};
```
gaoqi8649, line 1178: I think we should return ghost[0] here. And also need modify checkBlockWeight return true if weight < gas limit OR DAG anchor is ghost[0]

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1164,6 +1165,19 @@ blk_hash_t PbftManager::calculateOrderHash(std::vector<DagBlock> const &dag_bloc
   return dev::sha3(order_stream.out());
 }
 
+std::optional<blk_hash_t> findClosestAnchor(const std::vector<blk_hash_t> &ghost, std::vector<blk_hash_t> &dag_order,
+                                            uint32_t included) {
+  for (uint32_t i = included; i > 0; i--) {
+    if (std::count(ghost.begin(), ghost.end(), dag_order[i - 1])) {
+      if (dag_order.size() > i) {
+        dag_order.erase(dag_order.begin() + i, dag_order.end());
+      }
```
gaoqi8649, line 1174: I think we don't need remove dag blocks here, because we need call getDagBlockOrder later

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
gaoqi8649: other failed case has dag_blocks_order.clear() before return

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
gaoqi8649: If you modify this way, I think you cannot verify synced PBFT block when node just receive it because DAG blocks have not exist at the moment. That need to change to verify just before adding to PBFT chain

JakubFornadel (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1164,6 +1165,19 @@ blk_hash_t PbftManager::calculateOrderHash(std::vector<DagBlock> const &dag_bloc
...
+      return dag_order[i - 1];
+    }
+  }
+  return ghost[1];
```
gaoqi8649, line 1178: Are you sure this's ghost[1] not ghost[0]? I am not sure if include last finalized DAG anchor in ghost. Please double check
But in case if only has DAG genesis block, that will crash here

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
```diff
@@ -221,10 +222,30 @@ DagBlockManager::InsertAndVerifyBlockReturnType DagBlockManager::verifyBlock(con
...
     return InsertAndVerifyBlockReturnType::NotEligible;
   }
+  {
+    u256 total_block_weight = 0;
+    const auto &trxs = blk.getTrxs();
+    const auto &trxs_gas_estimations = blk.getTrxsGasEstimations();
+    if (trxs.size() != trxs_gas_estimations.size()) {
+      return InsertAndVerifyBlockReturnType::IncorrectTransactionsEstimation;
+    }
+    final_chain_->wait_for_finalization(*propose_period, stopped_);
```
gaoqi8649, line 232: This will cause stuck here until execute to proposed period.
I think that's better for mark the block for FutureBlock, and process it again later. Please see above line 212 - 217

kstdl: Oh. So this waits doesn't make sense because we are already implicitly checking this in `dpos_is_eligible` method. I will just remove it and add comment

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
kstdl: Yeah. It seems that `ghost[0]` is last finalized DAG anchor. I tried to return `ghost[0]` here and it broke DAG production.
If it will have only genesis block, it won't reach this code because of [pbft_manager.cpp:1237](https://github.com/Taraxa-project/taraxa-node/blob/6059ab3e099d3544ce8f233751963339d772b541/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp#L1237)

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
kstdl: https://github.com/Taraxa-project/taraxa-node/pull/1672#discussion_r867748891

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -67,6 +67,10 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
     }
   }
 
+  if (!pbft_mgr_->checkBlockWeight(sync_block)) {
+    throw MaliciousPeerException("Block summary gas amount is bigger then limit");
+  }
```
gaoqi8649, line 72: The check has issue when there are more DAG blocks in single level.
I think we cannot check here anymore. Because when node just receive synced PBFT block, node has not have DAG yet. That need check just before adding the block to PBFT chain

gaoqi8649 (approved):

PR #1674, opened 2022-04-05
chore: dag expiry and light node tests improvements

mfrankovi:
There were several issues related to using period limit for dag blocks expiry. New approach is using a dag block level limit which is much cleaner and easier to check without a need to pull proposal period from the database for each dag block. Level limit is based on the level of the last anchor from the last pbft block in the chain.

Dag expiry limit and max dag levels per period can now be configured to be able to test them without a need to have a large dag and large pbft chain.  This allowed for simpler light node tests and creation of additional test in dag_test which tests dag expiry.

### DIFF
--- a/libraries/config/include/config/config.hpp
+++ b/libraries/config/include/config/config.hpp
@@ -98,8 +98,10 @@ struct FullNodeConfig {
   ChainConfig chain = ChainConfig::predefined();
   state_api::Opts opts_final_chain;
   std::vector<logger::Config> log_configs;
-  bool is_light_node = false;       // Is light node
-  uint64_t light_node_history = 0;  // Number of periods to keep in history for a light node
+  bool is_light_node = false;                            // Is light node
+  uint64_t light_node_history = 0;                       // Number of periods to keep in history for a light node
+  uint32_t dag_expiry_limit = kDagExpiryLevelLimit;      // For unit tests only
+  uint32_t max_levels_per_period = kMaxLevelsPerPeriod;  // For unit tests only
 
   auto net_file_path() const { return data_path / "net"; }
 
--- a/libraries/core_libs/consensus/include/dag/dag.hpp
+++ b/libraries/core_libs/consensus/include/dag/dag.hpp
@@ -132,7 +132,9 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
 
   explicit DagManager(blk_hash_t const &genesis, addr_t node_addr, std::shared_ptr<TransactionManager> trx_mgr,
                       std::shared_ptr<PbftChain> pbft_chain, std::shared_ptr<DagBlockManager> dag_blk_mgr,
-                      std::shared_ptr<DbStorage> db, logger::Logger log_time);
+                      std::shared_ptr<DbStorage> db, logger::Logger log_time, bool is_light_node = false,
+                      uint64_t light_node_history = 0, uint32_t max_levels_per_period = kMaxLevelsPerPeriod,
+                      uint32_t dag_expiry_limit = kDagExpiryLevelLimit);
   virtual ~DagManager() { stop(); }
 
   DagManager(const DagManager &) = delete;
@@ -148,7 +150,7 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
   blk_hash_t const &get_genesis() { return genesis_; }
 
   bool pivotAndTipsAvailable(DagBlock const &blk);
-  void addDagBlock(DagBlock &&blk, SharedTransactions &&trxs = {}, bool proposed = false,
+  bool addDagBlock(DagBlock &&blk, SharedTransactions &&trxs = {}, bool proposed = false,
                    bool save = true);  // insert to buffer if fail
 
   // return block order
@@ -166,6 +168,9 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
    * @return number of dag blocks finalized
    */
   uint setDagBlockOrder(blk_hash_t const &anchor, uint64_t period, vec_blk_t const &dag_order);
+
+  uint64_t getLightNodeHistory() const { return light_node_history_; }
+  bool isLightNode() const { return is_light_node_; }
 
   std::optional<std::pair<blk_hash_t, std::vector<blk_hash_t>>> getLatestPivotAndTips() const;
 
@@ -192,6 +197,13 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
     SharedLock lock(mutex_);
     return std::make_pair(old_anchor_, anchor_);
   }
+
+  /**
+   * @brief Retrieves Dag expiry limit
+   *
+   * @return limit
+   */
+  uint32_t getDagExpiryLimit() const { return dag_expiry_limit_; }
 
   const std::pair<uint64_t, std::map<uint64_t, std::unordered_set<blk_hash_t>>> getNonFinalizedBlocks() const;
 
@@ -228,6 +240,7 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
   bool validateBlockNotExpired(const std::shared_ptr<DagBlock> &dag_block,
                                std::unordered_map<blk_hash_t, std::shared_ptr<DagBlock>> &expired_dag_blocks_to_remove);
   void handleExpiredDagBlocksTransactions(const std::vector<trx_hash_t> &transactions_from_expired_dag_blocks) const;
+  void clearLightNodeHistory();
 
   void worker();
   std::pair<blk_hash_t, std::vector<blk_hash_t>> getFrontier() const;  // return pivot and tips
@@ -250,6 +263,13 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
   DagFrontier frontier_;
   std::atomic<bool> stopped_ = true;
   std::thread block_worker_;
+
+  const bool is_light_node_ = false;
+  const uint64_t light_node_history_ = 0;
+  const uint32_t max_levels_per_period_;
+  const uint32_t dag_expiry_limit_;  // Any non finalized dag block with a level smaller by
+                                     // dag_expiry_limit_ than the current period anchor level is considered
+                                     // expired and it should be ignored or removed from DAG
 
   logger::Logger log_time_;
   LOG_OBJECTS_DEFINE
--- a/libraries/core_libs/consensus/include/dag/dag_block_manager.hpp
+++ b/libraries/core_libs/consensus/include/dag/dag_block_manager.hpp
@@ -30,7 +30,8 @@ class DagBlockManager {
 
   DagBlockManager(addr_t node_addr, SortitionConfig const &sortition_config, std::shared_ptr<DbStorage> db,
                   std::shared_ptr<TransactionManager> trx_mgr, std::shared_ptr<FinalChain> final_chain,
-                  std::shared_ptr<PbftChain> pbft_chain, logger::Logger log_time_, uint32_t queue_limit = 0);
+                  std::shared_ptr<PbftChain> pbft_chain, logger::Logger log_time_, uint32_t queue_limit = 0,
+                  uint32_t max_levels_per_period = kMaxLevelsPerPeriod);
   ~DagBlockManager();
 
   DagBlockManager(const DagBlockManager &) = delete;
@@ -59,6 +60,20 @@ class DagBlockManager {
    * @return true in case block was actually marked as seen(was not seen before), otherwise false (was already seen)
    */
   bool markDagBlockAsSeen(const DagBlock &dag_block);
+
+  /**
+   * @brief Sets dag expiry level
+   *
+   * @param dag_expiry_level
+   */
+  void setDagExpiryLevel(uint64_t dag_expiry_level) { dag_expiry_level_ = dag_expiry_level; }
+
+  /**
+   * @brief Retrieves Dag expiry level
+   *
+   * @return level
+   */
+  uint64_t getDagExpiryLevel() { return dag_expiry_level_; }
 
   std::shared_ptr<DagBlock> getDagBlock(blk_hash_t const &hash) const;
   bool pivotAndTipsValid(DagBlock const &blk);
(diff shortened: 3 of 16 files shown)


libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
```diff
@@ -181,17 +181,16 @@ DagBlockManager::InsertAndVerifyBlockReturnType DagBlockManager::verifyBlock(con
   }
 
   auto propose_period = db_->getProposalPeriodForDagLevel(blk.getLevel());
-  const auto expiry_period = pbft_chain_->getDagExpiryPeriod();
   // Verify DPOS
   if (!propose_period) {
     // Cannot find the proposal period in DB yet. The slow node gets an ahead block, remove from seen_blocks
     LOG(log_nf_) << "Cannot find proposal period " << *propose_period << " in DB for DAG block " << blk.getHash();
```
MatusKysel, line 187: can you please fix this log? if `propose_period`does not exist  `*propose_period` is UB

libraries/core_libs/consensus/include/dag/dag.hpp
```diff
@@ -251,6 +259,12 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
   std::atomic<bool> stopped_ = true;
   std::thread block_worker_;
 
+  const uint32_t dag_expiry_limit_;  // Any non finalized dag block with a level smaller by
+                                     // dag_expiry_limit_ than the current period anchor level is considered
+                                     // expired and it should be ignored or removed from DAG
+  uint64_t dag_expiry_level_ = 0;    // Level below which dag blocks are considered expired, it's value is
```
MatusKysel, line 265: does not need to be thread safe?

libraries/core_libs/consensus/include/dag/dag.hpp
```diff
@@ -221,6 +222,13 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
    */
   std::shared_mutex &getDagMutex() { return mutex_; }
 
+  /**
+   * @brief Retrieves Dag expiry level
+   *
+   * @return level
+   */
+  uint64_t getDagExpiryLevel() { return dag_expiry_level_; }
```
MatusKysel, line 230: why this is not part of dag block mnger, but we have this value in 2 classes?

libraries/core_libs/consensus/include/dag/dag_block_manager.hpp
```diff
@@ -60,6 +61,13 @@ class DagBlockManager {
    */
   bool markDagBlockAsSeen(const DagBlock &dag_block);
 
+  /**
+   * @brief Sets dag expiry level
+   *
+   * @param dag_expiry_level
+   */
+  void setDagExpiryLevel(uint64_t &dag_expiry_level) { dag_expiry_level_ = dag_expiry_level; }
```
MatusKysel, line 69: no `&`

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
```diff
@@ -256,6 +257,8 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
 
   SyncBlockQueue sync_queue_;
 
+  uint32_t max_levels_per_period_;
```
MatusKysel, line 260: const

MatusKysel (approved):

libraries/core_libs/consensus/include/dag/dag_block_manager.hpp
```diff
@@ -90,6 +98,10 @@ class DagBlockManager {
   std::condition_variable_any cond_for_verified_qu_;
   std::map<uint64_t, std::deque<DagBlock>> verified_qu_;
 
+  std::atomic_uint64_t dag_expiry_level_ =
```
gaoqi8649, line 101: There is another dag_expiry_level_ in dag.hpp. Does this duplicate?

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -610,6 +614,7 @@ uint DagManager::setDagBlockOrder(blk_hash_t const &new_anchor, uint64_t period,
         if (validateBlockNotExpired(dag_block, expired_dag_blocks_to_remove)) {
           addToDag(blk_hash, pivot_hash, dag_block->getTips(), dag_block->getLevel(), false);
         } else {
+          db_->removeDagBlock(blk_hash);
```
gaoqi8649, line 617: Dose this also need to remove from seen_blocks_?

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -68,7 +68,8 @@ void StatusPacketHandler::process(const PacketData& packet_data, const std::shar
       if (pbft_synced_period + node_history < peer_pbft_chain_size) {
         LOG(log_nf_) << "Light node is not able to serve our syncing request. " << packet_data.from_node_id_.abridged()
                      << " peer will be disconnected";
-        disconnect(peer->getId(), dev::p2p::UserReason);
+        disconnect(packet_data.from_node_id_, dev::p2p::UserReason);
```
gaoqi8649, line 71: What is the difference between "peer->getId()" and "packet_data.from_node_id_"? Are they same thing?

libraries/config/include/config/config.hpp
```diff
@@ -100,6 +100,8 @@ struct FullNodeConfig {
   std::vector<logger::Config> log_configs;
   bool is_light_node = false;       // Is light node
   uint64_t light_node_history = 0;  // Number of periods to keep in history for a light node
+  uint32_t dag_expiry_limit = kDagExpiryLevelLimit;
+  uint32_t max_levels_per_period = kMaxLevelsPerPeriod;
```
gaoqi8649, line 104: They don't get values from config file. Why not using value kDagExpiryLevelLimit/kMaxLevelsPerPeriod directly?

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -417,14 +417,18 @@ std::optional<SortitionParamsChange> DbStorage::getParamsChangeForPeriod(uint64_
   return SortitionParamsChange::from_rlp(dev::RLP(it->value().ToString()));
 }
 
-void DbStorage::clearPeriodDataHistory(uint64_t period, uint64_t dag_expiry_period, bool force) {
+void DbStorage::clearPeriodDataHistory(uint64_t period, uint64_t dag_expiry_level, bool force) {
   // Actual history size will be between 100% and 110% of light_node_history_ to avoid deleting on every period
   if (is_light_node_ && ((period % (std::max(light_node_history_ / 10, (uint64_t)1)) == 0) || force) &&
-      period > light_node_history_) {
+      period > light_node_history_ && dag_expiry_level > 0) {
```
gaoqi8649, line 423: dag_expiry_level > 0? I think that will delete DAG blocks that have not expired in db_->DeleteRange(write_options_, handle(Columns::period_data), toSlice(start), toSlice(end));

gaoqi8649: For example, dag_expiry_level = 1, we should only remove DAG blocks in level 1.
But proposal_period will be 0, in proposal_period 0 DAG levels should be [0, 100].
In the code, actually removed DAG levels to 100, I think that's wrong

tests/dag_test.cpp
```diff
@@ -227,6 +227,84 @@ TEST_F(DagTest, compute_epoch) {
...
+
+  orders = mgr->getDagBlockOrder(blk_new_anchor.getHash(), 2);
+  mgr->setDagBlockOrder(blk_new_anchor.getHash(), 2, orders);
+
+  // Verify that the block blk_at_limit which was initially part of the DAG became expired once new anchor moved the
+  // limit
+  EXPECT_FALSE(db_ptr->dagBlockInDb(blk_under_limit.getHash()));
+  EXPECT_FALSE(db_ptr->dagBlockInDb(blk_at_limit.getHash()));
+  EXPECT_TRUE(db_ptr->dagBlockInDb(blk_over_limit.getHash()));
+}
```
gaoqi8649, line 306: Could you please add tests for clearPeriodDataHistory? Not remove non-expired DAG blocks

libraries/config/include/config/config.hpp
gaoqi8649: I think that is for unit tests purpose. Could you please add comments "For unit tests config"?

libraries/core_libs/consensus/src/dag/dag_block_manager.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/include/dag/dag.hpp
mfrankovi: No, because every place where it is used is under DagManager::mutex_

libraries/core_libs/consensus/include/dag/dag.hpp
mfrankovi: So you would only put it in dag block manager and have dag manager get it from there? DagBlockManager and DagManager have mutexes where we already had some deadlocks. This value changes only once per period but it will be accessed very often when we verify dag blocks. I think having it in both places makes it more optimal for quick check for dag expiry. Could you propose how you would implement it?

libraries/core_libs/consensus/include/dag/dag_block_manager.hpp
mfrankovi: Fixed

libraries/core_libs/consensus/include/dag/dag_block_manager.hpp
mfrankovi: See my answer to Matus

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: I don't think so. Seen blocks will make us ignore an expired block if we receive it again.

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
mfrankovi: peer can be null here, if peer is not null it is the same value, but this is safer

libraries/config/include/config/config.hpp
mfrankovi: To be able to run tests with different values. Testing with max level of 100 requires building dag of 100 levels. See the dag_expiry test I added in dag_test

libraries/core_libs/storage/src/storage.cpp
mfrankovi: This line prevents deleting non-expired data:

// This prevents deleting any data needed for dag blocks proposal period, we only delete periods for the expired dag
    // blocks
    const uint64_t end = std::min(period - light_node_history_, *proposal_period);

tests/dag_test.cpp
mfrankovi: I have added test for this FullNodeTest clear_period_data

libraries/core_libs/consensus/include/dag/dag.hpp
mfrankovi: ok, I have put it in DagBlockManager

kstdl (approved):

libraries/config/include/config/config.hpp
gaoqi8649: Could you please add comments?

uint32_t dag_expiry_limit = kDagExpiryLevelLimit;               // For unit tests only
uint32_t max_levels_per_period = kMaxLevelsPerPeriod;   // For unit tests only

libraries/core_libs/storage/src/storage.cpp
mfrankovi: I have made modifications and now I delete based on this period:
const auto proposal_period = db_->getProposalPeriodForDagLevel(dag_expiry_level - max_levels_per_period_ - 1);
So I am getting the proposal period of dag_expiry_level - max_levels_per_period_ - 1 which ensures that this period cannot have a dag block of level dag_expiry_level  or higher

libraries/config/include/config/config.hpp
mfrankovi: Done

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -550,6 +553,26 @@ std::vector<blk_hash_t> DagManager::getDagBlockOrder(blk_hash_t const &anchor, u
   return blk_orders;
 }
 
+void DagManager::clearLightNodeHistory() {
+  const auto dag_expiry_level = dag_blk_mgr_->getDagExpiryLevel();
+  // Actual history size will be between 100% and 110% of light_node_history_ to avoid deleting on every period
+  if (((period_ % (std::max(light_node_history_ / 10, (uint64_t)1)) == 0)) && period_ > light_node_history_ &&
+      dag_expiry_level > max_levels_per_period_ + 1) {
```
gaoqi8649, line 560: I think should be "dag_expiry_level > max_levels_per_period_"

tests/full_node_test.cpp
```diff
@@ -1547,30 +1547,87 @@ TEST_F(FullNodeTest, transaction_validation) {
...
+                   nodes[1]->getPbftChain()->getPbftChainSizeExcludingEmptyPbftBlocks())
   });
-  uint32_t min_period_in_db = 0;
+  uint32_t non_empty_counter = 0;
   for (uint64_t i = 0; i < nodes[0]->getPbftChain()->getPbftChainSize(); i++) {
-    if (nodes[0]->getDB()->getPeriodDataRaw(i).size() > 0) {
-      min_period_in_db = i;
-      break;
+    const auto pbft_block = nodes[0]->getDB()->getPbftBlock(i);
+    if (pbft_block) {
```
gaoqi8649, line 1573: if (pbft_block && pbft_block->getPivotDagBlockHash() != NULL_BLOCK_HASH) {

gaoqi8649 (approved):

tests/full_node_test.cpp
mfrankovi: I want to count null block hashes as well. I just dont want to count the ones that are deleted from the db

PR #1680, opened 2022-04-11
chore: limit number of votes in packet

mfrankovi:
Limit number of votes in a single packet and throttle votes packets not to send too many votes at once. The same is already done for transactions.

### DIFF
--- a/libraries/common/include/common/constants.hpp
+++ b/libraries/common/include/common/constants.hpp
@@ -19,6 +19,11 @@ constexpr uint16_t kOnePercent = 100;
 constexpr uint16_t kMaxLevelsPerPeriod = 100;
 constexpr uint32_t kDagExpiryLevelLimit = 1000;
 
+const uint32_t kMaxVotesInPacket{1000};
+const uint32_t kMaxTransactionsInPacket{1000};
+
+const uint32_t kPeriodicEventsThreadCount{2};
+
 // The various denominations; here for ease of use where needed within code.
 static const u256 kOneTara = dev::exp10<18>();
 // static const u256 kFinney = exp10<15>();
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
@@ -39,18 +39,27 @@ void ExtVotesPacketHandler::sendPbftNextVotes(dev::p2p::NodeID const &peer_id,
     return;
   }
 
-  dev::RLPStream s(send_next_votes_bundle.size());
-  for (auto const &next_vote : send_next_votes_bundle) {
-    s.appendRaw(next_vote->rlp(true));
-    LOG(log_dg_) << "Send out next vote " << next_vote->getHash() << " to peer " << peer_id;
-  }
+  uint32_t index = 0;
+  while (index < send_next_votes_bundle.size()) {
+    uint32_t vote_count_to_send =
+        std::min(static_cast<size_t>(kMaxVotesInPacket), send_next_votes_bundle.size() - index);
 
-  if (sealAndSend(peer_id, SubprotocolPacketType::VotesSyncPacket, std::move(s))) {
-    LOG(log_nf_) << "Send out size of " << send_next_votes_bundle.size() << " PBFT next votes to " << peer_id;
-    if (auto peer = peers_state_->getPeer(peer_id)) {
-      for (auto const &v : send_next_votes_bundle) {
-        peer->markVoteAsKnown(v->getHash());
-      }
+    dev::RLPStream s(vote_count_to_send);
+    for (uint32_t i = index; i < index + vote_count_to_send; i++) {
+      const auto &next_vote = send_next_votes_bundle[i];
+      s.appendRaw(next_vote->rlp(true));
+      LOG(log_dg_) << "Send out next vote " << next_vote->getHash() << " to peer " << peer_id;
+    }
+
+    if (sealAndSend(peer_id, SubprotocolPacketType::VotesSyncPacket, std::move(s))) {
+      LOG(log_nf_) << "Send out size of " << vote_count_to_send << " PBFT next votes to " << peer_id;
+    }
+
+    index += vote_count_to_send;
+  }
+  if (auto peer = peers_state_->getPeer(peer_id)) {
+    for (auto const &v : send_next_votes_bundle) {
+      peer->markVoteAsKnown(v->getHash());
     }
   }
 }
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
@@ -23,8 +23,8 @@ void VotesSyncPacketHandler::validatePacketRlpFormat([[maybe_unused]] const Pack
 
 void VotesSyncPacketHandler::process(const PacketData &packet_data, const std::shared_ptr<TaraxaPeer> &peer) {
   const auto next_votes_count = packet_data.rlp_.itemCount();
-  if (next_votes_count == 0) {
-    LOG(log_er_) << "Receive 0 next votes from peer " << packet_data.from_node_id_
+  if (next_votes_count == 0 || next_votes_count > kMaxVotesInPacket) {
+    LOG(log_er_) << "Receive " << next_votes_count << " next votes from peer " << packet_data.from_node_id_
                  << ". The peer may be a malicious player, will be disconnected";
     disconnect(packet_data.from_node_id_, dev::p2p::UserReason);
     return;
--- a/libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
+++ b/libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
@@ -40,7 +40,7 @@ TaraxaCapability::TaraxaCapability(std::weak_ptr<dev::p2p::Host> host, const dev
       node_stats_(nullptr),
       packets_handlers_(std::make_shared<PacketsHandler>()),
       thread_pool_(std::make_shared<TarcapThreadPool>(conf.network_packets_processing_threads, node_addr)),
-      periodic_events_tp_(1, false) {
+      periodic_events_tp_(std::make_shared<util::ThreadPool>(kPeriodicEventsThreadCount, false)) {
   LOG_OBJECTS_CREATE("TARCAP");
 
   assert(host.lock());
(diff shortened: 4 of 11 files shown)


gaoqi8649 (approved):

MatusKysel (approved):

kstdl (approved):

libraries/core_libs/network/include/network/tarcap/packets_handlers/get_votes_sync_packet_handler.hpp
```diff
@@ -21,6 +21,8 @@ class GetVotesSyncPacketHandler final : public ExtVotesPacketHandler {
 
   std::shared_ptr<PbftManager> pbft_mgr_;
   std::shared_ptr<NextVotesManager> next_votes_mgr_;
+
+  const uint32_t kMaxVotesInPacket{1000};
```
JakubFornadel, line 25: shouldnt we check this limit when receiving votes packet ?

mfrankovi: Yes, I will add that.

libraries/core_libs/network/src/tarcap/packets_handlers/get_votes_sync_packet_handler.cpp
```diff
@@ -39,6 +39,12 @@ void GetVotesSyncPacketHandler::process(const PacketData &packet_data, const std
     std::vector<std::shared_ptr<Vote>> send_next_votes_bundle;
     for (auto &v : next_votes_bundle) {
       if (!peer->isVoteKnown(v->getHash())) {
+        if (send_next_votes_bundle.size() == kMaxVotesInPacket) {
+          sendPbftNextVotes(packet_data.from_node_id_, send_next_votes_bundle);
+          send_next_votes_bundle.clear();
+          // Allow other processing when sending large number of votes
+          thisThreadSleepForMilliSeconds(100);
```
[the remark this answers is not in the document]
mfrankovi: Sometimes there can be a lot of votes to actually send out and at the same time this could block other packets from being sent/processed like transactions and blocks. So the sleep is just to allow other processing on other threads to do its work. This is more a problem on the receiving end and not so much when we send it. We had this issue on testnet with transactions when we got stuck and we had a large transaction pool. Status message was not processed in time since we would flood the connection with transaction packet.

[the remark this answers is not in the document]
MatusKysel: I see what you want to achieve here, but sleep just blocks this thread and actually prolongs it's execution. So you are doing exact opposite. You should check how we handle trxs, and re-do it that way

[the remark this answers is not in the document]
mfrankovi: I made changes to use the periodic threadpool for this.

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -34,13 +36,17 @@ void ExtVotesPacketHandler::sendPbftVote(dev::p2p::NodeID const &peer_id, std::s
...
-                                              std::vector<std::shared_ptr<Vote>> const &send_next_votes_bundle) {
+                                              std::vector<std::shared_ptr<Vote>> const &send_next_votes_bundle,
+                                              uint32_t index) {
   if (send_next_votes_bundle.empty()) {
     return;
   }
 
-  dev::RLPStream s(send_next_votes_bundle.size());
-  for (auto const &next_vote : send_next_votes_bundle) {
+  uint32_t vote_count_to_send = std::min((size_t)kMaxVotesInPacket, send_next_votes_bundle.size() - index);
```
JakubFornadel, line 45: (size_t) is old style casting, static_cast<> is preffered

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
```diff
@@ -23,8 +25,8 @@ void VotesSyncPacketHandler::validatePacketRlpFormat([[maybe_unused]] const Pack
 
 void VotesSyncPacketHandler::process(const PacketData &packet_data, const std::shared_ptr<TaraxaPeer> &peer) {
   const auto next_votes_count = packet_data.rlp_.itemCount();
-  if (next_votes_count == 0) {
-    LOG(log_er_) << "Receive 0 next votes from peer " << packet_data.from_node_id_
+  if (next_votes_count == 0 || next_votes_count > kMaxVotesInPacket) {
```
MatusKysel, line 28: because of this we need to mark it as breaking-change

libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
```diff
@@ -40,7 +40,7 @@ TaraxaCapability::TaraxaCapability(std::weak_ptr<dev::p2p::Host> host, const dev
       node_stats_(nullptr),
       packets_handlers_(std::make_shared<PacketsHandler>()),
       thread_pool_(std::make_shared<TarcapThreadPool>(conf.network_packets_processing_threads, node_addr)),
-      periodic_events_tp_(1, false) {
+      periodic_events_tp_(std::make_shared<util::ThreadPool>(1, false)) {
```
JakubFornadel, line 43: we should probably have more than 1 thread here as we are starting to use it for many things ?

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
JakubFornadel, line 28: we should probably mark the peer malicious if he sends more votes than `kMaxVotesInPacket`, but we would need to support 2 latest node versions network protocols for that ? Otherwise we would mark malicious all of the nodes, who simply did not update to the newest version yet

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -53,6 +59,12 @@ void ExtVotesPacketHandler::sendPbftNextVotes(dev::p2p::NodeID const &peer_id,
       }
```
MatusKysel, line 59: ^^^ you are marking all votes as known before they are send

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -53,6 +59,12 @@ void ExtVotesPacketHandler::sendPbftNextVotes(dev::p2p::NodeID const &peer_id,
...
     }
   }
+
+  if (send_next_votes_bundle.size() - index > kMaxVotesInPacket) {
+    periodic_events_tp_->post(100, [this, index, peer_id, send_next_votes_bundle] {
+      sendPbftNextVotes(peer_id, send_next_votes_bundle, index + kMaxVotesInPacket);
```
MatusKysel, line 65: we should change it to move semantics for send_next_votes_bundle as we are making copy here

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -34,13 +36,17 @@ void ExtVotesPacketHandler::sendPbftVote(dev::p2p::NodeID const &peer_id, std::s
 }
 
 void ExtVotesPacketHandler::sendPbftNextVotes(dev::p2p::NodeID const &peer_id,
-                                              std::vector<std::shared_ptr<Vote>> const &send_next_votes_bundle) {
+                                              std::vector<std::shared_ptr<Vote>> const &send_next_votes_bundle,
+                                              uint32_t index) {
   if (send_next_votes_bundle.empty()) {
```
MatusKysel, line 41: I think here should be a check if the peer still exist (is connected)

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -53,6 +59,12 @@ void ExtVotesPacketHandler::sendPbftNextVotes(dev::p2p::NodeID const &peer_id,
       }
     }
   }
+
+  if (send_next_votes_bundle.size() - index > kMaxVotesInPacket) {
+    periodic_events_tp_->post(100, [this, index, peer_id, send_next_votes_bundle] {
```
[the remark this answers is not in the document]
mfrankovi: So we have cases where we might be sending very large number of votes or transactions. Usually this only happens after a restart or if a network got stuck for some time. Any thread that sends it in one or more packets would be blocked for some time. Currently we have this one thread in periodic_events_tp_ thread pool. So I think we either need to have multiple threads so that this kind of long operation does not block anything else or we have less threads but delay between sending so that other stuff can be done at the same time as well. Also there is always consideration about the receiving end. If we are sending more transactions/votes than what the other side can process there will be some packets buffering on the other end but maybe we are fine with that.

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
mfrankovi: Marked

libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
mfrankovi: Increase threads to 2 and put that in a constant

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
mfrankovi: yes, but we are recently introducing a lot of breaking changes. I think we need a way to handle this when we plan to do an update on mainnet and then deal with all of these breaking changes together and not one by one

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
mfrankovi: no longer applies

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
mfrankovi: no longer applies

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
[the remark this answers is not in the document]
mfrankovi: I have removed delays from sending multiple packets

libraries/common/include/common/constants.hpp
```diff
@@ -19,6 +19,11 @@ constexpr uint16_t kOnePercent = 100;
 constexpr uint16_t kMaxLevelsPerPeriod = 100;
 constexpr uint32_t kDagExpiryLevelLimit = 1000;
 
+const uint32_t kMaxVotesInPacket{1000};
+const uint32_t kMaxTransactionsInPacket{1000};
+
+const uint32_t kPeriodicEventsThreadCount{2};
```
JakubFornadel, line 25: so what is the purpose/idea behind this common constants file ? Because all of these new constants are not shared between multiple files/modules. Those are specific constants used on one specific file and could be part of the headers

mfrankovi: I am fine either way. I am not sure when we introduced this file but I do like to have similar constants grouped. For instance kMaxVotesInPacket and  kMaxTransactionsInPacket I would put in the same file but maybe it should be in somewhere under network. I think that it would be good to have constants defining number of threads used in different places also to be grouped so that we can easily see how many threads we actually startup in our node but I can see arguments against it as well.

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -39,19 +39,28 @@ void ExtVotesPacketHandler::sendPbftNextVotes(dev::p2p::NodeID const &peer_id,
...
 
-  if (sealAndSend(peer_id, SubprotocolPacketType::VotesSyncPacket, std::move(s))) {
-    LOG(log_nf_) << "Send out size of " << send_next_votes_bundle.size() << " PBFT next votes to " << peer_id;
-    if (auto peer = peers_state_->getPeer(peer_id)) {
-      for (auto const &v : send_next_votes_bundle) {
-        peer->markVoteAsKnown(v->getHash());
+    if (sealAndSend(peer_id, SubprotocolPacketType::VotesSyncPacket, std::move(s))) {
+      LOG(log_nf_) << "Send out size of " << send_next_votes_bundle.size() << " PBFT next votes to " << peer_id;
+      if (auto peer = peers_state_->getPeer(peer_id)) {
+        for (auto const &v : send_next_votes_bundle) {
```
MatusKysel, line 57: you are again marking votes that are not send, or I am missing something?

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -39,19 +39,28 @@ void ExtVotesPacketHandler::sendPbftNextVotes(dev::p2p::NodeID const &peer_id,
...
+      LOG(log_dg_) << "Send out next vote " << next_vote->getHash() << " to peer " << peer_id;
+    }
 
-  if (sealAndSend(peer_id, SubprotocolPacketType::VotesSyncPacket, std::move(s))) {
-    LOG(log_nf_) << "Send out size of " << send_next_votes_bundle.size() << " PBFT next votes to " << peer_id;
-    if (auto peer = peers_state_->getPeer(peer_id)) {
-      for (auto const &v : send_next_votes_bundle) {
-        peer->markVoteAsKnown(v->getHash());
+    if (sealAndSend(peer_id, SubprotocolPacketType::VotesSyncPacket, std::move(s))) {
+      LOG(log_nf_) << "Send out size of " << send_next_votes_bundle.size() << " PBFT next votes to " << peer_id;
```
MatusKysel, line 55: also this log msg is valid only for old behaviour as you are sending them by chunks now

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
mfrankovi: You are correct. Please review again, it should be fixed now

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
mfrankovi: Logging for each chunk

Issue #1461, opened 2022-01-19
Create Debug log profile in configs for users


JakubFornadel:
## Task Description

Create Debug log profile in configs with preset channels so if some users complain, we can just tell them to enable this profile, run node for a while and send us the logs.

## Epic Parent

[Feature: LOgging refactor](https://github.com/Taraxa-project/taraxa-node/issues/1501).

PR #1698, opened 2022-04-29
feat: add debug log profile that not writes to config file

kstdl:
## Purpose
#1461 

During node creation add debug logging profile which can be enabled by adding `--log-configurations "debug"` to `taraxad` arguments


gaoqi8649 (approved):
I think put all debug logs in a separate folder will be better, like:
.taraxa/logs/user and .taraxa/logs/debug

kstdl:
> I think put all debug logs in a separate folder will be better, like: .taraxa/logs/user and .taraxa/logs/debug

Makes sense. I changed path and tested that it works correctly

libraries/cli/src/config.cpp
```diff
@@ -247,6 +252,26 @@ bool Config::nodeConfigured() { return node_configured_; }
...
+        {
+          "type": "file",
+          "file_name": "debug/TaraxaDebug_N1_%m%d%Y_%H%M%S_%5N.log",
+          "rotation_size": 10000000,
+          "time_based_rotation": "0,0,0",
+          "format": "%ThreadID% %ShortNodeId% %Channel% [%TimeStamp%] %SeverityStr%: %Message%",
+          "max_size": 1000000000
+        }
+      ]
+    })";
```
MatusKysel, line 270: how does this work for other configurations e.g. --log-configurations "trace"? or specific channel  as per help msg `Log confifugrations to use: [channel:level, ....]`

MatusKysel (approved):

libraries/cli/src/config.cpp
kstdl: I didn't notice that, but as I see it is just wrong help message. So It just enables predefined configuration with name you are specifying (see [`tools.cpp:106`](https://github.com/Taraxa-project/taraxa-node/blob/1cbb8bd9ff5b30f92a0bf588864c6b1868281eb6/libraries/cli/src/tools.cpp#L106)).

libraries/cli/src/config.cpp
```diff
@@ -247,6 +252,26 @@ bool Config::nodeConfigured() { return node_configured_; }
...
+  auto debug_config = util::readJsonFromString(debug_config_json);
+  logging_config["configurations"].append(debug_config);
```
JakubFornadel, line 272: what happens if someone already has "debug" profile in his config and we add different profile here with the same name ?

kstdl: Right now we will have 2 configs with "debug" name. So with `--log-configurations "debug"` both of them will be enabled

gaoqi8649 (approved):

PR #1705, opened 2022-05-03
Release/testnet v1.0.3

JakubFornadel:

### DIFF
--- a/libraries/cli/include/cli/devnet_config.hpp
+++ b/libraries/cli/include/cli/devnet_config.hpp
@@ -214,7 +214,7 @@ constexpr std::string_view devnet_json = R"foo({
         },
         "execution_options": {
           "disable_gas_fee": false,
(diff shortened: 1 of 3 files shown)


libraries/cli/include/cli/devnet_config.hpp
```diff
@@ -214,7 +214,7 @@ constexpr std::string_view devnet_json = R"foo({
         },
         "execution_options": {
           "disable_gas_fee": false,
-          "disable_nonce_check": false
+          "disable_nonce_check": true
```
MatusKysel, line 217: it should be changed for testnet only

JakubFornadel: why ? Devnet has the same issue and it wont work because of this so we should probably find the fix locally and let the networks run ?

mfrankovi (approved):

PR #1708, opened 2022-05-03
fix: pbft_next_votes_sync_in_behind_round seg fault

mfrankovi:
In test pbft_next_votes_sync_in_behind_round there was a segfault caused by DagManager calling trx_mgr after it was set to nullptr. Fix is not to set trx_mgr to nullptr. I am not sure what was the reason this was ever set to nullptr. It was actually done by a commit by me 2 years ago when the code base was different.

### DIFF
--- a/libraries/core_libs/consensus/src/dag/dag.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag.cpp
@@ -317,7 +317,6 @@ void DagManager::stop() {
     return;
   }
   std::unique_lock lock(mutex_);
-  trx_mgr_ = nullptr;
   block_worker_.join();
 }
 


MatusKysel (approved):

JakubFornadel (approved):

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -317,7 +317,6 @@ void DagManager::stop() {
     return;
   }
   std::unique_lock lock(mutex_);
-  trx_mgr_ = nullptr;
```
gaoqi8649, line 320: I don't get the fix.
DagManager::stop only be called when node stop https://github.com/Taraxa-project/taraxa-node/blob/develop/libraries/core_libs/node/src/node.cpp#L287
Node stop only happen when end of unit test. Why does that cause seg fault?

mfrankovi: there is a packet being processed at the time we are stopping the node. At that time trx_mgr_ is set to nullptr. At that point packet that is being processed which invokes DagManager methods which require trx_mgr_. The segfault actually happened here:
https://github.com/Taraxa-project/taraxa-node/blob/70fe5b968f107022a568b129d4d457da0e721987/libraries/core_libs/consensus/src/dag/dag.cpp#L806

because trx_mgr_ is nullptr.
Crash was easily reproducible, I could get it regularly. After the fix I do not see the segfault happening anymore.

gaoqi8649 (approved):

PR #1712, opened 2022-05-05
chore: enable nonce skipping

MatusKysel:
## Purpose
Fixes https://github.com/Taraxa-project/taraxa-node/issues/1711

### DIFF
--- a/libraries/cli/include/cli/default_config.hpp
+++ b/libraries/cli/include/cli/default_config.hpp
@@ -198,7 +198,8 @@ constexpr std::string_view default_json = R"foo({
         },
         "execution_options": {
           "disable_gas_fee": false,
-          "disable_nonce_check": false
+          "disable_nonce_check": false,
+          "enable_nonce_skipping": true
         },
         "genesis_balances": {
           "2cd4da7d3b345e022ca7e997c2bb3276a4d3d2e9": "0x1027e72f1f12813088000000",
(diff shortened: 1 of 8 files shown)


libraries/cli/include/cli/default_config.hpp
```diff
@@ -183,7 +183,8 @@ constexpr std::string_view default_json = R"foo({
         },
         "execution_options": {
           "disable_gas_fee": false,
-          "disable_nonce_check": false
+          "disable_nonce_check": false,
```
gaoqi8649, line 186: Do we still need disable_nonce_check? I think if enable_nonce_skipping is true, we must have nonce check

gaoqi8649:
If a trx has been finalized in a PBFT block, but failed execution in EVM. I think we will not finalize the trx again

gaoqi8649 (approved):

libraries/cli/include/cli/default_config.hpp
MatusKysel: I think we should keep nonce check value, just in case we want to completely disable this option.

MatusKysel:
> If a trx has been finalized in a PBFT block, but failed execution in EVM. That will not finalize the trx again
> If there are t1, t2, t4 in trx queue. Do we have protection of only be able to add t1, t2 to DAG block, and keep t4 in trx queue because the node has not have t3 yet?

No yet, and this protection is little bit more complicated than skipping the nonce

mfrankovi (approved):

PR #1714, opened 2022-05-06
chore: low nonce and insufficient balance transactions handling

mfrankovi:
Transactions with low nonce and insufficient balance should not be placed into transactions pool to propose dag blocks from but at the same time there are scenarios when such transactions might be needed by incoming dag blocks.

My proposed solution is keeping these transactions in a separate memory map. Once a dag block arrives if it is missing a transaction it will check in these low nonce and insufficient balance transactions as well.

These transactions will expire if no dag blocks arrive after a certain number of periods are executed.

Insufficient balance transactions have different handling than low nonce because there is a chance that such transactions will have sufficient balance after a new block is executed so on block execution they might be moved to regular transactions pool if this is the case

### DIFF
--- a/libraries/core_libs/consensus/include/transaction/transaction_manager.hpp
+++ b/libraries/core_libs/consensus/include/transaction/transaction_manager.hpp
@@ -10,6 +10,8 @@
 
 namespace taraxa {
 
+enum class TransactionStatus { Verified = 0, Invalid, LowNonce, InsufficentBalance };
+
 class DagBlock;
 class DagManager;
 class FullNode;
@@ -46,13 +48,20 @@ class TransactionManager : public std::enable_shared_from_this<TransactionManage
   std::pair<bool, std::string> insertTransaction(const std::shared_ptr<Transaction> &trx);
 
   /**
+   * @brief Invoked when block finalized in final chain
+   *
+   * @param block_number block number finalized
+   */
+  void blockFinalized(uint64_t block_number);
+
+  /**
    * @brief Inserts batch of verified transactions to transaction pool
    *
    * @note Some of the transactions might be already processed -> they are not processed and inserted again
    * @param txs transactions to be processed
    * @return number of successfully inserted unseen transactions
    */
-  uint32_t insertValidatedTransactions(SharedTransactions &&txs);
+  uint32_t insertValidatedTransactions(std::vector<std::pair<std::shared_ptr<Transaction>, TransactionStatus>> &&txs);
 
   /**
    * @brief Marks transaction as known (was successfully verified and pushed into the tx pool)
@@ -127,7 +136,7 @@ class TransactionManager : public std::enable_shared_from_this<TransactionManage
   std::shared_ptr<Transaction> getNonFinalizedTransaction(trx_hash_t const &hash) const;
   unsigned long getTransactionCount() const;
   void recoverNonfinalizedTransactions();
-  std::pair<bool, std::string> verifyTransaction(const std::shared_ptr<Transaction> &trx) const;
+  std::pair<TransactionStatus, std::string> verifyTransaction(const std::shared_ptr<Transaction> &trx) const;
 
  private:
   /**
--- a/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
@@ -43,41 +43,42 @@ uint64_t TransactionManager::estimateTransactionGas(std::shared_ptr<Transaction>
   return result.gas_used;
 }
 
-std::pair<bool, std::string> TransactionManager::verifyTransaction(const std::shared_ptr<Transaction> &trx) const {
+std::pair<TransactionStatus, std::string> TransactionManager::verifyTransaction(
+    const std::shared_ptr<Transaction> &trx) const {
   // ONLY FOR TESTING
   if (!final_chain_) [[unlikely]] {
-    return {true, ""};
+    return {TransactionStatus::Verified, ""};
   }
 
   if (trx->getChainID() != conf_.chain.chain_id) {
-    return {false, "chain_id mismatch"};
+    return {TransactionStatus::Invalid, "chain_id mismatch"};
   }
 
   // Ensure the transaction doesn't exceed the current block limit gas.
   if (FinalChain::GAS_LIMIT < trx->getGas()) {
-    return {false, "invalid gas"};
+    return {TransactionStatus::Invalid, "invalid gas"};
   }
 
   try {
     trx->getSender();
   } catch (Transaction::InvalidSignature const &) {
-    return {false, "invalid signature"};
+    return {TransactionStatus::Invalid, "invalid signature"};
   }
 
   const auto account = final_chain_->get_account(trx->getSender()).value_or(taraxa::state_api::ZeroAccount);
 
   // Ensure the transaction adheres to nonce ordering
   if (account.nonce && account.nonce > trx->getNonce()) {
-    return {false, "nonce too low"};
+    return {TransactionStatus::LowNonce, "nonce too low"};
   }
 
   // Transactor should have enough funds to cover the costs
   // cost == V + GP * GL
   if (account.balance < trx->getCost()) {
-    return {false, "insufficient balance"};
+    return {TransactionStatus::InsufficentBalance, "insufficient balance"};
   }
 
-  return {true, ""};
+  return {TransactionStatus::Verified, ""};
 }
 
 bool TransactionManager::checkMemoryPoolOverflow() {
@@ -114,11 +115,12 @@ std::pair<bool, std::string> TransactionManager::insertTransaction(const std::sh
     }
   }
 
-  if (const auto [is_valid, reason] = verifyTransaction(trx); !is_valid) {
+  const auto [status, reason] = verifyTransaction(trx);
+  if (status != TransactionStatus::Verified) {
     return {false, reason};
   }
 
-  if (insertValidatedTransactions({trx})) {
+  if (insertValidatedTransactions({{trx, status}})) {
     return {true, "Can not insert transactions"};
   } else {
     const auto period = db_->getTransactionPeriod(trx->getHash());
@@ -130,8 +132,9 @@ std::pair<bool, std::string> TransactionManager::insertTransaction(const std::sh
   }
 }
 
-uint32_t TransactionManager::insertValidatedTransactions(SharedTransactions &&txs) {
-  SharedTransactions unseen_txs;
+uint32_t TransactionManager::insertValidatedTransactions(
+    std::vector<std::pair<std::shared_ptr<Transaction>, TransactionStatus>> &&txs) {
+  std::vector<std::pair<std::shared_ptr<Transaction>, TransactionStatus>> unseen_txs;
   std::vector<trx_hash_t> txs_hashes;
 
   if (txs.empty()) {
@@ -143,12 +146,14 @@ uint32_t TransactionManager::insertValidatedTransactions(SharedTransactions &&tx
     return 0;
   }
   txs_hashes.reserve(txs.size());
-  std::transform(txs.begin(), txs.end(), std::back_inserter(txs_hashes), [](const auto &t) { return t->getHash(); });
+  std::transform(txs.begin(), txs.end(), std::back_inserter(txs_hashes),
+                 [](const auto &t) { return t.first->getHash(); });
 
   // This lock synchronizes inserting and removing transactions from transactions memory pool with database insertion.
   // It is very important to lock checking the db state of transaction together with transaction pool checking to be
   // protected from new DAG block and Sync block transactions insertions which are inserted directly in the database.
   std::unique_lock transactions_lock(transactions_mutex_);
+  const auto last_block_number = final_chain_->last_block_number();
 
   // Check the db with a multiquery if transactions are really new
   auto db_seen = db_->transactionsInDb(txs_hashes);
@@ -157,16 +162,16 @@ uint32_t TransactionManager::insertValidatedTransactions(SharedTransactions &&tx
       unseen_txs.push_back(std::move(txs[i]));
     } else {
       // In case we received a new tx that is already in db, mark it as known in cache
-      markTransactionKnown(txs[i]->getHash());
+      markTransactionKnown(txs[i].first->getHash());
     }
   }
 
   size_t trx_inserted_count = 0;
   // Save transactions in memory pool
   for (auto &trx : unseen_txs) {
-    auto tx_hash = trx->getHash();
+    auto tx_hash = trx.first->getHash();
     LOG(log_dg_) << "Transaction " << tx_hash << " inserted in trx pool";
-    if (transactions_pool_.insert(std::move(trx))) {
+    if (transactions_pool_.insert(std::move(trx), last_block_number)) {
       trx_inserted_count++;
       markTransactionKnown(tx_hash);
     }
@@ -322,7 +327,7 @@ void TransactionManager::moveNonFinalizedTransactionsToTransactionsPool(std::uno
       db_->removeTransactionToBatch(trx_hash, write_batch);
       nonfinalized_transactions_in_dag_.erase(trx_hash);
       auto transaction = trx->second;
-      transactions_pool_.insert(std::move(transaction));
+      transactions_pool_.insert({std::move(transaction), TransactionStatus::Verified});
     }
   }
   db_->commitWriteBatch(write_batch);
@@ -347,6 +352,11 @@ bool TransactionManager::checkBlockTransactions(DagBlock const &blk) {
   }
 
   return true;
+}
+
+void TransactionManager::blockFinalized(uint64_t block_number) {
+  std::unique_lock transactions_lock(transactions_mutex_);
+  transactions_pool_.blockFinalized(block_number);
 }
 
 }  // namespace taraxa
--- a/libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
@@ -1,5 +1,7 @@
 #include "transaction/transaction_queue.hpp"
 
+#include "transaction/transaction_manager.hpp"
+
 namespace taraxa {
 
 auto priorityComparator = [](const std::shared_ptr<Transaction> &first, const std::shared_ptr<Transaction> &second) {
@@ -15,13 +17,21 @@ TransactionQueue::TransactionQueue() : priority_queue_{priorityComparator} {}
 
 size_t TransactionQueue::size() const { return hash_queue_.size(); }
 
-bool TransactionQueue::contains(const trx_hash_t &hash) const { return hash_queue_.contains(hash); }
+bool TransactionQueue::contains(const trx_hash_t &hash) const {
+  return hash_queue_.contains(hash) || non_proposable_transactions_.contains(hash);
+}
 
 std::shared_ptr<Transaction> TransactionQueue::get(const trx_hash_t &hash) const {
-  const auto it = hash_queue_.find(hash);
-  if (it == hash_queue_.end()) return nullptr;
+  if (const auto it = hash_queue_.find(hash); it != hash_queue_.end()) {
+    return *(it->second);
+  }
 
-  return *(it->second);
+  if (const auto transaction = non_proposable_transactions_.find(hash);
+      transaction != non_proposable_transactions_.end()) {
+    return transaction->second.second;
+  }
+
+  return nullptr;
 }
 
 std::vector<std::shared_ptr<Transaction>> TransactionQueue::get(uint64_t count) const {
@@ -52,18 +62,43 @@ bool TransactionQueue::erase(const trx_hash_t &hash) {
   return true;
 }
 
-bool TransactionQueue::insert(std::shared_ptr<Transaction> &&transaction) {
-  assert(transaction);
-  if (hash_queue_.contains(transaction->getHash())) return false;
+bool TransactionQueue::insert(std::pair<std::shared_ptr<Transaction>, TransactionStatus> &&transaction,
+                              uint64_t last_block_number) {
+  assert(transaction.first);
+  const auto tx_hash = transaction.first->getHash();
 
-  const auto tx_hash = transaction->getHash();
-  const auto it = priority_queue_.insert(std::move(transaction));
+  if (contains(tx_hash)) return false;
 
-  // This assert is here to check if priorityComparator works correctly. If object is not inserted, then there could be
-  // something wrong with comparator
-  assert(it != priority_queue_.end());
-  hash_queue_[tx_hash] = it;
+  switch (transaction.second) {
+    case TransactionStatus::Verified: {
+      const auto it = priority_queue_.insert(std::move(transaction.first));
+
+      // This assert is here to check if priorityComparator works correctly. If object is not inserted, then there could
+      // be something wrong with comparator
+      assert(it != priority_queue_.end());
+      hash_queue_[tx_hash] = it;
+    } break;
+    case TransactionStatus::LowNonce:
+    case TransactionStatus::InsufficentBalance:
+      if (non_proposable_transactions_.size() < kNonProposableTransactionsLimit)
+        non_proposable_transactions_[tx_hash] = {last_block_number, transaction.first};
+      else
+        return false;
+      break;
+    default:
+      assert(false);
+  }
   return true;
+}
+
+void TransactionQueue::blockFinalized(uint64_t block_number) {
+  for (auto it = non_proposable_transactions_.begin(); it != non_proposable_transactions_.end();) {
+    if (it->second.first + kNonProposableTransactionsPeriodExpiryLimit < block_number) {
+      it = non_proposable_transactions_.erase(it);
+    } else {
+      ++it;
+    }
+  }
 }
 
 }  // namespace taraxa
--- a/libraries/core_libs/network/include/network/tarcap/taraxa_peer.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/taraxa_peer.hpp
@@ -65,6 +65,24 @@ class TaraxaPeer : public boost::noncopyable {
 
   const dev::p2p::NodeID &getId() const { return id_; }
 
+  /**
+   * @brief Reports suspicious pacet
+   *
+   * @return true in case suspicious packet count within current minute is greater than kMaxSuspiciousPacketPerMinute
+   */
+  bool reportSuspiciousPacket() {
+    uint64_t now =
+        std::chrono::duration_cast<std::chrono::minutes>(std::chrono::system_clock::now().time_since_epoch()).count();
+    // Reset counter if no suspicious packet sent for over a minute
+    if (now > timestamp_suspicious_packet_) {
+      suspicious_packet_count_ = 1;
+      timestamp_suspicious_packet_ = now;
+    } else {
+      suspicious_packet_count_++;
+    }
+    return suspicious_packet_count_ > kMaxSuspiciousPacketPerMinute;
+  }
+
   std::atomic<bool> syncing_ = false;
   std::atomic<uint64_t> dag_level_ = 0;
   std::atomic<uint64_t> pbft_chain_size_ = 0;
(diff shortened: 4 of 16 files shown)


libraries/core_libs/consensus/include/transaction/transaction_manager.hpp
```diff
@@ -43,6 +45,27 @@ class TransactionManager : public std::enable_shared_from_this<TransactionManage
    */
   std::pair<bool, std::string> insertTransaction(const std::shared_ptr<Transaction> &trx);
 
+  /**
+   * @brief Inserts low nonce transaction
+   *
+   * @param trx transaction
+   */
+  void insertLowNonceTransaction(std::shared_ptr<Transaction> const &trx);
```
JakubFornadel, line 53: const Type

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -166,7 +170,17 @@ std::pair<std::vector<std::shared_ptr<Transaction>>, std::vector<trx_hash_t>> Tr
...
+      if (transaction.second) {
+        result.first.emplace_back(transaction.first.second);
+      } else {
+        transaction = insufficient_balance_transactions_.get(hash);
+        if (transaction.second) {
+          result.first.emplace_back(transaction.first.second);
+        } else {
+          result.second.emplace_back(hash);
+        }
+      }
```
JakubFornadel, line 183: It is pretty hard to read, let's restructure the code little bit:

```
for (...) {
  if (trx) {
    result.first.emplace_back(trx);
    continue;
  }
 
  if (auto transaction = low_nonce_transactions_.get(hash); transaction.second) {
    result.first.emplace_back(transaction.first.second);
    continue;
  }

    if (transaction = insufficient_balance_transactions_.get(hash); transaction.second) {
    result.first.emplace_back(transaction.first.second);
  } else {
    result.second.emplace_back(hash);
  } 
}
```

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -317,13 +331,45 @@ bool TransactionManager::checkBlockTransactions(DagBlock const &blk) {
...
-        LOG(log_er_) << "Block " << blk.getHash() << " has missing transaction " << tx_hash;
-        return false;
+        if (low_nonce_transactions_.count(tx_hash)) {
+          LOG(log_dg_) << "Block " << blk.getHash() << " has low nonce transaction " << tx_hash;
+        } else if (insufficient_balance_transactions_.count(tx_hash)) {
+          LOG(log_dg_) << "Block " << blk.getHash() << " has insufficient balance transaction " << tx_hash;
+        } else {
+          LOG(log_er_) << "Block " << blk.getHash() << " has missing transaction " << tx_hash;
+          return false;
+        }
```
JakubFornadel, line 341: same here. The rule is to eliminate nested if conditions as much as possible because it makes the code easier to read:

for (...) {
  if (transactions_pool_.contains(tx_hash) || nonfinalized_transactions_in_dag_.contains(tx_hash) ||
          db_->transactionFinalized(tx_hash)) {
     continue;
  }

  // rest of the code
}

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
```diff
@@ -75,11 +75,24 @@ void DagSyncPacketHandler::process(const PacketData& packet_data, const std::sha
...
+      case TransactionStatus::Known:
+        continue;
+      case TransactionStatus::LowNonce:
+        trx_mgr_->insertLowNonceTransaction(trx);
+        continue;
+      case TransactionStatus::InsufficentBalance:
+        trx_mgr_->insertInsufficientBalanceTransaction(trx);
+        continue;
+      case TransactionStatus::Verified:
+        break;
```
JakubFornadel, line 95: break ?

JakubFornadel (commented):
So in this design, there is a bug: In both `insertLowNonceTransaction` & `insertInsufficientBalanceTransaction` you should be locking `transactions_mutex_`... and the fact that the bug exists suggests the design itself might not be ideal.

- You could put these 2 new containers `low_nonce_transactions_` & `insufficient_balance_transactions_`  into the TransactionQueue class (in fact they are in queue - just need to wait) 
- TransactionQueue::insert would have also enum type argument(standard, LowNonce, InsufficientBalance) based on which you insert it into the standard queue or the 2 new queues/containers
- Like this, most of the code would be encapsulated and isolated inside the TransactionQueue, which is good, it is more prone to bugs with mutexes etc...

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -21,41 +21,45 @@ TransactionManager::TransactionManager(FullNodeConfig const &conf, std::shared_p
...
+    return {TransactionStatus::Invalid, "invalid signature"};
   }
 
   const auto account = final_chain_->get_account(trx->getSender()).value_or(taraxa::state_api::ZeroAccount);
 
   // Ensure the transaction adheres to nonce ordering
   if (account.nonce && account.nonce > trx->getNonce()) {
-    return {false, "nonce too low"};
+    // Transactions might not be in cache but it could be an old transaction that is already executed
+    if (db_->transactionInDb(trx->getHash())) return {TransactionStatus::Known, "known"};
```
JakubFornadel, line 51: no need to check if tx is in db here, in fact verify function should simply verify tx and nothing else. Every single  transaction, which is going to be included into the transaction pool is checked against db anyway so if you use [this suggested design](), everything would be solved

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -21,41 +21,45 @@ TransactionManager::TransactionManager(FullNodeConfig const &conf, std::shared_p
...
+    return {TransactionStatus::LowNonce, "nonce too low"};
   }
 
   // Transactor should have enough funds to cover the costs
   // cost == V + GP * GL
   if (account.balance < trx->getCost()) {
-    return {false, "insufficient balance"};
+    if (db_->transactionInDb(trx->getHash())) return {TransactionStatus::Known, "known"};
```
JakubFornadel, line 58: same here

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
JakubFornadel: This whole piece of code could be rather inside transactions_pool_.get.

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
JakubFornadel: also this could be inside transactions_pool_.contains

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
```diff
@@ -75,11 +75,24 @@ void DagSyncPacketHandler::process(const PacketData& packet_data, const std::sha
       continue;
     }
 
-    if (const auto [is_valid, reason] = trx_mgr_->verifyTransaction(trx); !is_valid) {
-      std::ostringstream err_msg;
-      err_msg << "DagBlock transaction " << trx->getHash() << " validation failed: " << reason;
-
-      throw MaliciousPeerException(err_msg.str());
+    const auto [status, reason] = trx_mgr_->verifyTransaction(trx);
+    switch (status) {
```
JakubFornadel, line 79: Wth suggested new design, we could get rid of this while switch and use just:

```
if (const auto [status, reason] = trx_mgr_->verifyTransaction(trx); status == Invalid) {
  // mark malicious peer
}
```

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -15,11 +17,30 @@ TransactionQueue::TransactionQueue() : priority_queue_{priorityComparator} {}
...
+    if (transaction != low_nonce_transactions_.end()) {
+      return transaction->second.second;
+    } else {
+      transaction = insufficient_balance_transactions_.find(hash);
+      if (transaction != insufficient_balance_transactions_.end()) {
+        return transaction->second.second;
+      }
+    }
+    return nullptr;
+  }
```
JakubFornadel, line 43:
```
if (it != hash_queue_.end()) {
  return *(it->second);
}

if (auto transaction = low_nonce_transactions_.find(hash); transaction != low_nonce_transactions_.end()) {
  return transaction->second.second;
}

if (transaction = insufficient_balance_transactions_.find(hash); transaction != insufficient_balance_transactions_.end()) {
  return transaction->second.second;
}

return nullptr;
```

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -52,18 +73,50 @@ bool TransactionQueue::erase(const trx_hash_t &hash) {
...
+        insufficient_balance_transactions_[transaction.first->getHash()] = {last_block_number, transaction.first};
+      break;
+    default:
+      assert(false);
+  }
   return true;
 }
 
+void TransactionQueue::blockFinalized(uint64_t block_number) {
+  for (auto it = insufficient_balance_transactions_.begin(); it != insufficient_balance_transactions_.end();) {
```
JakubFornadel, line 106: unnecessary semicolon at the end of for loop

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -52,18 +73,50 @@ bool TransactionQueue::erase(const trx_hash_t &hash) {
...
+    if (it->second.first + kTransactionsExpiryLimit < block_number) {
+      it = insufficient_balance_transactions_.erase(it);
+    } else {
+      ++it;
+    }
+  }
+
+  for (auto it = low_nonce_transactions_.begin(); it != low_nonce_transactions_.end();) {
```
JakubFornadel, line 114: unnecessary semicolon at the end of for loop

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -52,18 +73,50 @@ bool TransactionQueue::erase(const trx_hash_t &hash) {
...
+    if (it->second.first + kTransactionsExpiryLimit < block_number)
```
JakubFornadel, line 115: in the same loop above you use parenthesis, here you dont - inconsistent

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
```diff
@@ -58,30 +58,29 @@ void DagSyncPacketHandler::process(const PacketData& packet_data, const std::sha
...
-  SharedTransactions new_transactions;
+  std::vector<std::pair<std::shared_ptr<Transaction>, TransactionStatus>> new_transactions;
   std::string transactions_to_log;
   for (uint64_t i = 0; i < transactions_count; i++) {
-    std::shared_ptr<Transaction> trx;
+    std::pair<std::shared_ptr<Transaction>, TransactionStatus> trx;
 
     try {
-      trx = std::make_shared<Transaction>(*it++);
+      trx.first = std::make_shared<Transaction>(*it++);
```
JakubFornadel, line 67: I dont see where are you setting trx.second ?

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
```diff
@@ -24,31 +24,33 @@ inline void TransactionPacketHandler::process(const PacketData &packet_data, con
...
+    std::pair<std::shared_ptr<Transaction>, TransactionStatus> transaction;
 
     try {
-      transaction = std::make_shared<Transaction>(packet_data.rlp_[tx_idx].data().toBytes());
+      transaction.first = std::make_shared<Transaction>(packet_data.rlp_[tx_idx].data().toBytes());
     } catch (const Transaction::InvalidSignature &e) {
       throw MaliciousPeerException("Unable to parse transaction: " + std::string(e.what()));
     }
 
     if (dag_blk_mgr_) [[likely]] {  // ONLY FOR TESTING
```
JakubFornadel, line 39: if dag_blk_mgr == nullptr (tests), you dont set transaction.second - is that ok ?

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -349,4 +353,18 @@ bool TransactionManager::checkBlockTransactions(DagBlock const &blk) {
...
+  auto &insufficient_balance_transactions = transactions_pool_.getInsufficientBalanceTransactions();
+  for (auto it = insufficient_balance_transactions.begin(); it != insufficient_balance_transactions.end();) {
+    if (verifyTransaction(it->second.second).first == TransactionStatus::Verified) {
+      it = insufficient_balance_transactions.erase(it);
+      transactions_pool_.insert({it->second.second, TransactionStatus::Verified});
+    } else {
+      it++;
+    }
+  }
+  transactions_pool_.blockFinalized(block_number);
```
JakubFornadel, line 367: this should be called before for loop, otherwise we are processing txs for +1 block than specified in kTransactionsExpiryLimit

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -349,4 +353,18 @@ bool TransactionManager::checkBlockTransactions(DagBlock const &blk) {
   return true;
 }
 
+void TransactionManager::blockFinalized(uint64_t block_number) {
+  std::unique_lock transactions_lock(transactions_mutex_);
+  auto &insufficient_balance_transactions = transactions_pool_.getInsufficientBalanceTransactions();
```
JakubFornadel, line 358: here you are processing  `insufficient_balance_transactions_` after block is finalized. Where are you processing `low_nonce_transactions_` ? I dont see anything except saving txs there and then removing them after they expire

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -114,11 +115,11 @@ std::pair<bool, std::string> TransactionManager::insertTransaction(const std::sh
     }
   }
 
-  if (const auto [is_valid, reason] = verifyTransaction(trx); !is_valid) {
+  if (const auto [status, reason] = verifyTransaction(trx); status != TransactionStatus::Verified) {
     return {false, reason};
   }
 
-  if (insertValidatedTransactions({trx})) {
+  if (insertValidatedTransactions({{trx, TransactionStatus::Verified}})) {
```
JakubFornadel, line 122: why is this hardcoded to TransactionStatus::Verified ?

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -52,18 +73,50 @@ bool TransactionQueue::erase(const trx_hash_t &hash) {
   return true;
 }
 
-bool TransactionQueue::insert(std::shared_ptr<Transaction> &&transaction) {
-  assert(transaction);
-  if (hash_queue_.contains(transaction->getHash())) return false;
+bool TransactionQueue::insert(std::pair<std::shared_ptr<Transaction>, TransactionStatus> &&transaction,
+                              uint64_t last_block_number) {
+  assert(transaction.first);
+  if (hash_queue_.contains(transaction.first->getHash())) return false;
```
JakubFornadel, line 79: Is this check supposed to protect tx queue before duplicate insertion ? If yes, it wont work properly as hash_queue_ is adjusted only in case of TransactionStatus::Verified, not for LowNonce and InsufficentBalance. Also it is somehow related to some comparator but that seems ok that we add txs there only tx status == Verified

@MatusKysel please confirm

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
mfrankovi: low nonce transactions can never become valid because nonce is only increased, never reduced. We only need them to support any dag block that might have included them at the point that the node that created dag block thought they were valid

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
mfrankovi: It was missing. Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
mfrankovi: It is the only possible value but it is better to actually use the result of verification. Fixed

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
mfrankovi: I changed it to check for both lownonce and insufficientbalance.

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
JakubFornadel: ah ok

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
JakubFornadel: should this `transaction.second = TransactionStatus::Verified;` be outside of `if (dag_blk_mgr_)` to acutally fix it ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1634,13 +1634,16 @@ void PbftManager::finalize_(SyncBlock &&sync_block, std::vector<h256> &&finalize
         }
 
         db_->addProposalPeriodDagLevelsMapToBatch(anchor->getLevel() + max_levels_per_period_, period, batch);
+        ptr->finalized();
```
MatusKysel, line 1637: maybe it would be better to use `block_applying_emitter_` as it run on separate thread and it will do the same work. This way we are not slowing down finalization process

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -349,4 +354,18 @@ bool TransactionManager::checkBlockTransactions(DagBlock const &blk) {
...
+  transactions_pool_.blockFinalized(block_number);
+  auto &insufficient_balance_transactions = transactions_pool_.getInsufficientBalanceTransactions();
+  for (auto it = insufficient_balance_transactions.begin(); it != insufficient_balance_transactions.end();) {
+    if (verifyTransaction(it->second.second).first == TransactionStatus::Verified) {
+      it = insufficient_balance_transactions.erase(it);
+      transactions_pool_.insert({it->second.second, TransactionStatus::Verified});
+    } else {
+      it++;
+    }
+  }
```
MatusKysel, line 368: this is ugly, let's rather send `verifyTransaction` as std::function arg and do the whole thing inside the pool, also get rid of the `getInsufficientBalanceTransactions`

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -52,18 +72,51 @@ bool TransactionQueue::erase(const trx_hash_t &hash) {
...
-  assert(transaction);
-  if (hash_queue_.contains(transaction->getHash())) return false;
+bool TransactionQueue::insert(std::pair<std::shared_ptr<Transaction>, TransactionStatus> &&transaction,
+                              uint64_t last_block_number) {
+  assert(transaction.first);
+  if (contains(transaction.first->getHash())) return false;
 
-  const auto tx_hash = transaction->getHash();
-  const auto it = priority_queue_.insert(std::move(transaction));
+  const auto tx_hash = transaction.first->getHash();
```
MatusKysel, line 80: if you do this line before it would be cleaner

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -52,18 +72,51 @@ bool TransactionQueue::erase(const trx_hash_t &hash) {
   return true;
 }
 
-bool TransactionQueue::insert(std::shared_ptr<Transaction> &&transaction) {
-  assert(transaction);
-  if (hash_queue_.contains(transaction->getHash())) return false;
+bool TransactionQueue::insert(std::pair<std::shared_ptr<Transaction>, TransactionStatus> &&transaction,
```
MatusKysel, line 75: also I think there should be some check if the `last_block_number` for other status ten verified maybe just assert

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
MatusKysel: I  do not think we can actually do this, this way are re-validating 1000 trx for 10 blocks :) I guess we should not do it at all. This is place for attack that we multiply by ourselfs

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
```diff
@@ -58,30 +58,30 @@ void DagSyncPacketHandler::process(const PacketData& packet_data, const std::sha
...
+    if (verification.first == TransactionStatus::Invalid) {
       std::ostringstream err_msg;
-      err_msg << "DagBlock transaction " << trx->getHash() << " validation failed: " << reason;
-
+      err_msg << "DagBlock transaction " << trx.first->getHash() << " validation failed: " << verification.second;
       throw MaliciousPeerException(err_msg.str());
     }
-
+    trx.second = verification.first;
     new_transactions.push_back(std::move(trx));
```
MatusKysel, line 85: isn't it cleaner to leave as it is and save the status with transaction here? I guess  using `trx.first` it is just confusing and the status is only used here

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
```diff
@@ -58,30 +58,30 @@ void DagSyncPacketHandler::process(const PacketData& packet_data, const std::sha
...
       continue;
     }
 
-    if (const auto [is_valid, reason] = trx_mgr_->verifyTransaction(trx); !is_valid) {
+    auto verification = trx_mgr_->verifyTransaction(trx.first);
+    if (verification.first == TransactionStatus::Invalid) {
       std::ostringstream err_msg;
-      err_msg << "DagBlock transaction " << trx->getHash() << " validation failed: " << reason;
-
+      err_msg << "DagBlock transaction " << trx.first->getHash() << " validation failed: " << verification.second;
```
MatusKysel, line 81: hmm, I do not think this is good, maybe I am wrong, but doesn't this mean I can spam the network with transactions with 0 funds, everyone will accept them and no one mark me as malicious? I think we need to do it other way. Maybe we should just save those trx, but not the actual dag block and do nothing

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
```diff
@@ -24,31 +24,34 @@ inline void TransactionPacketHandler::process(const PacketData &packet_data, con
...
+        err_msg << "Transaction " << transaction.first->getHash() << " validation failed: " << verification.second;
         throw MaliciousPeerException(err_msg.str());
       }
+      transaction.second = verification.first;
     }
-    received_transactions += transaction->getHash().abridged() + " ";
-    peer->markTransactionAsKnown(transaction->getHash());
+    received_transactions += transaction.first->getHash().abridged() + " ";
+    peer->markTransactionAsKnown(transaction.first->getHash());
     transactions.push_back(std::move(transaction));
```
MatusKysel, line 55: same as https://github.com/Taraxa-project/taraxa-node/pull/1714/files#r873469852 maybe you should insert status here not before as do not care and it complicates everything with first and second

MatusKysel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
mfrankovi: I think that we should have some limit on how many transaction with insufficient funds or low nonce we allow from a node and than mark node as malicious if it goes over this threshold. But the whole idea of having insufficient funds and low nonce is to allow dag blocks that might contain them.

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
mfrankovi: I dont understand this comment

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
mfrankovi: I think that we should. Maybe the number 1000 is too much. But I think this needs to be handled on setting node malicious if they send too many of these transactions in certain time limit. But I would create a separate issue for that as part of general ddos protection.

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
mfrankovi: Fixed

libraries/core_libs/network/include/network/tarcap/taraxa_peer.hpp
```diff
@@ -65,6 +65,24 @@ class TaraxaPeer : public boost::noncopyable {
...
+  /**
+   * @brief Reports suspicious pacet
+   *
+   * @return true in case suspicious packet count within current minute is greater than kMaxSuspiciousPacketPerMinute
+   */
+  bool reportSuspiciousPacket() {
+    uint64_t now =
+        std::chrono::duration_cast<std::chrono::minutes>(std::chrono::system_clock::now().time_since_epoch()).count();
+    // Reset counter if no suspicious packet sent for over a minute
+    if (timestamp_suspicious_packet_ > now) {
```
MatusKysel, line 77: how can this be true?

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
MatusKysel: it is fine now

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -1,5 +1,7 @@
 #include "transaction/transaction_queue.hpp"
 
+#include "transaction/transaction_manager.hpp"
```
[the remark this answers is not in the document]
mfrankovi: TransactionStatus is defined in transaction manager

libraries/core_libs/network/include/network/tarcap/taraxa_peer.hpp
mfrankovi: Fixed and also I added a unit test in NetworkTest to verify this method is acting as expected

JakubFornadel (approved):

PR #1721, opened 2022-05-10
chore: increase number of peer slots on devnet machines

MatusKysel:
## Purpose
Increase number of slots on devnet as nodes failed to connect to each because of this `Disconnect (reason: Peer had too many connections.)`

### DIFF
--- a/libraries/cli/include/cli/devnet_config.hpp
+++ b/libraries/cli/include/cli/devnet_config.hpp
@@ -15,8 +15,8 @@ constexpr std::string_view devnet_json = R"foo({
   "network_simulated_delay": 0,
   "network_transaction_interval": 100,
   "network_bandwidth": 40,
-  "network_ideal_peer_count": 5,
-  "network_max_peer_count": 15,
+  "network_ideal_peer_count": 15,
+  "network_max_peer_count": 30,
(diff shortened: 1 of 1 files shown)


libraries/cli/include/cli/devnet_config.hpp
```diff
@@ -15,8 +15,8 @@ constexpr std::string_view devnet_json = R"foo({
   "network_simulated_delay": 0,
   "network_transaction_interval": 100,
   "network_bandwidth": 40,
-  "network_ideal_peer_count": 5,
-  "network_max_peer_count": 15,
+  "network_ideal_peer_count": 15,
+  "network_max_peer_count": 30,
```
gaoqi8649, line 19: But that will also change for consensus nodes. 
I think the fix should only change on RPC nodes, and not for consensus nodes

JakubFornadel (approved):

libraries/cli/include/cli/devnet_config.hpp
MatusKysel: it does not matter how many consensus node have it is devnet. Anyway we increased those values for rest of the networks - mainnet, testnet

mfrankovi (approved):

PR #1725, opened 2022-05-11
feat: prepare multiple capabilities support

kstdl:


JakubFornadel:
So without really going into the detailed review, the reason why I wanted to see a draft is that in my view this task can be implemented basically in few lines. I might be of course wrong as I did not spend too much time getting into the details so I might be missing something but here is how I would do it:

-  the architecture fot supporting 2 different protocol version is already there and this [is the place of importance](https://github.com/Taraxa-project/taraxa-node/blob/develop/libraries/core_libs/network/src/network.cpp#L40):
```
auto construct_capabilities = [&, this](auto host) {
    assert(!host.expired());

    taraxa_capability_ = std::make_shared<network::tarcap::TaraxaCapability>(host, key, conf_, db, pbft_mgr, pbft_chain,
                                                                             vote_mgr, next_votes_mgr, dag_mgr,
                                                                             dag_blk_mgr, trx_mgr, key.address());
    return dev::p2p::Host::CapabilityList{taraxa_capability_};
};
 
host_ = dev::p2p::Host::make(net_version, construct_capabilities, key, net_conf, taraxa_net_conf, network_file_path);
```
- we need to do there is the possibility to create 2 TaraxaCapability objects 
- TaraxaCapability class must support having dynamic NET_VERSION initialized during construction, not like now (see method unsigned `TaraxaCapability::version() const { return TARAXA_NET_VERSION; }`) 
- method TaraxaCapability::registerPacketHandlers should not be called inside constructor(it would be actually deleted) and we would support registering handlers one by one manually - this way you can have two objects of the same type(TaraxaCapability) with different packet handlers registered
- each specific PacketHandler class would declare it's methods (validatePacketRlpFormat & process) as virtual - this way we can reause most of the packet handlers and simply create new derived handler class for the one that changed
- no need to have different SubprotocolPacketType - we can share the same between two capabilities: the rule is that you just can only add new packet types

So overall if we would need to support 2 different capabilities for 2 different protocol versions, we would only need to create new directory lets say old_tarcap with single new class of the packet handler that changed + register that new capability with different registered packet handlers in  network and thats it

MatusKysel:
Why we are evening working on this task? I understand both approaches, but both of them could be implemented when we need it. @JakubFornadel architecture could be use in case of some changes in handler, but @kstdl architecture will be probably needed in case of changes in PocketTypes (adding, removing new type). Anyways any of this could be done when it is needed, right now all those changes are just adding unnecessary complexity without actually adding anything new (maybe in time when we will need something like this whole taraxa capability will be different, who knows)

JakubFornadel:
> Why we are evening working on this task? I understand both approaches, but both of them could be implemented when we need it. @JakubFornadel architecture could be use in case of some changes in handler, but @kstdl architecture will be probably needed in case of changes in PocketTypes (adding, removing new type). Anyways any of this could be done when it is needed, right now all those changes are just adding unnecessary complexity without actually adding anything new (maybe in time when we will need something like this whole taraxa capability will be different, who knows)

we are working on it because in my ming it was supposed to be 1 day task and we did not have anything else at that time. I agree that it does not have priority and it is not worth spending too much time on it. Anyway we cant take it back as we already used 1 week for it so lets finish it. My architecture also works either with adding or removing new type -> you are controlling what packet types you support by registering packet handlers for specific packet types. So if you remove some packet type in new version, you simply wont register handler for it and after next protocol version increase you can even delete it from packetType enum....

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -73,17 +73,6 @@ void StatusPacketHandler::process(const PacketData& packet_data, const std::shar
...
-    // We need logic when some different node versions might still be compatible
-    if (node_major_version != TARAXA_MAJOR_VERSION || node_minor_version != TARAXA_MINOR_VERSION) {
-      // Log this only if we have 0 peers, so it is error of this node
-      LOG((peers_state_->getPeersCount()) ? log_nf_ : log_er_)
-          << "Incorrect node version: "
-          << getFormattedVersion({node_major_version, node_minor_version, node_patch_version}) << ", our node version"
-          << TARAXA_VERSION << ", host " << packet_data.from_node_id_.abridged() << " will be disconnected";
-      disconnect(packet_data.from_node_id_, dev::p2p::UserReason);
-      return;
-    }
```
gaoqi8649, line 85: We don't check node version anymore?

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
```diff
@@ -113,7 +102,8 @@ void StatusPacketHandler::process(const PacketData& packet_data, const std::shar
                  << genesis_hash << ", peer pbft chain size " << selected_peer->pbft_chain_size_ << ", peer syncing "
                  << std::boolalpha << selected_peer->syncing_ << ", peer pbft round " << selected_peer->pbft_round_
                  << ", peer pbft previous round next votes size " << selected_peer->pbft_previous_round_next_votes_size_
-                 << ", node major version" << node_major_version << ", node minor version" << node_minor_version;
+                 << ", node major version" << node_major_version << ", node minor version" << node_minor_version
+                 << ", node minor version" << node_patch_version;
```
gaoqi8649, line 106: typo "node patch version"

libraries/core_libs/network/include/network/tarcap/taraxa_capability.hpp
```diff
@@ -35,18 +35,25 @@ class TaraxaPeer;
...
-                   addr_t const &node_addr = {});
+                   unsigned version);
 
   virtual ~TaraxaCapability() = default;
   TaraxaCapability(const TaraxaCapability &ro) = delete;
   TaraxaCapability &operator=(const TaraxaCapability &ro) = delete;
   TaraxaCapability(TaraxaCapability &&ro) = delete;
   TaraxaCapability &operator=(TaraxaCapability &&ro) = delete;
-
+  static std::shared_ptr<TaraxaCapability> make(
```
MatusKysel, line 45: isn't it better to be called make_shared ? Just a question

libraries/core_libs/network/include/network/tarcap/taraxa_capability.hpp
```diff
@@ -100,18 +107,21 @@ class TaraxaCapability : public dev::p2p::CapabilityFace {
 
   // END METHODS USED IN TESTS ONLY
 
+  virtual void initPeriodicEvents(const NetworkConfig &conf, const std::shared_ptr<PbftManager> &pbft_mgr,
```
MatusKysel, line 110: Maybe I am wrong but do they need to be public? maybe protected is enough or?

MatusKysel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
kstdl: Yes. We will work with node that supports the same version of capability
For example node1 has capabilities [1,2] and node2 has [2,3]. So they can communicate with capability with version 2

libraries/core_libs/network/include/network/tarcap/taraxa_capability.hpp
kstdl: Host and some other classes have the same `make` naming. I don't think it is clearifies a lot, but we can rename all of them

MatusKysel: I will leave up to you

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
gaoqi8649: Could you please show me where the code do the check? Thanks

gaoqi8649 (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
kstdl: Here we removing all capabilities that can't be used [Host.cpp:239](https://github.com/Taraxa-project/taraxa-node/blob/cc087eee3e366acafe8b2851d9ba8efb4bbe34da/libraries/aleth/libp2p/Host.cpp#L239). In this line [Host.cpp:275](https://github.com/Taraxa-project/taraxa-node/blob/cc087eee3e366acafe8b2851d9ba8efb4bbe34da/libraries/aleth/libp2p/Host.cpp#L275) we are disconnecting if there is no common capability

libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
```diff
@@ -28,39 +28,54 @@
...
+                         dag_blk_mgr, trx_mgr, node_addr);
 
   // Inits periodic events. Must be called after registerHandlers !!!
-  initPeriodicEvents(conf, pbft_mgr, trx_mgr, packets_stats);
+  initPeriodicEvents(net_conf_, pbft_mgr, trx_mgr, packets_stats_);
 }
 
 void TaraxaCapability::initBootNodes(const std::vector<NodeConfig> &network_boot_nodes, const dev::KeyPair &key) {
-  auto resolveHost = [](string const &addr, uint16_t port) {
+  auto resolveHost = [](const string &addr, uint16_t port) {
```
JakubFornadel, line 78: std::string

libraries/core_libs/network/src/network.cpp
```diff
@@ -37,16 +40,18 @@ Network::Network(NetworkConfig const &config, std::filesystem::path const &netwo
...
-                                                                             vote_mgr, next_votes_mgr, dag_mgr,
-                                                                             dag_blk_mgr, trx_mgr, key.address());
-    return dev::p2p::Host::CapabilityList{taraxa_capability_};
-  };
+  if (!construct_capabilities) {
+    construct_capabilities = [&](std::weak_ptr<dev::p2p::Host> host) {
+      assert(!host.expired());
+
+      auto taraxa_capability =
+          network::tarcap::TaraxaCapability::make(host, key, conf_, TARAXA_NET_VERSION, db, pbft_mgr, pbft_chain,
```
JakubFornadel, line 48: So how does this static `make` method together with `init` method makes it any different than calling `registerPacketHandlers` & `initPeriodicEvents` directly in constructor ?

libraries/core_libs/network/src/tarcap/packets_handlers/status_packet_handler.cpp
gaoqi8649: Thank you

libraries/core_libs/network/src/network.cpp
JakubFornadel: ok, in this case first part of my comment still applies: 
So how does this static make method together with init method makes it any different than calling virtual registerPacketHandlers & initPeriodicEvents directly in constructor ?

PR #1736, opened 2022-05-13
Use template for getSpecificHandler

gaoqi8649:


JakubFornadel:
nevermind. This is not good - leave this idea

PR #1716, opened 2022-05-06
Send votes in bundle of maximum N votes

gaoqi8649:
Fix https://github.com/Taraxa-project/taraxa-node/issues/1700

PR #1737, opened 2022-05-16
chore: create templated getSpecificHandler - no need to use repetitiv…

JakubFornadel:
## Purpose

Note: just a quicker way how to explain this: https://github.com/Taraxa-project/taraxa-node/pull/1716#discussion_r872120851

### DIFF
--- a/libraries/core_libs/network/include/network/tarcap/packets_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handler.hpp
@@ -23,20 +23,24 @@ class PacketsHandler {
   const std::shared_ptr<PacketHandler>& getSpecificHandler(SubprotocolPacketType packet_type) const;
 
   /**
-   * @brief templated getSpecificHandler method, which internally casts fetched PacketHandler to specified
-   * PacketHandlerType
-   * @see non-templated getSpecificHandler method for other specs
+   * @brief templated getSpecificHandler method for getting specific packet handler based on
+   * PacketHandlerType::kPacketType_
+   * @tparam PacketHandlerType
+   *
+   * @return std::shared_ptr<PacketHandlerType>
    */
   template <typename PacketHandlerType>
-  std::shared_ptr<PacketHandlerType> getSpecificHandler(SubprotocolPacketType packet_type);
+  std::shared_ptr<PacketHandlerType> getSpecificHandler();
 
   /**
-   * @brief Registers handler for specific packet type
+   * @brief Registers packet handler
    *
-   * @param packet_type
-   * @param handler
+   * @tparam PacketHandlerType
+   * @tparam Args
+   * @param args to be forwarded to the PacketHandlerType ctor
    */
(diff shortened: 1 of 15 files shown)


libraries/core_libs/network/include/network/tarcap/packets_handler.hpp
```diff
@@ -35,4 +43,9 @@ class PacketsHandler {
   std::unordered_map<SubprotocolPacketType, std::shared_ptr<PacketHandler>> packets_handlers_;
 };
 
+template <typename PacketHandlerType>
+std::shared_ptr<PacketHandlerType> PacketsHandler::getSpecificHandler(SubprotocolPacketType packet_type) {
+  return std::static_pointer_cast<PacketHandlerType>(getSpecificHandler(packet_type));
+}
```
MatusKysel, line 49: would it be possible to have some kind of table so we do not need to specify both? packet type and packet type handler? or maybe `PacketHandlerType` could have some kind of static variable or not sure how it is handled and based on that we would get the type? `PacketHandlerType::packet_type` ???

MatusKysel (approved):

kstdl (approved):

libraries/core_libs/network/include/network/tarcap/packets_handler.hpp
JakubFornadel: yes I was thinking about it too, there could be a big switch and thats about it - not sure what is better solution, have this big switch or just specify the type ?

mfrankovi (approved):

libraries/core_libs/network/include/network/tarcap/packets_handler.hpp
JakubFornadel: @MatusKysel @kstdl fixed, please re - review :)

kstdl (approved):

libraries/core_libs/network/include/network/tarcap/packets_handler.hpp
```diff
@@ -20,19 +20,43 @@ class PacketsHandler {
...
+  void registerHandler(Args&&... args);
 
  private:
   // Map of all packets handlers, factory method selects specific packet handler for processing based on packet type
   std::unordered_map<SubprotocolPacketType, std::shared_ptr<PacketHandler>> packets_handlers_;
 };
 
+template <typename PacketHandlerType>
+std::shared_ptr<PacketHandlerType> PacketsHandler::getSpecificHandler() {
+  return std::static_pointer_cast<PacketHandlerType>(getSpecificHandler(PacketHandlerType::k_packet_type));
```
MatusKysel, line 52: hmm so are we agreed on k_something_something ?

JakubFornadel: Aah I forgot, (yeah I still think its bullshit to have different styles for variables :D) but you guys decided on slack poll so democracy it is - will refactor style...

gaoqi8649 (approved):

PR #1744, opened 2022-05-20
Cert Votes Refactor

gaoqi8649:


mfrankovi:
This is the implementation of this design:
Currently when we save SyncBlock with savePeriodData into period_data column it consists of:

Pbft block
Ordered dag blocks
Ordered unique transactions which were not executed in some previous pbft block
Cert votes
One thing that I never liked is that SyncBlock is not unique across the network. Cert votes may be different from node to node depending which votes that are over 2t+1 we actually saved on each individual node. Reward votes that we include in the next proposed pbft block will be a superset of the cert votes from the previous block.
The idea I also proposed recently is that we save the difference between reward votes and cert votes in the next block would not work because the difference would not be the same across all nodes.
My proposed change and implementation of reward votes would be this:

SyncBlock will no longer contain cert votes for the current period. Instead it will only contain reward votes for the previous period. So SyncBlock would consist of:

Pbft block period N
Ordered dag blocks
Ordered unique transactions which were not executed in some previous pbft block
Reward votes for period N-1
When a pbft block is cert voted on and pbft block is pushed into the chain, cert votes for this latest pbft block in the chain will be kept/saved in separate db column last_block_cert_votes and not saved within period_data as we do now. This will be in db so it will be persistent. This db column will only contain cert votes for the last block in the chain. This will be a small size column which should make read/writes to it fast.
When proposing a new pbft block, this table will be the source for selecting reward votes.

Syncing:
In our current implementation SyncBlock contains 2t+1 votes necessary to validate the block. In this new approach when syncing there will be a need to use rewards votes from block N+1 to validate block N. We already have a queue of blocks waiting for verification so this should be simple to implement. When syncing to the final block in the chain we already have a flag indicating it is the last block. In the new approach with the last block that is syncing, cert votes should be sent as well that are retrieved from last_block_cert_votes column in the db so that the last block can be validated.

With this approach writing to period_data will be final with no updating later needed. All the nodes will have exactly same data in period_data. Difference between the nodes will only be in the last_block_cert_votes

mfrankovi:
This is the implementation of this design:
Currently when we save SyncBlock with savePeriodData into period_data column it consists of:

Pbft block
Ordered dag blocks
Ordered unique transactions which were not executed in some previous pbft block
Cert votes
One thing that I never liked is that SyncBlock is not unique across the network. Cert votes may be different from node to node depending which votes that are over 2t+1 we actually saved on each individual node. Reward votes that we include in the next proposed pbft block will be a superset of the cert votes from the previous block.
The idea I also proposed recently is that we save the difference between reward votes and cert votes in the next block would not work because the difference would not be the same across all nodes.
My proposed change and implementation of reward votes would be this:

SyncBlock will no longer contain cert votes for the current period. Instead it will only contain reward votes for the previous period. So SyncBlock would consist of:

Pbft block period N
Ordered dag blocks
Ordered unique transactions which were not executed in some previous pbft block
Reward votes for period N-1
When a pbft block is cert voted on and pbft block is pushed into the chain, cert votes for this latest pbft block in the chain will be kept/saved in separate db column last_block_cert_votes and not saved within period_data as we do now. This will be in db so it will be persistent. This db column will only contain cert votes for the last block in the chain. This will be a small size column which should make read/writes to it fast.
When proposing a new pbft block, this table will be the source for selecting reward votes.

Syncing:
In our current implementation SyncBlock contains 2t+1 votes necessary to validate the block. In this new approach when syncing there will be a need to use rewards votes from block N+1 to validate block N. We already have a queue of blocks waiting for verification so this should be simple to implement. When syncing to the final block in the chain we already have a flag indicating it is the last block. In the new approach with the last block that is syncing, cert votes should be sent as well that are retrieved from last_block_cert_votes column in the db so that the last block can be validated.

With this approach writing to period_data will be final with no updating later needed. All the nodes will have exactly same data in period_data. Difference between the nodes will only be in the last_block_cert_votes

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1561,9 +1561,10 @@ bool PbftManager::pushCertVotedPbftBlockIntoChain_(taraxa::blk_hash_t const &cer
     return false;
   }
 
-  cert_sync_block_.cert_votes = std::move(cert_votes_for_round);
+  // TODO: Only include reward votes that are part of the pbft block and not all cert votes
+  cert_sync_block_.cert_votes = db_->getLastBlockCertVotes();
```
MatusKysel, line 1565: do we need to get them from DB? isn't it better to get them from some memory and this should be used on restart?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1636,7 +1637,8 @@ void PbftManager::finalize_(SyncBlock &&sync_block, std::vector<h256> &&finalize
   }
 }
 
-bool PbftManager::pushPbftBlock_(SyncBlock &&sync_block, vec_blk_t &&dag_blocks_order) {
+bool PbftManager::pushPbftBlock_(SyncBlock &&sync_block, std::vector<std::shared_ptr<Vote>> &&cert_votes,
```
kstdl, line 1640: Maybe we can create a new class inherited from `SyncBlock` and it will also include this cert_votes. I think it will be cleaner and easier to understand than this additional argument in many methods. 
Also maybe we should consider renaming `SyncBlock` to `FullPbftBlockPeriodData` or something simplier, but more accurate than current name

libraries/core_libs/consensus/src/pbft/sync_queue.cpp
```diff
@@ -13,7 +13,12 @@ uint64_t SyncBlockQueue::getPeriod() const {
 
 size_t SyncBlockQueue::size() const {
   std::shared_lock lock(queue_access_);
-  return queue_.size();
+  // Each sync block requires cert votes from the next block or from the last_block_cert_votes_ if it is the last block,
+  // a block is counted only if it has cert votes as well
+  if (last_block_cert_votes_.size() || queue_.size() == 0)
+    return queue_.size();
+  else
+    return queue_.size() - 1;
```
MatusKysel, line 21: I do not understand when this can happen? aren't we pushing block with votes all the time?

libraries/core_libs/consensus/src/pbft/sync_queue.cpp
```diff
@@ -22,7 +27,8 @@ void SyncBlockQueue::clear() {
   queue_.clear();
```
MatusKysel, line 27: you should probably clear votes?

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -79,10 +79,22 @@ void GetPbftSyncPacketHandler::sendPbftBlocks(dev::p2p::NodeID const &peer_id, s
...
-    s.appendList(3);
-    s << (pbft_chain_synced && last_block);
-    s << last_block;
-    s.appendRaw(data);
+    if (pbft_chain_synced && last_block) {
+      s.appendList(4);
+      s << (pbft_chain_synced && last_block);
+      s << last_block;
+      s.appendRaw(data);
+      const auto votes = db_->getLastBlockCertVotes();
```
MatusKysel, line 87: again, this could be part of pbft managers memory

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -57,6 +60,13 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
     throw MaliciousPeerException("Unable to parse SyncBlock: " + std::string(e.what()));
   }
 
+  std::vector<std::shared_ptr<Vote>> cert_votes;
+  if (pbft_chain_synced) {
+    const auto cert_votes_count = packet_data.rlp_[3].itemCount();
```
MatusKysel, line 65: reserve size based on this value

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -99,7 +109,7 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
   }
 
   // Check cert vote matches
-  for (auto const &vote : sync_block.cert_votes) {
+  for (auto const &vote : cert_votes) {
```
MatusKysel, line 112: so we are checking this only for the latest block, is that ok?

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -872,6 +872,34 @@ void DbStorage::addNextVotesToBatch(uint64_t pbft_round, std::vector<std::shared
   insert(write_batch, Columns::next_votes, toSlice(pbft_round), toSlice(s.out()));
 }
 
+void DbStorage::saveLastBlockCertVotes(std::vector<std::shared_ptr<Vote>> const& cert_votes) {
+  dev::RLPStream s(cert_votes.size());
+  for (auto const& v : cert_votes) {
+    insert(Columns::next_votes, toSlice(v->getHash()), toSlice(v->rlp(true, true)));
```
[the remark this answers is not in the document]
mfrankovi: no, but actually this method is not used at all so I will remove it

libraries/core_libs/consensus/src/pbft/sync_queue.cpp
mfrankovi: yes

MatusKysel (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: There is definitely this naming mismatch of column period_data and SyncBlock. The two should have some common name. I am not sure what would be the best name for both.

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/pbft/sync_queue.cpp
mfrankovi: Fixed

libraries/core_libs/storage/src/storage.cpp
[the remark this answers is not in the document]
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -57,6 +60,14 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
     throw MaliciousPeerException("Unable to parse SyncBlock: " + std::string(e.what()));
   }
 
+  std::vector<std::shared_ptr<Vote>> cert_votes;
+  if (pbft_chain_synced) {
```
JakubFornadel, line 64: isnt `const bool pbft_chain_synced = packet_data.rlp_[0].toInt<bool>();` redundant ? You can say if the packet is the last one based on item count now ?

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
JakubFornadel: What does it mean we did not get it yet ? Yo can (and probably should ? ) perform this check for both  sync_block.cert_votes as well as cert_votes

libraries/core_libs/consensus/include/pbft/sync_queue.hpp
```diff
@@ -26,15 +26,17 @@ class SyncBlockQueue {
...
-  bool push(SyncBlock &&block, dev::p2p::NodeID const &node_id, uint64_t max_pbft_size);
+  bool push(SyncBlock &&block, dev::p2p::NodeID const &node_id, uint64_t max_pbft_size,
+            std::vector<std::shared_ptr<Vote>> &&cert_votes);
 
   /**
    * @brief Pop the first block from syncing queue
    * @return the first block and peer node ID
    */
-  std::pair<SyncBlock, dev::p2p::NodeID> pop();
+  std::tuple<SyncBlock, std::vector<std::shared_ptr<Vote>>, dev::p2p::NodeID> pop();
```
JakubFornadel, line 39: adjust return comment

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1561,9 +1561,10 @@ bool PbftManager::pushCertVotedPbftBlockIntoChain_(taraxa::blk_hash_t const &cer
     return false;
   }
 
-  cert_sync_block_.cert_votes = std::move(cert_votes_for_round);
+  // TODO: Only include reward votes that are part of the pbft block and not all cert votes
+  cert_sync_block_.cert_votes = db_->getLastBlockCertVotes();
 
-  if (!pushPbftBlock_(std::move(cert_sync_block_), std::move(dag_blocks_order))) {
+  if (!pushPbftBlock_(std::move(cert_sync_block_), std::move(cert_votes_for_round), std::move(dag_blocks_order))) {
```
JakubFornadel, line 1567: current_round_cert_votes might be more appropriate ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1912,16 +1915,17 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
                  << " doesn't have enough valid cert votes. Clear synced PBFT blocks! DPOS total votes count: "
                  << getDposTotalVotesCount();
     sync_queue_.clear();
-    net->getSpecificHandler<network::tarcap::PbftSyncPacketHandler>()->handleMaliciousSyncPeer(sync_block.second);
+    net->getSpecificHandler<network::tarcap::PbftSyncPacketHandler>()->handleMaliciousSyncPeer(node_id);
     return std::nullopt;
   }
 
-  return std::optional<SyncBlock>(std::move(sync_block.first));
+  return std::optional<std::pair<SyncBlock, std::vector<std::shared_ptr<Vote>>>>({std::move(sync_block), cert_votes});
```
JakubFornadel, line 1922: std::move(cert_votes) ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1912,16 +1915,17 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
...
 }
 
-void PbftManager::syncBlockQueuePush(SyncBlock &&block, dev::p2p::NodeID const &node_id) {
+void PbftManager::syncBlockQueuePush(SyncBlock &&block, dev::p2p::NodeID const &node_id,
+                                     std::vector<std::shared_ptr<Vote>> &&cert_votes) {
```
JakubFornadel, line 1926: cuurent_block_cert_votes ?

JakubFornadel (changes requested):
If I am not mistaken, function `SyncBlock::hasEnoughValidCertVotes` validates cert_votes against actual dpos dpos_period_, which is +1 compared to the period of cert_votes from sync block ? We should definitely rename it to previous_block_cert_votes

libraries/types/pbft_block/include/pbft/sync_block.hpp
```diff
@@ -29,7 +29,7 @@ class SyncBlock {
   explicit SyncBlock(bytes const& all_rlp);
 
   std::shared_ptr<PbftBlock> pbft_blk;
-  std::vector<std::shared_ptr<Vote>> cert_votes;
+  std::vector<std::shared_ptr<Vote>> cert_votes;  // These votes are the cert votes that validate previous pbft block
```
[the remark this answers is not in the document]
JakubFornadel: I would call it  `previous_block_cert_votes` with comment that it is equal to the current_block_reward_votes

libraries/core_libs/storage/include/storage/storage.hpp
```diff
@@ -111,8 +111,9 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
     COLUMN(pbft_cert_voted_block);
     COLUMN(pbft_head);
     COLUMN(verified_votes);
-    COLUMN(soft_votes);  // only for current PBFT round
-    COLUMN(next_votes);  // only for previous PBFT round
+    COLUMN(soft_votes);             // only for current PBFT round
+    COLUMN(next_votes);             // only for previous PBFT round
+    COLUMN(last_block_cert_votes);  // cert votes for last block in pbft chain
```
JakubFornadel, line 116: isnt this already stored in db ?

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -133,7 +144,7 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
   LOG(log_tr_) << "Synced PBFT block hash " << pbft_blk_hash << " with " << sync_block.cert_votes.size()
                << " cert votes";
   LOG(log_tr_) << "Synced PBFT block " << sync_block;
-  pbft_mgr_->syncBlockQueuePush(std::move(sync_block), packet_data.from_node_id_);
+  pbft_mgr_->syncBlockQueuePush(std::move(sync_block), packet_data.from_node_id_, std::move(cert_votes));
```
gaoqi8649, line 147: @JakubFornadel ask that what happens if cert_votes is empty due to pbft_chain_synced == false ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
gaoqi8649: That will change in my PR, get them from reward votes in memory

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -23,8 +23,11 @@ PbftSyncPacketHandler::PbftSyncPacketHandler(
       network_sync_level_size_(network_sync_level_size) {}
 
 void PbftSyncPacketHandler::validatePacketRlpFormat(const PacketData &packet_data) const {
-  if (constexpr size_t required_size = 3; packet_data.rlp_.itemCount() != required_size) {
-    throw InvalidRlpItemsCountException(packet_data.type_str_, packet_data.rlp_.itemCount(), required_size);
+  constexpr size_t standard_required_size = 3;
+  constexpr size_t last_block_required_size = 4;
+  if (packet_data.rlp_.itemCount() != standard_required_size &&
+      packet_data.rlp_.itemCount() != last_block_required_size) {
+    throw InvalidRlpItemsCountException(packet_data.type_str_, packet_data.rlp_.itemCount(), standard_required_size);
```
gaoqi8649, line 30: The if condition checks itemCount() != 3 && itemCount() != 4, but the throw exception will only show standard_required_size

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
```diff
@@ -79,10 +79,22 @@ void GetPbftSyncPacketHandler::sendPbftBlocks(dev::p2p::NodeID const &peer_id, s
...
+      s.appendList(votes.size());
+      for (const auto &vote : votes) {
+        s.appendRaw(vote->rlp(true));
+      }
+    } else {
+      s.appendList(3);
+      s << (pbft_chain_synced && last_block);
+      s << last_block;
+      s.appendRaw(data);
+    }
```
gaoqi8649, line 97: I think that's dangerous here.

For example:
1. N1 has PBFT chain size 2
2. N2 sync from N1
3. N1 send B1 to N2, then lose connection to N2.
4. N2 will find another peer to sync
5. In the meanwhile, N2 will process B1 from syncing queue
6. N2 failed processing B1 because no cert votes for B1
7. N2 will not receive B1 again, because N2 has marked B1 as known block

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
JakubFornadel: I already know what happens, so you dont need to ask it in my name :) Either you are interested in answer or delete this

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
JakubFornadel: that is ok

libraries/types/pbft_block/include/pbft/sync_block.hpp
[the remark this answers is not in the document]
mfrankovi: Done

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: Done

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
mfrankovi: I meant that we could not verify the current block unless it is the last one before we get the votes that are in the next block. There is a full vote verification when we try to push blocks into chain but I added an additional check in PbftManager::PeriodDataQueuePush so that we do not push in the queue if the votes are invalid

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
mfrankovi: I guess so although it seems cleaner this way because we are not using some magic number for this. It does not add much to the packet size. But yes I could remove it if you think that is better

libraries/core_libs/consensus/include/pbft/sync_queue.hpp
mfrankovi: Done

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: Done

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: Done

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: Done

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -57,6 +60,14 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
     throw MaliciousPeerException("Unable to parse SyncBlock: " + std::string(e.what()));
   }
 
+  std::vector<std::shared_ptr<Vote>> cert_votes;
```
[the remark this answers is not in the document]
mfrankovi: Done

mfrankovi:
> If I am not mistaken, function `SyncBlock::hasEnoughValidCertVotes` validates cert_votes against actual dpos dpos_period_, which is +1 compared to the period of cert_votes from sync block ? We should definitely rename it to previous_block_cert_votes

I am not sure I understand your comment. So this function verified current block with the current block votes that are supplied as parameter of the function. It does not use the internal votes from previous block

libraries/core_libs/storage/include/storage/storage.hpp
mfrankovi: the whole vote handling is kind of complex and probably needs refactoring. I know that the same cert votes I am storing in last_block_cert_votes were probably already stored in one of these *_votes tables but I found it too complex to reuse it for now

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
mfrankovi: B1 will stay in the queue until we get B2, it will not be removed from the queue

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
mfrankovi: I guess other guys can weigh in but I am pretty sure that rocksdb memory caching does exactly the same thing so if db provides memory caching of data I do not see why we need to duplicate it and implement one in our code.

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
JakubFornadel: wait, this field `sync_block.cert_votes` are basically cert_votes for the previous block if I understand it correctly ? So you can definitely verify correctness of these votes by checking if those are the votes for the previous block + you can verify if the cert_votes for the last block are pointing to it

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
```diff
@@ -197,14 +197,16 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
    * @brief Get PBFT blocks syncing queue size
    * @return PBFT syncing queue size
    */
-  size_t syncBlockQueueSize() const;
+  size_t PeriodDataQueueSize() const;
```
JakubFornadel, line 200: capital first letter

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
```diff
@@ -197,14 +197,16 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
...
 
   /**
    * @brief Push synced PBFT block in syncing queue
    * @param block synced PBFT block from peer
+   * @param cuurent_block_cert_votes cert votes for PeriodData pbft block period
    * @param node_id peer node ID
    */
-  void syncBlockQueuePush(SyncBlock &&block, dev::p2p::NodeID const &node_id);
+  void PeriodDataQueuePush(PeriodData &&block, dev::p2p::NodeID const &node_id,
```
JakubFornadel, line 208: capital first letter

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
```diff
@@ -465,27 +467,29 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
...
   /**
    * @brief Push a new PBFT block into the PBFT chain
-   * @param sync_block PBFT block, cert votes, DAG blocks, and transactions
+   * @param sync_block PBFT block, cert votes for previous period, DAG blocks, and transactions
+   * @param cert_votes cert votes for pbft block period
    * @param dag_blocks_order DAG blocks hashes
    * @return true if push a new PBFT block into the PBFT chain
    */
-  bool pushPbftBlock_(SyncBlock &&sync_block, vec_blk_t &&dag_blocks_order = {});
+  bool pushPbftBlock_(PeriodData &&sync_block, std::vector<std::shared_ptr<Vote>> &&cert_votes,
```
JakubFornadel, line 491: refactor also arg name, not just type

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
```diff
@@ -465,27 +467,29 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
...
+                                        std::vector<std::shared_ptr<Vote>> &&current_round_cert_votes);
 
   /**
    * @brief Final chain executes a finalized PBFT block
    * @param sync_block PBFT block, cert votes, DAG blocks, and transactions
    * @param finalized_dag_blk_hashes DAG blocks hashes
    * @param sync it's true when it's last finalized block
    */
-  void finalize_(SyncBlock &&sync_block, std::vector<h256> &&finalized_dag_blk_hashes, bool sync = false);
+  void finalize_(PeriodData &&sync_block, std::vector<h256> &&finalized_dag_blk_hashes, bool sync = false);
```
JakubFornadel, line 482: refactor also arg name, not just type

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
```diff
@@ -587,7 +591,7 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
   blk_hash_t soft_voted_block_for_this_round_ = NULL_BLOCK_HASH;
 
   // Full sync block for pbft block that is being currently cert voted for
-  SyncBlock cert_sync_block_;
+  PeriodData cert_sync_block_;
```
JakubFornadel, line 594: based on the name of the variable and comment above, I am kind of lost what this variable actually holds ?

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
JakubFornadel: the problem I see here is that we are setting 2 independent data fields and it could break processing. Someone could create a bug when he set this flag to true but does not provide the the current block cert votes and in such case we would try to process non-existent data field. I would definitely create `standard_size & last_block_size` as class constants and create flag for this if like this: `is_last_block = itemsCount == last_block_size`

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
mfrankovi: Yes and I am doing this in PbftManager::PeriodDataQueuePush

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
mfrankovi: One thing that we do know that in our process it is just rocksdb caching that is taking most of the RAM. Caching can be configured in rocksdb globally and per column. By introducing our own caches, we are basically doubling the memory usage and we gain no performance. In my opinion we should either:
1. Use rocksdb memory caching especially for small columns unless we can find a specific reason why we would cache data better than rocksdb
2. This is something I would not recommend but turn off rocksdb caching and have our own caching mechanism. Having both on does not make sense to me

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
mfrankovi: Ok, I implemented it this way now

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
mfrankovi: Fixed

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
mfrankovi: Fixed

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
mfrankovi: Fixed

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
mfrankovi: Updated name and comment

libraries/core_libs/consensus/include/pbft/period_data_queue.hpp
```diff
@@ -4,37 +4,39 @@
...
   /**
    * @brief Push a synced block in queue
    * @param block block synced from peer
    * @param node_id peer node ID
    * @param max_pbft_size maximum PBFT chain size
+   * @param cert_votes cert votes
    * @return true if pushed
    */
-  bool push(SyncBlock &&block, dev::p2p::NodeID const &node_id, uint64_t max_pbft_size);
+  bool push(PeriodData &&block, dev::p2p::NodeID const &node_id, uint64_t max_pbft_size,
```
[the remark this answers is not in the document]
mfrankovi: Fixed

libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
mfrankovi: not sure which thing exactly you are referring to here?

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
mfrankovi: actually still working on this, my initial change was not good

mfrankovi: Fixed

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1912,26 +1915,43 @@ std::optional<SyncBlock> PbftManager::processSyncBlock() {
...
+
+  auto previous_block_hash = sync_queue_.lastBlockHash();
+  if (previous_block_hash.has_value()) {
+    // Check cert vote matches
+    for (auto const &vote : period_data.previous_block_cert_votes) {
+      if (vote->getBlockHash() != *previous_block_hash) {
+        LOG(log_er_) << "Invalid cert votes block hash " << vote->getBlockHash() << " instead of "
+                     << *previous_block_hash << " from peer " << node_id.abridged() << " received, stop syncing.";
+        auto net = network_.lock();
+        if (net) net->getSpecificHandler<network::tarcap::PbftSyncPacketHandler>()->handleMaliciousSyncPeer(node_id);
```
JakubFornadel, line 1938: arent you missing return here ?

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
JakubFornadel: @MatusKysel I leave it to you as it is your orig. comment.

MatusKysel: If this won't be fixed in this PR let's create task for it, otherwise we will forget  and we could gain some performance here

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -872,6 +872,27 @@ void DbStorage::addNextVotesToBatch(uint64_t pbft_round, std::vector<std::shared
   insert(write_batch, Columns::next_votes, toSlice(pbft_round), toSlice(s.out()));
 }
 
+void DbStorage::addLastBlockCertVotesToBatch(std::vector<std::shared_ptr<Vote>> const& cert_votes, Batch& write_batch) {
+  auto it = std::unique_ptr<rocksdb::Iterator>(db_->NewIterator(read_options_, handle(Columns::last_block_cert_votes)));
+  for (it->SeekToFirst(); it->Valid(); it->Next()) {
+    remove(write_batch, Columns::last_block_cert_votes, it->key());
+  }
```
MatusKysel, line 879: isn't there other (faster) way to clean whole column?

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
mfrankovi: Db usually also have write-through caching. When you write data to rocksdb there is actually a setting where you can choose if you wait for the call to return even if the data is not written to the disk yet. It is written to memory cache on write as well.

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -91,49 +99,67 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
...
+                     << " from peer " << packet_data.from_node_id_.abridged() << " received, stop syncing.";
+        handleMaliciousSyncPeer(packet_data.from_node_id_);
+        return;
+      }
+    }
+  }
+
+  // Check votes match the hash of previous block in the queue
+  auto last_pbft_block = pbft_mgr_->periodDataQueueLastPbftBlock();
+  if (last_pbft_block) {
```
JakubFornadel, line 123: this is not going to validate the hashes in many cases, which is not ok. The thing is, we have to have a way to validate these votes hashes every single time we receive sync data. Anytime you start syncing and get first block, we will skip validation due to this condition, anytime the processing thread that is popping data from sync queue faster than we are pushing it there (as a result queue is almost always empty) we will skip validation... We have to save somehow somewhere last processed pbft_block hash and this validation must be run every single time we receive sync data.

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
MatusKysel: I know, I in past I have made some test for trxs/dags and this cache is still like 10x slower then access to memory

mfrankovi: So I did a quick test. I save 1000 votes in the column. And then I test 3 scenarios:
1. Reading all 1000 votes from database which includes rlp deserialization and this took 14139µs
2. Reading 1000 votes from local memory but also from vector or serialized rlp data and this took 13684µs
3. Simple copy of 1000 votes from memory took 250µs

Based on this I would say that reading from rocksdb memory cache is as fast as reading from local memory but the main bottleneck is rlp serialization. 
Based on this it would be much faster to keep local copy of votes so I can implement that but if we are going for some great performance maybe we should consider if storing data as rlp is efficient and consider some other faster serialization format

libraries/core_libs/storage/include/storage/storage.hpp
mfrankovi: I think that as for just storing votes it could be ok as is, I think having only cert votes for the previous block in a separate column so that they can easily be fetched in a batch when required by syncing or finalizing next block is better than mixing these cert votes with cert votes for the following block and then having to filter them out when pulling them. This votes move from one column to the other is quite fast and it is only done once on block finalization. I mean it is possible that some other design could be more efficient and whole vote handling needs a refactor but I am not sure how to put that in a simple comment

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
mfrankovi: Fixed by also going to chain last pbft block if queue is empty

libraries/core_libs/network/src/tarcap/packets_handlers/get_pbft_sync_packet_handler.cpp
JakubFornadel: I created issue to check other serialization formats https://github.com/Taraxa-project/taraxa-node/issues/1799. If you feel like you resolved the other part of the comment(having these votes in memory), I think you can resolve this and merge without as he is not working today and qi is waiting for this PR

mfrankovi: @MatusKysel I will merge and if you feel this is still unresolved create an issue with the details on how it should be improved

PR #1746, opened 2022-05-24
chore: switch to new dpos2.0

### LINKED ISSUE #1743: switch to new DPOS 2.0 contract

MatusKysel:
fixes #1743

### DIFF
--- a/tests/pbft_manager_test.cpp
+++ b/tests/pbft_manager_test.cpp
@@ -375,7 +375,7 @@ TEST_F(PbftManagerTest, full_node_lambda_input_test) {
   EXPECT_EQ(pbft_mgr->getPbftInitialLambda(), 2000);
 }
 
-TEST_F(PbftManagerTest, check_get_eligible_vote_count) {
+TEST_F(PbftManagerTest, DISABLED_check_get_eligible_vote_count) {
   auto node_cfgs = make_node_cfgs<5>(5);
   auto node_1_expected_bal = own_effective_genesis_bal(node_cfgs[0]);
   for (auto &cfg : node_cfgs) {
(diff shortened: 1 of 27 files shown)


tests/pbft_manager_test.cpp
```diff
@@ -46,24 +46,24 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
+  //   nodes[0]->getTransactionManager()->insertTransaction(trx);
+  //   trxs_count++;
+  //   EXPECT_HAPPENS({120s, 1s}, [&](auto &ctx) {
+  //     for (auto &node : nodes) {
+  //       if (ctx.fail_if(!node->getFinalChain()->transaction_location(trx->getHash()))) {
+  //         return;
+  //       }
+  //     }
+  //   });
+  // }
```
gaoqi8649, line 66: The part of code is for delegate different amount of stake to nodes. I don't understand why comment out

tests/pbft_manager_test.cpp
```diff
@@ -391,31 +391,31 @@ TEST_F(PbftManagerTest, check_get_eligible_vote_count) {
...
+
+  //   trxs_count++;
+  //   EXPECT_HAPPENS({120s, 1s}, [&](auto &ctx) {
+  //     for (auto &node : nodes) {
+  //       if (ctx.fail_if(!node->getFinalChain()->transaction_location(trx->getHash()))) {
+  //         return;
+  //       }
+  //     }
+  //   });
+  // }
```
gaoqi8649, line 418: Same here

gaoqi8649 (commented):
I think we should keep hardfork functionality. we may use it later

tests/pbft_manager_test.cpp
MatusKysel: we do not have same kind of transactions to delegate, it's commented out as this needs to be reworked

MatusKysel:
> I think we should keep hardfork functionality. we may use it later

I did not removed it just disabled for now as it does not have support in new contract

JakubFornadel (approved):

PR #1748, opened 2022-05-24
Change source of tested helm chart

rjonczy:
## Purpose

Main purpose of this PR is:
- refactor build process

## Changes made

- changed tested helm chart
- changed default resources to be empty object `{}`
- changed node pool where chart is tested
- changed upload of helm chart artifacts to repo


MatusKysel (approved):

sfrique (approved):
I guess my only concern is that we don't have any limits yet, so these tests can consume a lot of resources and bring the node pool down.
The "good" thing is that the tests should run for a limited time right?

But if we want to play safe, I would add a `pr-tests.values.yaml`  and use it in the helm test.

rjonczy:
@sfrique added limits for 3 type of nodes and merging.

PR #1754, opened 2022-05-25
Reward stats

JakubFornadel:


justinsnapp (approved):

libraries/config/include/config/state_api_config.hpp
```diff
@@ -42,9 +42,12 @@ Json::Value enc_json(DPOSConfig const& obj);
...
 struct ExecutionOptions {
-  bool disable_nonce_check = 0;
-  bool disable_gas_fee = 0;
-  bool enable_nonce_skipping = 0;
+  bool disable_nonce_check = false;
+  bool disable_gas_fee = false;
+  bool enable_nonce_skipping = false;
+  // Do not process reward pool (newly minted tokens & tx fees) according to the reward statistics,
+  // give whole reward pool to the pbft block proposer
+  bool disable_stats_rewards = false;
```
MatusKysel, line 50: hmm maybe irrelevant comment but maybe this param should be part of same config as `disable_block_rewards`

libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -89,23 +90,48 @@ class FinalChainImpl final : public FinalChain {
...
+    SharedTransactions txs_to_execute;
+    txs_to_execute.reserve(new_blk.transactions.size());
+
+    // Dag blocks validators that included transactions to be executed as first in their blocks
+    std::vector<addr_t> txs_to_execute_validators;
+    txs_to_execute_validators.reserve(new_blk.transactions.size());
+
+    // Creates rewards stats
+    RewardsStats rewards_stats;
+    rewards_stats.processStats(new_blk);
```
MatusKysel, line 104: maybe I am wrong but `txs_to_execute_validators` could be returned by this function -> `processStats` and txs_to_execute we will not need as we can use `new_blk.transactions`

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
```diff
@@ -0,0 +1,75 @@
+#include "final_chain/rewards_stats.hpp"
+
+namespace taraxa {
+
+bool RewardsStats::addTransaction(const trx_hash_t& tx_hash, const addr_t& validator) {
+  auto found_tx = txs_validators_.find(tx_hash);
+
+  // Already processed tx
+  if (found_tx != txs_validators_.end()) {
+    return false;
```
MatusKysel, line 10: maybe there should be assert as all trxs needs to be unique in syncblock, or not here but in processblock should be assert on return value of this function

MatusKysel (approved):

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
JakubFornadel: ah this must be without assert, it is actually processing also duplicate txs(from dag blocks -> txs hashes):
```
for (const auto& dag_block : sync_blk.dag_blocks) {
    const addr_t& dag_block_author = dag_block.getSender();
    }
  }
```

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+  auto found_validator = validators_stats_.find(found_tx->second);
+  assert(found_validator != validators_stats_.end());
+
+  assert(found_validator->second.unique_txs_count_);
+
+  found_validator->second.unique_txs_count_--;
+  total_unique_txs_count_--;
+}
+
```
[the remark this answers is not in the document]
JakubFornadel: we will need it if we re-nable protection_service in final chain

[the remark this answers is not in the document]
MatusKysel: ok, i see

PR #1775, opened 2022-06-02
chore/improve smoke tests

rjonczy:
## Purpose

When container with `taraxa-node` is crashing during smoke tests, it is hard to understand reason of failure. 

Added logs to better understand why container is crashing.

## Changes made

- removed `--rm` when running container with node during smoke tests 
- added logs to get insights why container crashed


MatusKysel (approved):

kstdl (approved):

.circleci/config.yml
```diff
@@ -532,39 +532,49 @@ jobs:
               docker network rm \
               smoketest-${DOCKER_BRANCH_TAG:0:52} >/dev/null;
             fi
+            
             docker network create --driver=bridge \
-                        smoketest-${DOCKER_BRANCH_TAG:0:52}
-            docker run --rm -d \
```
sfrique, line 537: We should remove the container at some point to avoid the container not being able to execute because there is one with the same name.

We could remove before or after then run..  or maybe add the build number to the name, but there are a lot of places we should update this.

rjonczy: @sfrique there is a separate task for that:


```
      - run:
          name: Cleanup smoke tests
          when: always
          command: |
            docker kill smoketest-${DOCKER_BRANCH_TAG:0:52} || true
            docker network rm smoketest-${DOCKER_BRANCH_TAG:0:52} || true
```

rjonczy: ok, just seen now that it is killing but not removing. OK will remove to that task

rjonczy: @sfrique hmmm, however i am not sure about that. Because every build new / fresh vm is created. There shouldn't be any container in new build, no ?

sfrique: well, this was added because we had errors.
so I am not sure, I just recall having a container with the same name error.

sfrique: it should be fine to remove after the kill

rjonczy: added cleaning container to task with cleaning smoke tests

PR #1782, opened 2022-06-03
feat: new genesis structure for initial validators

kstdl:
https://github.com/Taraxa-project/taraxa-evm/pull/71

### DIFF
--- a/libraries/cli/include/cli/devnet_config.hpp
+++ b/libraries/cli/include/cli/devnet_config.hpp
@@ -196,7 +196,7 @@ constexpr std::string_view devnet_json = R"foo({
               "owner": "0x780fe8b2226cf212c55635de399ee4c2a860810c",
               "commission": 0,
               "endpoint": "",
-              "description": "1 initial validator",
+              "description": "Taraxa devnet validator 1",
               "delegations": {
                 "0x7e4aa664f71de4e9d0b4a6473d796372639bdcde": "0x84595161401484a000000"
               }
@@ -206,7 +206,7 @@ constexpr std::string_view devnet_json = R"foo({
               "owner": "0x56e0de6933d9d0453d0363caf42b136eb5854e4e",
               "commission": 0,
               "endpoint": "",
-              "description": "2 initial validator",
+              "description": "Taraxa devnet validator 2",
               "delegations": {
                 "0x7e4aa664f71de4e9d0b4a6473d796372639bdcde": "0x84595161401484a000000"
               }
@@ -216,7 +216,7 @@ constexpr std::string_view devnet_json = R"foo({
               "owner": "0x71bdcbec7e3642782447b0fbf31eed068dfbdbb1",
               "commission": 0,
               "endpoint": "",
-              "description": "3 initial validator",
+              "description": "Taraxa devnet validator 3",
               "delegations": {
                 "0x7e4aa664f71de4e9d0b4a6473d796372639bdcde": "0x84595161401484a000000"
               }
@@ -226,7 +226,7 @@ constexpr std::string_view devnet_json = R"foo({
               "owner": "0xac24bc60a491bd0c29414e5f34aa6bbd8d4aa499",
               "commission": 0,
               "endpoint": "",
-              "description": "5 initial validator",
+              "description": "Taraxa devnet validator 5",
               "delegations": {
                 "0x7e4aa664f71de4e9d0b4a6473d796372639bdcde": "0x84595161401484a000000"
               }
@@ -236,7 +236,7 @@ constexpr std::string_view devnet_json = R"foo({
               "owner": "0x635d3831ad5d1252a2a07f09b8d3539b3af34df8",
               "commission": 0,
               "endpoint": "",
-              "description": "6 initial validator",
+              "description": "Taraxa devnet validator 6",
               "delegations": {
                 "0x7e4aa664f71de4e9d0b4a6473d796372639bdcde": "0x84595161401484a000000"
               }
@@ -246,7 +246,7 @@ constexpr std::string_view devnet_json = R"foo({
               "owner": "0x43af71034ed7fd0b54496a30ba4a5889a94e7088",
               "commission": 0,
               "endpoint": "",
-              "description": "7 initial validator",
+              "description": "Taraxa devnet validator 7",
               "delegations": {
                 "0x7e4aa664f71de4e9d0b4a6473d796372639bdcde": "0x84595161401484a000000"
               }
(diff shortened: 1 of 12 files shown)


MatusKysel (approved):
I think after this change we need to adjust a lot of scripts, so maybe @rattrap and @rjonczy should also look at it. I would also add json example to description, so they know how it is changing.

MatusKysel:
one more thing, local-net will not work after this change I guess

tests/final_chain_test.cpp
```diff
@@ -277,6 +278,36 @@ TEST_F(FinalChainTest, nonce_skipping) {
...
+  const dev::KeyPair key = dev::KeyPair::create();
+  const std::vector<dev::KeyPair> validator_keys = {dev::KeyPair::create(), dev::KeyPair::create(),
+                                                    dev::KeyPair::create()};
+  cfg.state.genesis_balances = {};
+  cfg.state.dpos.emplace();
+  cfg.state.dpos->eligibility_balance_threshold = 100000;
+  cfg.state.dpos->vote_eligibility_balance_step = 10000;
+  cfg.state.dpos->maximum_stake = 10000000;
+  cfg.state.dpos->minimum_deposit = 1000;
+  cfg.state.dpos->eligibility_balance_threshold = 1000;
```
JakubFornadel, line 291: lets put everywhere real values as we had problems with dpos tests when they passed just because the data we set there were nonsense...
Every value here should be multiplied by 10^18

tests/final_chain_test.cpp
```diff
@@ -277,6 +278,36 @@ TEST_F(FinalChainTest, nonce_skipping) {
...
+
+  for (const auto& vk : validator_keys) {
+    state_api::ValidatorInfo validator{vk.address(), key.address(), 0, "", "", {}};
+    validator.delegations.emplace(key.address(), cfg.state.dpos->maximum_stake);
+    validator.delegations.emplace(validator_keys.front().address(), cfg.state.dpos->maximum_stake);
```
JakubFornadel, line 296: For the same validator, you are creating 2 delegators with  maximum_stake delegation, right ? If that is the case, the second delegation should not get through because maximum_stake is max. total stake for validator...

tests/final_chain_test.cpp
```diff
@@ -277,6 +278,36 @@ TEST_F(FinalChainTest, nonce_skipping) {
...
+    cfg.state.dpos->initial_validators.emplace_back(validator);
+  }
+
+  cfg.state.genesis_balances[key.address()] = 1000000000;
+  cfg.state.genesis_balances[validator_keys.front().address()] = 1000000000;
+  init();
+  auto total_votes = SUT->dpos_eligible_total_vote_count(SUT->last_block_number());
+  for (const auto& vk : validator_keys) {
+    auto address_votes = SUT->dpos_eligible_vote_count(SUT->last_block_number(), vk.address());
+    EXPECT_EQ(total_votes / validator_keys.size(), address_votes);
```
JakubFornadel, line 306: probably not a good idea to use division in uint result comparison. Let's say someone changes the code above so 
```
total_votes  = 5
validator_keys = 3 
and the divisionr esult = 1
```

tests/final_chain_test.cpp
```diff
@@ -277,6 +278,36 @@ TEST_F(FinalChainTest, nonce_skipping) {
...
+    EXPECT_EQ(2 * cfg.state.dpos->maximum_stake / cfg.state.dpos->vote_eligibility_balance_step, address_votes);
```
JakubFornadel, line 307: same here ?

tests/final_chain_test.cpp
JakubFornadel: maybe we should rename maximum_stake to something like `validator_max_total_stake` so it is not so confusing ?

tests/util_test/util.hpp
```diff
@@ -313,8 +319,22 @@ inline auto own_balance(std::shared_ptr<FullNode> const& node) {
...
-  return cfg.chain.final_chain.state.effective_genesis_balance(dev::toAddress(dev::Secret(cfg.node_secret)));
+inline state_api::BalanceMap effective_genesis_balances(const state_api::Config& cfg) {
+  if (!cfg.dpos) {
+    return cfg.genesis_balances;
+  }
+
+  state_api::BalanceMap effective_balances = cfg.genesis_balances;
+  for (const auto& validator_info : cfg.dpos->initial_validators) {
+    for (const auto& [delegator, amount] : validator_info.delegations) {
+      effective_balances[delegator] -= amount;
```
JakubFornadel, line 330: so what happens if validator himself is not among his delegators ? Is this scenario somehow covered ? Because in dpos contract we require self.delegation when registering new validator

kstdl: There should be no issue if we will not self-delegate. We are skipping this checks in initial delegation creation

libraries/cli/include/cli/devnet_config.hpp
```diff
@@ -185,34 +185,203 @@ constexpr std::string_view devnet_json = R"foo({
...
-              "0x9588f6457d67792f141424983a93978f44331054": "0x84595161401484a000000",
-              "0xb48da366e19f141f5647dcdb0960eb88719e1c8d": "0x84595161401484a000000",
-              "0x2b95aeed70a4e21fcb8665b1a1e16eb114a1caf3": "0x84595161401484a000000"
+           "initial_validators": [
+            {
+              "address": "0x780fe8b2226cf212c55635de399ee4c2a860810c",
+              "owner": "0x780fe8b2226cf212c55635de399ee4c2a860810c",
+              "commission": 0,
+              "endpoint": "",
+              "description": "1 initial validator",
```
JakubFornadel, line 199: I would call it somehow more official, e.g. Taraxa validator 1

mfrankovi (approved):

libraries/cli/include/cli/devnet_config.hpp
```diff
@@ -185,34 +185,203 @@ constexpr std::string_view devnet_json = R"foo({
...
+              "delegations": {
+                "0x7e4aa664f71de4e9d0b4a6473d796372639bdcde": "0x84595161401484a000000"
+              }
+            },
+            {
+              "address": "0x56e0de6933d9d0453d0363caf42b136eb5854e4e",
+              "owner": "0x56e0de6933d9d0453d0363caf42b136eb5854e4e",
+              "commission": 0,
+              "endpoint": "",
+              "description": Taraxa devnet validator "2",
```
MatusKysel, line 209: missing ""

kstdl: Nice catch, thanks. Multiple replacement gone wrong

PR #1793, opened 2022-06-07
Remove gas flag

### LINKED ISSUE #1620: Remove disable_gas_fee flag
We should remove `disable_gas_fee` as it causing a lot of problems, because it set to different settings around the project

MatusKysel:
## Purpose
fixes #1620

### DIFF
--- a/tests/full_node_test.cpp
+++ b/tests/full_node_test.cpp
@@ -1184,13 +1184,9 @@ TEST_F(FullNodeTest, receive_send_transaction) {
 TEST_F(FullNodeTest, detect_overlap_transactions) {
   auto node_cfgs = make_node_cfgs<20>(5);
   auto nodes = launch_nodes(node_cfgs);
-  std::unordered_map<addr_t, val_t> expected_balances;
+  const auto expected_balances = effective_genesis_balances(node_cfgs[0].chain.final_chain.state);
   const auto node_1_genesis_bal = own_effective_genesis_bal(node_cfgs[0]);
-  for (const auto &node : nodes) {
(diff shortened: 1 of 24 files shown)


tests/full_node_test.cpp
```diff
@@ -1183,12 +1183,18 @@ TEST_F(FullNodeTest, receive_send_transaction) {
...
   auto node_cfgs = make_node_cfgs<20>(5);
-  auto node_1_genesis_bal = own_effective_genesis_bal(node_cfgs[0]);
   auto nodes = launch_nodes(node_cfgs);
+  std::unordered_map<addr_t, val_t> expected_balances;
+  const auto node_1_genesis_bal = own_effective_genesis_bal(node_cfgs[0]);
+  for (const auto &node : nodes) {
+    expected_balances[node->getAddress()] = own_effective_genesis_bal(node->getConfig());
+  }
+
+  expected_balances[nodes[0]->getAddress()] = own_effective_genesis_bal(nodes[0]->getConfig());
```
kstdl, line 1193: Maybe it could be replaced with just 
`auto expected_balances = effective_genesis_balances(node_cfgs[0])`

kstdl: But configs for all nodes should be the same. Other way it won't work. 
`effective_genesis_balances` is other method used by `own_effective_genesis_bal` and it is returning the same unordered_map type

MatusKysel: ooo, I see we have function that will do the same :D sorry

kstdl (approved):

PR #1805, opened 2022-06-15
clear_period_data test fix

mfrankovi:
This test was intermittently failing because of very small values of max_levels_per_period and light_node_history which were set to these low levels so that the test can be performed quickly but it could cause one of these issues:
1. One node would get ahead the second one with dag blocks and would send ahead/future block which would cause the disconnect
2. One node would get ahead the second one with pbft chain and light node would delete old blocks so the other nodes could not sync from the light node

Fix is only sending a new transaction to a node if both nodes have same pbft chain size. Setting as light node, node that is not eligible to propose dag blocks not to allow the proposer to create dag blocks faster than other blocks process them.

Besides the test fix, I have also changed the log level for these two cases so that we always pick up when a node disconnect for one of these two reasons.

### DIFF
--- a/tests/full_node_test.cpp
+++ b/tests/full_node_test.cpp
@@ -1608,22 +1608,25 @@ TEST_F(FullNodeTest, light_node) {
   EXPECT_LE(non_empty_counter, node_cfgs[0].light_node_history + 1);
 }
 
-TEST_F(FullNodeTest, DISABLED_clear_period_data) {
+TEST_F(FullNodeTest, clear_period_data) {
   auto node_cfgs = make_node_cfgs<10>(2);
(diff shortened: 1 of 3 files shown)


MatusKysel (approved):

kstdl (approved):
After rebase make sure that `clear_period_data` is enabled

tests/full_node_test.cpp
```diff
@@ -1610,20 +1610,23 @@ TEST_F(FullNodeTest, light_node) {
 
 TEST_F(FullNodeTest, DISABLED_clear_period_data) {
```
kstdl, line 1611: remove `DISABLED_`

PR #1807, opened 2022-06-15
feat: configs are now json files which format are verified in cmake

kstdl:


.gitignore
```diff
@@ -20,9 +20,12 @@ local
 clang-format
 clang-tidy
 
-# header with version of taraxad (it generated by cmake)
+# generated header with version of taraxad (it generated by cmake)
```
JakubFornadel, line 23: fix `it generated by cmake`

libraries/core_libs/node/src/node.cpp
```diff
@@ -84,23 +84,29 @@ void FullNode::init() {
   }
   LOG(log_nf_) << "DB initialized ...";
 
+  std::cout << "conf_.chain.genesisHash() 2: " << conf_.chain.genesisHash() << std::endl;
+  std::cout << enc_json(conf_.chain.final_chain.state) << std::endl;
```
JakubFornadel, line 88: std cout ?

JakubFornadel (approved):

libraries/core_libs/node/src/node.cpp
kstdl: I think I mentioned yesterday that I have some issues with tests, so this is here because of that. 
P.S. it is still a draft :)

libraries/cli/CMakeLists.txt
```diff
@@ -19,6 +15,44 @@ target_include_directories(cli PUBLIC ${CMAKE_CURRENT_SOURCE_DIR}/include)
...
+    list(GET config_files ${i} config_file)
+    file(READ ${config_file} ${config_name})
+
+    set(config_str ${${config_name}})
+
+    string(JSON result ERROR_VARIABLE json_error TYPE ${config_str} chain_config)
+    if(NOT(${result} STREQUAL "OBJECT"))
+        message(FATAL_ERROR "Can't parse ${config_name}: ${json_error}")
+    endif()
+endforeach()
```
MatusKysel, line 56: is this where we validate json files? I would add some comment here what is this code doing

MatusKysel (approved):

PR #1808, opened 2022-06-15
feat: use exist method to check if object is in the DB

### LINKED ISSUE #1738: RocksDB keyMayExist
As we are often search for existence of key in our DB by multi-get and the result has more negative outputs then positive (key does NOT exist).  We should measure if the usage of 
```
  // If the key definitely does not exist in the database, then this method
  // returns false, else true. If the caller wants to obtain value when the key
  // is found in memory, a bool for 'value_found' must be passed. 'value_found'
  // will be true on return if value has been set properly.
  // This check is potentially lighter-weight than invoking DB::Get(). One way
  // to make this lighter weight is to avoid doing any IOs.
  // Default implementation here returns true and sets 'value_found' to false
  virtual bool KeyMayExist(const ReadOptions& /*options*/,
                           ColumnFamilyHandle* /*column_family*/,
                           const Slice& /*key*/, std::string* /*value*/,
                           std::string* /*timestamp*/,
                           bool* value_found = nullptr) {
```
is not much better as it not getting any data

MatusKysel:
## Purpose
Fixes #1738

### DIFF
--- a/libraries/core_libs/storage/include/storage/storage.hpp
+++ b/libraries/core_libs/storage/include/storage/storage.hpp
@@ -418,40 +418,6 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
   }
 
   void forEach(Column const& col, OnEntry const& f);
-
-  //   class MultiGetQuery {
-  //     std::shared_ptr<DbStorage> const db_;
-  //     std::vector<rocksdb::ColumnFamilyHandle*> cfs_;
-  //     std::vector<Slice> keys_;
-  //     std::vector<string> str_pool_;
-
-  //    public:
-  //     explicit MultiGetQuery(std::shared_ptr<DbStorage> const& db, uint capacity = 0);
-
-  //     template <typename T>
-  //     MultiGetQuery& append(Column const& col, std::vector<T> const& keys, bool copy_key = true) {
-  //       auto h = db_->handle(col);
-  //       for (auto const& k : keys) {
-  //         cfs_.emplace_back(h);
-  //         if (copy_key) {
-  //           auto const& slice = toSlice(k);
-  //           keys_.emplace_back(toSlice(str_pool_.emplace_back(slice.data(), slice.size())));
-  //         } else {
-  //           keys_.emplace_back(toSlice(k));
-  //         }
-  //       }
-  //       return *this;
-  //     }
-
-  //     template <typename T>
-  //     MultiGetQuery& append(Column const& col, T const& key, bool copy_key = true) {
-  //       return append(col, std::vector{toSlice(key)}, copy_key);
-  //     }
-
-  //     uint size();
-  //     std::vector<string> execute(bool and_reset = true);
-  //     MultiGetQuery& reset();
-  //   };
 };
 
 using DB = DbStorage;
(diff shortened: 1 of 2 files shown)


libraries/core_libs/storage/src/storage.cpp
```diff
@@ -495,17 +495,12 @@ std::optional<std::pair<uint32_t, uint32_t>> DbStorage::getTransactionPeriod(trx
...
-
-  DbStorage::MultiGetQuery db_query(shared_from_this(), trx_hashes.size());
-  db_query.append(DbStorage::Columns::trx_period, trx_hashes);
-  auto db_trxs_period = db_query.execute();
-  for (size_t idx = 0; idx < db_trxs_period.size(); idx++) {
-    auto& trx_raw_period = db_trxs_period[idx];
-    result.push_back(!trx_raw_period.empty());
+  std::vector<bool> result(trx_hashes.size(), false);
+  for (size_t i = 0; i < trx_hashes.size(); ++i) {
+    if (exist(toSlice(trx_hashes[i].asBytes()), Columns::trx_period)) {
```
mfrankovi, line 500: So even though multiget is a single call and here we are doing multiple calls of this exist method, you measured it and it is faster to use exist?

MatusKysel: yop, check the issue I put there result. It is 2x as fast because exist does not fetch data and it is very fast when trx does not exist

libraries/core_libs/storage/include/storage/storage.hpp
```diff
@@ -419,39 +419,39 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
...
-    template <typename T>
-    MultiGetQuery& append(Column const& col, T const& key, bool copy_key = true) {
-      return append(col, std::vector{toSlice(key)}, copy_key);
-    }
-
-    uint size();
-    std::vector<string> execute(bool and_reset = true);
-    MultiGetQuery& reset();
-  };
+  //   class MultiGetQuery {
```
mfrankovi, line 422: what is the point in commenting this out? Maybe leave it in since someone might use it later on or just completely remove it?

mfrankovi (approved):

gaoqi8649 (approved):

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -589,22 +584,12 @@ bool DbStorage::transactionFinalized(trx_hash_t const& hash) {
...
-  }
-
-  db_query.append(DbStorage::Columns::trx_period, trx_hashes);
-  db_trxs_statuses = db_query.execute();
-  for (size_t idx = 0; idx < db_trxs_statuses.size(); idx++) {
-    auto& trx_raw_status = db_trxs_statuses[idx];
-    result[idx] = result[idx] || (!trx_raw_status.empty());
+  for (size_t i = 0; i < trx_hashes.size(); ++i) {
+    const auto key = trx_hashes[i].asBytes();
+    if (exist(toSlice(key), Columns::transactions) || exist(toSlice(key), Columns::trx_period)) {
```
JakubFornadel, line 589: isnt it more probable that tx is rather already in Columns::trx_period than  Columns::transactions ? In such case we should change the order in this if

MatusKysel: nope, we use this call when we insert new verified transaction so they are more likely be part of `Columns::transactions`

JakubFornadel (approved):

kstdl (approved):

PR #1814, opened 2022-06-17
Remove chain config predefined

gaoqi8649:


tests/util_test/conf/conf_taraxa1.json
```diff
@@ -46,5 +46,94 @@
         ]
       }
     ]
+  },
```
mfrankovi, line 49: Maybe as part of this task we should remove all these files conf_taraxa1, conf_taraxa2, ... and only have one single file and then in code modify the fields that are different for each nodes. From what I can see only differences are ports and data path between these files

kstdl: I think we don't even need this one file. Before it was just default config(with default values for chain_config) with different ports and datadirs, so we can just use defaults I guess

PR #1816, opened 2022-06-20
feat: add vote rewards

MatusKysel:
## Purpose
EVM PR https://github.com/Taraxa-project/taraxa-evm/pull/87

### DIFF
--- a/libraries/core_libs/consensus/include/final_chain/final_chain.hpp
+++ b/libraries/core_libs/consensus/include/final_chain/final_chain.hpp
@@ -229,8 +229,8 @@ class FinalChain {
   }
 };
 
-std::shared_ptr<FinalChain> NewFinalChain(std::shared_ptr<DB> const& db, taraxa::ChainConfig const& config,
-                                          addr_t const& node_addr = {});
+std::shared_ptr<FinalChain> NewFinalChain(const std::shared_ptr<DB>& db, const taraxa::ChainConfig& config,
+                                          const addr_t& node_addr = {});
 /** @} */
 
 }  // namespace taraxa::final_chain
--- a/libraries/core_libs/consensus/include/final_chain/final_chain.hpp
+++ b/libraries/core_libs/consensus/include/final_chain/final_chain.hpp
@@ -5,7 +5,7 @@
 #include "common/event.hpp"
 #include "common/range_view.hpp"
 #include "common/types.hpp"
-#include "config/final_chain_config.hpp"
+#include "config/chain_config.hpp"
 #include "final_chain/data.hpp"
 #include "final_chain/state_api.hpp"
 #include "storage/storage.hpp"
@@ -229,7 +229,7 @@ class FinalChain {
   }
 };
 
-std::shared_ptr<FinalChain> NewFinalChain(std::shared_ptr<DB> const& db, Config const& config,
+std::shared_ptr<FinalChain> NewFinalChain(std::shared_ptr<DB> const& db, taraxa::ChainConfig const& config,
                                           addr_t const& node_addr = {});
 /** @} */
 
--- a/libraries/core_libs/consensus/src/final_chain/final_chain.cpp
+++ b/libraries/core_libs/consensus/src/final_chain/final_chain.cpp
@@ -1,5 +1,7 @@
 #include "final_chain/final_chain.hpp"
 
+#include <cstdint>
+
 #include "common/constants.hpp"
 #include "common/thread_pool.hpp"
 #include "final_chain/replay_protection_service.hpp"
@@ -11,6 +13,7 @@ namespace taraxa::final_chain {
 
 class FinalChainImpl final : public FinalChain {
   std::shared_ptr<DB> db_;
+  const uint32_t commitee_size_;
   std::unique_ptr<ReplayProtectionService> replay_protection_service_;
   StateAPI state_api_;
 
@@ -32,11 +35,12 @@ class FinalChainImpl final : public FinalChain {
   LOG_OBJECTS_DEFINE
 
  public:
-  FinalChainImpl(std::shared_ptr<DB> const& db, Config const& config, addr_t const& node_addr)
+  FinalChainImpl(std::shared_ptr<DB> const& db, taraxa::ChainConfig const& config, addr_t const& node_addr)
       : db_(db),
+        commitee_size_(config.pbft.committee_size),
         // replay_protection_service_(NewReplayProtectionService({}, db)),
         state_api_([this](auto n) { return block_hash(n).value_or(ZeroHash()); },  //
-                   config.state,
+                   config.final_chain.state,
                    {
                        1500,
                        4,
@@ -64,7 +68,8 @@ class FinalChainImpl final : public FinalChain {
 
           // Creates rewards stats
           RewardsStats rewards_stats;
-          std::vector<addr_t> txs_validators = rewards_stats.processStats(period_data);
+          std::vector<addr_t> txs_validators = rewards_stats.processStats(
+              period_data, dpos_eligible_total_vote_count(period_data.pbft_blk->getPeriod() - 1), commitee_size_);
 
           auto res = state_api_.transition_state(
               {blk->author, blk->gas_limit, blk->timestamp, BlockHeader::difficulty()},
@@ -77,11 +82,12 @@ class FinalChainImpl final : public FinalChain {
     } else {
       assert(state_db_descriptor.blk_num == 0);
       auto batch = db_->createWriteBatch();
-      last_block_ = append_block(batch, config.genesis_block_fields.author, config.genesis_block_fields.timestamp,
-                                 GAS_LIMIT, state_db_descriptor.state_root);
+      last_block_ =
+          append_block(batch, config.final_chain.genesis_block_fields.author,
+                       config.final_chain.genesis_block_fields.timestamp, GAS_LIMIT, state_db_descriptor.state_root);
       db_->commitWriteBatch(batch, db_opts_w_);
     }
-    delegation_delay_ = config.state.dpos->delegation_delay;
+    delegation_delay_ = config.final_chain.state.dpos->delegation_delay;
   }
 
   void stop() override { executor_thread_.stop(); }
@@ -104,7 +110,8 @@ class FinalChainImpl final : public FinalChain {
 
     RewardsStats rewards_stats;
     // returns list of validators for new_blk.transactions
-    std::vector<addr_t> txs_validators = rewards_stats.processStats(new_blk);
+    std::vector<addr_t> txs_validators = rewards_stats.processStats(
+        new_blk, dpos_eligible_total_vote_count(new_blk.pbft_blk->getPeriod() - 1), commitee_size_);
 
     block_applying_emitter_.emit(block_header()->number + 1);
 
@@ -442,7 +449,7 @@ class FinalChainImpl final : public FinalChain {
   };
 };
 
-std::shared_ptr<FinalChain> NewFinalChain(std::shared_ptr<DB> const& db, Config const& config,
+std::shared_ptr<FinalChain> NewFinalChain(std::shared_ptr<DB> const& db, taraxa::ChainConfig const& config,
                                           addr_t const& node_addr) {
   return make_shared<FinalChainImpl>(db, config, node_addr);
 }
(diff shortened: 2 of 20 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1726,6 +1730,24 @@ bool PbftManager::pushPbftBlock_(PeriodData &&period_data, std::vector<std::shar
...
+    return;
+  }
+  const auto dpos_votes_count = final_chain_->dpos_eligible_total_vote_count(period_data.pbft_blk->getPeriod() - 1);
+  const auto min_weight = std::min<size_t>(COMMITTEE_SIZE, dpos_votes_count) * 2 / 3 + 1;
+  size_t total_weight = 0;
+
+  for (size_t i = 0; i < period_data.previous_block_cert_votes.size(); i++) {
+    assert(period_data.previous_block_cert_votes[i]->getWeight());
+    total_weight += *period_data.previous_block_cert_votes[i]->getWeight();
+    if (total_weight > min_weight) {
```
JakubFornadel, line 1745: if dpos_votes_count is lets say 2000 and I have 1600 so my vote would be enough, this condition would count it as bonus vote, which is not true. Overall this for loop with indexes could be probably more simple:
```

bool is_bonus_vote = false;
for (const auto&  vote: period_data.previous_block_cert_votes) {
  ...
 
  // need this delay to not count first vote > min_weight as bonus vote
  if (is bonus_vote) {
    period_data.bonus_votes_count++;
  }
  
  if (total_weight > min_weight) { 
    is_bonus_vote = true;
  }
}
```

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -29,9 +29,9 @@ void PbftSyncPacketHandler::validatePacketRlpFormat(const PacketData &packet_dat
 
   // PeriodData rlp parsing cannot be done through util::rlp_tuple, which automatically checks the rlp size so it is
   // checked here manually
-  if (constexpr size_t required_size = 4; packet_data.rlp_[1].itemCount() != required_size) {
+  if (packet_data.rlp_[1].itemCount() != PeriodData::kItemCount) {
```
JakubFornadel, line 32: kItemsCount

libraries/types/pbft_block/src/period_data.cpp
```diff
@@ -26,15 +26,17 @@ PeriodData::PeriodData(dev::RLP&& rlp) {
     dag_blocks.emplace_back(dag_block_rlp);
   }
 
-  for (auto const trx_rlp : *it) {
+  for (auto const trx_rlp : *it++) {
     transactions.emplace_back(std::make_shared<Transaction>(trx_rlp));
   }
+
+  bonus_votes_count = (*it).toInt<size_t>();
```
JakubFornadel, line 33: this info is inside previous_block_cert_votes, isnt it ?

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
```diff
@@ -66,7 +64,9 @@ void RewardsStats::initStats(const PeriodData& sync_blk) {
     }
   }
 
-  // TODO: add votes to rewards_stats
+  for (const auto& vote : sync_blk.previous_block_cert_votes) {
+    addVote(vote);
```
JakubFornadel, line 68: Why do we need to have bonus_count info inside PeriodData ? Isnt it enough to calculate it here and set it to rewards_stats as this info i only needed in dpos contract ?

MatusKysel: you do not have vote weigh in this time as it is calculated on vote verification. e.g. after restarting node some pbft blocks could be finalized but not executed, so you would need to revalidate all the votes again

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
MatusKysel: but this way you need to iterate whole array I do not need to. And yes, if I change order of line 1744 and put it under if, I would fix the case you mentioned

libraries/types/pbft_block/include/pbft/period_data.hpp
```diff
@@ -31,9 +31,13 @@ class PeriodData {
   std::shared_ptr<PbftBlock> pbft_blk;
   std::vector<std::shared_ptr<Vote>> previous_block_cert_votes;  // These votes are the cert votes of previous block
                                                                  // which match reward votes in current pbft block
+  size_t bonus_votes_count;
```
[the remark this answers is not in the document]
MatusKysel: this is public member no _

libraries/types/pbft_block/src/period_data.cpp
MatusKysel: yes, but I need to save it to DB or re-validate all votes, not sure what is better

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
JakubFornadel: So during normal node operation, these votes are already verified and for the " node restart scenario", isnt the max possible difference between finalized & executed blocks just a few blocks ? If yes, I think it is worth to do this verification during node init as it is 1 time operation and it shouldnt really take too much time

libraries/types/pbft_block/src/period_data.cpp
JakubFornadel: this is related to the previous comment right ? You would need to revalidate few blocks only after "node restart" ? If so, I think revalidating is better

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
MatusKysel: After https://github.com/Taraxa-project/taraxa-node/pull/1818 merged  I will change this architecture

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
MatusKysel: yes, there should be break

libraries/types/pbft_block/include/pbft/period_data.hpp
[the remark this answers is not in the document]
MatusKysel: hmm, so the whole class is wrong then?

libraries/types/pbft_block/src/period_data.cpp
MatusKysel: yes, this will be removed

libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
MatusKysel: actually it should be without s, it is numer_of_items or ItemCount not ItemsCount

mfrankovi (approved):

libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -32,11 +35,12 @@ class FinalChainImpl final : public FinalChain {
   LOG_OBJECTS_DEFINE
 
  public:
-  FinalChainImpl(std::shared_ptr<DB> const& db, Config const& config, addr_t const& node_addr)
+  FinalChainImpl(std::shared_ptr<DB> const& db, taraxa::ChainConfig const& config, addr_t const& node_addr)
```
JakubFornadel, line 38: same here

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
```diff
@@ -42,35 +44,46 @@ std::optional<addr_t> RewardsStats::getTransactionValidator(const trx_hash_t& tx
...
+  assert(vote->getWeight());
+
+  if (validator_stats.vote_weight_) {
+    return false;
+  }
 
-  validator_stats.valid_cert_vote_ = true;
-  total_unique_votes_count_++;
+  const auto weight = *vote->getWeight();
+  if (total_votes_weight_ > min_weight) {
```
JakubFornadel, line 58: let' say there are 4 votes with weight 1 and min_weight == 3, bonus_votes_weight_ will be 0 in such scenario but it should be 1. There should be `>=`

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
```diff
@@ -42,35 +44,46 @@ std::optional<addr_t> RewardsStats::getTransactionValidator(const trx_hash_t& tx
...
     for (const auto& tx_hash : dag_block.getTrxs()) {
       addTransaction(tx_hash, dag_block_author);
     }
   }
-
-  // TODO: add votes to rewards_stats
+  max_votes_weight_ = std::min<uint64_t>(committee_size, dpos_vote_count);
+  const uint64_t min_weight = max_votes_weight_ * 2 / 3 + 1;
+  for (const auto& vote : sync_blk.previous_block_cert_votes) {
+    addVote(vote, min_weight);
```
JakubFornadel, line 80: why don't you create class member min_weight_ and set it few lines above instead of sending the same value hundreds of times into the addVote function ?

libraries/core_libs/consensus/include/final_chain/final_chain.hpp
```diff
@@ -229,7 +229,7 @@ class FinalChain {
   }
 };
 
-std::shared_ptr<FinalChain> NewFinalChain(std::shared_ptr<DB> const& db, Config const& config,
+std::shared_ptr<FinalChain> NewFinalChain(std::shared_ptr<DB> const& db, taraxa::ChainConfig const& config,
```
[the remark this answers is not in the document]
MatusKysel: I did just called replace, not even check the where it is used

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
```diff
@@ -42,35 +44,46 @@ std::optional<addr_t> RewardsStats::getTransactionValidator(const trx_hash_t& tx
...
-bool RewardsStats::addVote(const Vote& vote) {
+bool RewardsStats::addVote(const std::shared_ptr<Vote>& vote, uint64_t min_weight) {
   // Set valid cert vote to validator
-  auto& validator_stats = validators_stats_[vote.getVoterAddr()];
-  assert(validator_stats.valid_cert_vote_ == false);
+  auto& validator_stats = validators_stats_[vote->getVoterAddr()];
+  assert(validator_stats.vote_weight_ == 0);
+  assert(vote->getWeight());
+
+  if (validator_stats.vote_weight_) {
```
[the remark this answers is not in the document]
MatusKysel: it can happen in release build

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
```diff
@@ -42,35 +44,46 @@ std::optional<addr_t> RewardsStats::getTransactionValidator(const trx_hash_t& tx
...
+    return false;
+  }
 
-  validator_stats.valid_cert_vote_ = true;
-  total_unique_votes_count_++;
+  const auto weight = *vote->getWeight();
```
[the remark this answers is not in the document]
MatusKysel: why? this is doing exactly same and it is more efficient https://en.cppreference.com/w/cpp/utility/optional/value

libraries/core_libs/consensus/src/final_chain/rewards_stats.cpp
MatusKysel: because we are passing whole class to go and I did not want to add additional values that is not required by go

libraries/types/pbft_block/include/pbft/period_data.hpp
```diff
@@ -34,6 +34,8 @@ class PeriodData {
   std::vector<DagBlock> dag_blocks;
   SharedTransactions transactions;
 
+  const static size_t kItemCount = 4;
```
kstdl, line 37: kRlpItemCount?

kstdl (approved):

PR #1817, opened 2022-06-21
chore: remove dag transaction limit and refactor dag block gas limit

mfrankovi:
Hard transaction limit of 250 transactions in a DAG block is removed since dag gas limit is the mechanism used to limit DAG blocks size

Gas limit for dag block was refactored to fix an existing bug and make it more efficienct

### DIFF
--- a/libraries/core_libs/consensus/src/dag/block_proposer.cpp
+++ b/libraries/core_libs/consensus/src/dag/block_proposer.cpp
@@ -100,15 +100,16 @@ bool SortitionPropose::propose() {
     }
   }
 
-  SharedTransactions shared_trxs = proposer->getShardedTrxs();
-  if (shared_trxs.empty()) {
+  auto [transactions, estimations] = proposer->getShardedTrxs(*proposal_period, dag_blk_mgr_->getDagConfig().gas_limit);
+  if (transactions.empty()) {
     last_frontier_ = frontier;
     num_tries_ = 0;
     return false;
   }
   LOG(log_nf_) << "VDF computation time " << vdf.getComputationTime() << " difficulty " << vdf.getDifficulty();
   last_frontier_ = frontier;
-  proposer->proposeBlock(std::move(frontier), propose_level, *proposal_period, std::move(shared_trxs), std::move(vdf));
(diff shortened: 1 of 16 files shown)


MatusKysel (approved):

libraries/core_libs/consensus/src/dag/block_proposer.cpp
```diff
@@ -162,7 +162,7 @@ SharedTransactions BlockProposer::getShardedTrxs() {
     return {};
   }
 
-  SharedTransactions to_be_packed_trx = trx_mgr_->packTrxs(bp_config_.transaction_limit);
+  SharedTransactions to_be_packed_trx = trx_mgr_->packTrxs();
```
JakubFornadel, line 165: not ideal. If you have 10k txs in queue, you will do 10k copies of shared_pointers and put them into newly allocated vector here. Most of them are not needed because you will later use only few hundred of these txs... We could somehow change the code:

- either we will calculate tx gas while getting it from tx queue (this would block the queue for long time as there is a lock)
- get txs in batches by (250/500 ?), calculate the estimation and if didnt reach the limit then get another batch... 
- by knowing the tx with smallest gas, we could know how many such txs could be in single block and we can ask that many txs from queue (not all of them like now)

mfrankovi: I refactored this part, please review again

libraries/core_libs/consensus/src/dag/block_proposer.cpp
```diff
@@ -217,9 +217,6 @@ void BlockProposer::proposeBlock(DagFrontier&& frontier, level_t level, uint64_t
   for (const auto& trx : trxs) {
     auto weight = trx_mgr_->estimateTransactionGas(trx, proposal_period);
     block_weight += weight;
```
MatusKysel, line 219: `estimateTransactionGas` is expensive call and now you are executing it 2x. I would modify this code too, so as we already calling it from `packTrxs`. And `packTrxs` should also return estimations array

JakubFornadel (approved):

PR #1835, opened 2022-06-30
chore: optimize transactions processing

mfrankovi:
Anything running under transaction_mutex in transaction_manager can be a major bottleneck for high TPS load.

transaction_accepted_.emit is moved out mutex since there was no need to run it under mutex.

Estimating transaction gas done under transaction mutex was a major bottleneck for transactions processing since estimation takes time. To avoid that batch of transactions from memory pool is retrieved and estimation is now done outside of mutex. I made an assumption that dag block gas limit divided by 21000 is max number of transaction we can fit under a single DAG block. I am not sure if this is a valid assumption. Is it possible to have transaction that does not take at least 21000 gas?

With these changes I am able to get  over 20k TPS processed in a single node

### DIFF
--- a/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
@@ -209,31 +209,37 @@ std::shared_ptr<Transaction> TransactionManager::getTransaction(trx_hash_t const
 }
 
 void TransactionManager::saveTransactionsFromDagBlock(SharedTransactions const &trxs) {
-  // This lock synchronizes inserting and removing transactions from transactions memory pool with database insertion.
-  // Unique lock here makes sure that transactions we are removing are not reinserted in transactions_pool_
-  std::unique_lock transactions_lock(transactions_mutex_);
+  std::vector<trx_hash_t> accepted_transactions;
+  accepted_transactions.reserve(trxs.size());
   auto write_batch = db_->createWriteBatch();
   vec_trx_t trx_hashes;
   std::transform(trxs.begin(), trxs.end(), std::back_inserter(trx_hashes),
                  [](std::shared_ptr<Transaction> const &t) { return t->getHash(); });
 
-  auto trx_in_db = db_->transactionsInDb(trx_hashes);
-  for (uint64_t i = 0; i < trxs.size(); i++) {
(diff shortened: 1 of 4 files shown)


libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -211,29 +211,35 @@ std::shared_ptr<Transaction> TransactionManager::getTransaction(trx_hash_t const
 void TransactionManager::saveTransactionsFromDagBlock(SharedTransactions const &trxs) {
   // This lock synchronizes inserting and removing transactions from transactions memory pool with database insertion.
   // Unique lock here makes sure that transactions we are removing are not reinserted in transactions_pool_
```
MatusKysel, line 213: move this comment to lock location

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -297,21 +303,22 @@ std::vector<std::shared_ptr<Transaction>> TransactionManager::getNonfinalizedTrx
...
   std::vector<uint64_t> estimations;
-  transactions_pool_.resetGetNextTransactionIterator();
   uint64_t total_weight = 0;
-  while (true) {
-    auto trx = transactions_pool_.getNextTransaction();
-    if (trx == nullptr) break;
-    uint64_t weight = estimateTransactionGas(trx, proposal_period);
+  {
+    std::shared_lock transactions_lock(transactions_mutex_);
+    uint64_t max_transactions_in_block = weight_limit / kMinTxGas;
```
MatusKysel, line 311: this could be const and maybe not under lock?

MatusKysel (approved):

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
mfrankovi: Done

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
mfrankovi: Set to const and moved out

JakubFornadel (approved):

PR #1839, opened 2022-07-04
feat: separate step logic from manager into specific classes

kstdl:
Separated steps specific logic into different classes with common base class `Step` . Some of common logic is in this `Step` class.

### DIFF
--- a/libraries/core_libs/consensus/include/pbft/step/finish.hpp
+++ b/libraries/core_libs/consensus/include/pbft/step/finish.hpp
@@ -1,0 +1,20 @@
+#pragma once
+#include "logger/logger.hpp"
+#include "pbft/step/step.hpp"
+
+namespace taraxa::step {
+
+class Finish : public Step {
+ public:
+  Finish(uint64_t id, std::shared_ptr<CommonState> state, std::shared_ptr<NodeFace> node)
+      : Step(id, std::move(state), std::move(node)) {
+    init();
+    assert(id % 2 == 0);
+  }
+  void run() override;
+
+ private:
+  void init() override;
+};
+
+}  // namespace taraxa::step
--- a/libraries/core_libs/consensus/include/pbft/step/step.hpp
+++ b/libraries/core_libs/consensus/include/pbft/step/step.hpp
@@ -1,0 +1,109 @@
+#pragma once
+#include <cassert>
+#include <memory>
+
+#include "common/types.hpp"
+#include "network/network.hpp"
+
+namespace taraxa {
+
+using time_point = std::chrono::system_clock::time_point;
+
+#define kNullBlockHash blk_hash_t(0)
+
+enum PbftStates { propose = 1, filter, certify, finish, polling };
+
+// TODO: Should be removed or minimized
+struct CommonState {
+  PbftStates state_ = propose;
+
+  time_point round_clock_initial_datetime_;
+  time_point now_;
+
+  time_point time_began_waiting_next_voted_block_;
+  time_point time_began_waiting_soft_voted_block_;
+  std::chrono::duration<double> duration_;
+  uint64_t next_step_time_ms_ = 0;
+  uint64_t elapsed_time_in_round_ms_ = 0;
+
+  blk_hash_t previous_round_next_voted_value_ = kNullBlockHash;
+  blk_hash_t own_starting_value_for_round_ = kNullBlockHash;
+
+  blk_hash_t last_soft_voted_value_ = kNullBlockHash;
+  blk_hash_t last_cert_voted_value_ = kNullBlockHash;
+
+  blk_hash_t soft_voted_block_for_this_round_ = kNullBlockHash;
+};
+
+// TODO: Should be real node interface?
+struct NodeFace {
+  NodeFace(const addr_t& addr, const secret_t& node_sk, const vrf_wrapper::vrf_sk_t& vrf_sk,
+           std::shared_ptr<DbStorage> db, std::shared_ptr<NextVotesManager> nvm, std::shared_ptr<PbftChain> pc,
+           std::shared_ptr<VoteManager> vm, std::shared_ptr<DagManager> dm, std::shared_ptr<DagBlockManager> dbm,
+           std::shared_ptr<TransactionManager> tm, std::shared_ptr<FinalChain> fc)
+      : node_addr_(addr),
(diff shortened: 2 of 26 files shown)


libraries/core_libs/consensus/include/pbft/step/finish.hpp
```diff
@@ -0,0 +1,20 @@
...
+#include "pbft/step/step.hpp"
+
+namespace taraxa::step {
+
+class Finish : public Step {
+ public:
+  Finish(uint64_t id, std::shared_ptr<CommonState> state, std::shared_ptr<NodeFace> node)
+      : Step(id, std::move(state), std::move(node)) {
+    init();
+    assert(id % 2 == 0);
```
MatusKysel, line 12: finish needs to be always even?

libraries/core_libs/consensus/include/pbft/step/step.hpp
```diff
@@ -0,0 +1,109 @@
...
+#include <memory>
+
+#include "common/types.hpp"
+#include "network/network.hpp"
+
+namespace taraxa {
+
+using time_point = std::chrono::system_clock::time_point;
+
+#define kNullBlockHash blk_hash_t(0)
```
MatusKysel, line 12: this is already defined `static const blk_hash_t kNullBlockHash = blk_hash_t(0);`

libraries/core_libs/consensus/include/pbft/step/step.hpp
```diff
@@ -0,0 +1,109 @@
...
+  std::shared_ptr<TransactionManager> trx_mgr_;
+  std::shared_ptr<FinalChain> final_chain_;
+};
+
+class Step {
+ public:
+  Step(uint64_t id, std::shared_ptr<CommonState> state, std::shared_ptr<NodeFace> node)
+      : id_(id), state_(std::move(state)), node_(std::move(node)) {
+    const auto& node_addr = node_->node_addr_;
+    LOG_OBJECTS_CREATE("STEP" + std::to_string(id_));
```
MatusKysel, line 76: maybe we could call `init()` here?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -31,64 +36,44 @@ PbftManager::PbftManager(const PbftConfig &conf, const blk_hash_t &dag_genesis_b
...
 
-void PbftManager::setNetwork(std::weak_ptr<Network> network) { network_ = move(network); }
+void PbftManager::setNetwork(std::weak_ptr<Network> network) { node_->network_ = std::move(network); }
 
 void PbftManager::start() {
   if (bool b = true; !stopped_.compare_exchange_strong(b, !b)) {
     return;
   }
 
+  node_->pbft_manager_ = shared_from_this();
```
MatusKysel, line 60: weak_from_this ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -313,16 +276,17 @@ size_t PbftManager::getDposWeightedVotesCount() const { return weighted_votes_co
...
   } catch (state_api::ErrFutureBlock &c) {
     LOG(log_er_) << c.what() << ". Period " << dpos_period_ << " is too far ahead of DPOS";
     return 0;
   }
 }
 
 void PbftManager::setPbftStep(size_t const pbft_step) {
-  last_step_ = step_;
-  db_->savePbftMgrField(PbftMgrRoundStep::PbftStep, pbft_step);
+  static std::default_random_engine random_engine_{std::random_device{}()};
```
MatusKysel, line 287: this is not member anymore, not sure if you should use _ as suffix

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -360,87 +323,86 @@ bool PbftManager::resetRound_() {
...
 
   LOG(log_nf_) << "Determined round from votes: " << consensus_pbft_round;
-  round_clock_initial_datetime_ = now_;
+  shared_state_->round_clock_initial_datetime_ = shared_state_->now_;
   // Update current round and reset step to 1
   round_ = consensus_pbft_round;
   resetStep_();
-  state_ = value_proposal_state;
+  shared_state_->state_ = propose;
+  current_step_ = std::make_unique<step::Propose>(shared_state_, node_);
```
MatusKysel, line 331: I would create some kind of factory function that will just take `step::Propose` type as arg and will do this thing. But this is just my suggestion, so you do not need to do `shared_state_, node_` all the time

MatusKysel (approved):

tests/pbft_manager_test.cpp
```diff
@@ -179,67 +179,67 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
-  }
-  EXPECT_EQ(skipped_soft_voting, true);
-
-  auto start_round = pbft_mgr->getPbftRound();
-  pbft_mgr->resume();
-
-  std::cout << "Wait ensure node is still advancing in rounds... " << std::endl;
-  EXPECT_HAPPENS({60s, 50ms}, [&](auto &ctx) { WAIT_EXPECT_NE(ctx, start_round, pbft_mgr->getPbftRound()) });
-}
+// TEST_F(PbftManagerTest, terminate_soft_voting_pbft_block) {
```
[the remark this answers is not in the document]
kstdl: This is the only test that was using "only for tests" methods in pbft manager. So I commented test and this methods. I don't like idea that we have some public methods just for tests and maybe it will be easier to rework this test later. So I left it commented just to not forget about it

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
kstdl: I just decided to left it for next task(my current task), because it will be changed in it even more I think

libraries/core_libs/consensus/include/pbft/step/step.hpp
kstdl: No. It is bad idea to call any virtual method from constructor.
Virtual table is not initialized yet, so it will try to call `init()` from this(base) class, it is not expected behaviour. In our case it won't work at all because this `init()` is pure virtual

PR #1852, opened 2022-07-12
fix: sending transactions

mfrankovi:
When sending transactions in dag block packet handler kMaxTransactionsInPacket limit was not used which would make us fail such a packet on receiving end and mark node as malicious.

### DIFF
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
@@ -83,16 +83,20 @@ void DagBlockPacketHandler::sendBlock(dev::p2p::NodeID const &peer_id, taraxa::D
   std::unique_lock lock(peer->mutex_for_sending_dag_blocks_);
 
   // Transactions are first sent in transactions packet before sending the block
-  if (!trxs.empty()) {
-    dev::RLPStream s;
+  uint32_t index = 0;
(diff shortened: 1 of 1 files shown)


justinsnapp (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
```diff
@@ -83,16 +83,20 @@ void DagBlockPacketHandler::sendBlock(dev::p2p::NodeID const &peer_id, taraxa::D
   std::unique_lock lock(peer->mutex_for_sending_dag_blocks_);
 
   // Transactions are first sent in transactions packet before sending the block
-  if (!trxs.empty()) {
-    dev::RLPStream s;
+  uint32_t index = 0;
+  while (index < trxs.size()) {
+    uint32_t trx_count_to_send = std::min(static_cast<size_t>(kMaxTransactionsInPacket), trxs.size() - index);
```
MatusKysel, line 88: can be const

MatusKysel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
mfrankovi: Fixed

kstdl (approved):

Issue #1878, opened 2022-07-21
Process unsupported transaction type RLP and return an adequate error


kstdl:
Right now sending unsupported typed transaction to node will return an RLP error that can confuse. Process it correctly and return correct error
```{"error":{"code":-32603,"message":"Unhandled API exception: /opt/taraxa/libraries/aleth/libdevcore/RLP.cpp(24): Throw in function dev::RLP::RLP(dev::bytesConstRef, dev::RLP::Strictness)\\nDynamic exception type: boost::wrapexcept<dev::OversizeRLP>\\n[dev::tag_comment*] = Actual size: 1, data size: 5025\\n"},"id":45,"jsonrpc":"2.0"}```

PR #1879, opened 2022-07-21
fix: describe RLP parsing fail

kstdl:
#1878

### DIFF
--- a/libraries/types/transaction/src/transaction.cpp
+++ b/libraries/types/transaction/src/transaction.cpp
@@ -1,3 +1,4 @@
+
 #include "transaction/transaction.hpp"
 
 #include <libdevcore/CommonJS.h>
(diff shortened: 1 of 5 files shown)


libraries/types/transaction/src/transaction.cpp
```diff
@@ -34,12 +35,26 @@ Transaction::Transaction(const trx_nonce_t &nonce, const val_t &value, const val
   getSender();
 }
 
+Transaction::Transaction(const bytes &_bytes, bool verify_strict, const h256 &hash) {
+  dev::RLP rlp;
+  try {
+    rlp = dev::RLP(_bytes);
+  } catch (const dev::RLPException &e) {
+    std::string error_msg =
```
MatusKysel, line 43: Do we have issue for that? if so maybe we should add TODO here, so it is easier to find what needs to be removed/changed

MatusKysel (approved):

mfrankovi (approved):

PR #1884, opened 2022-07-25
dag blocks processing refactor

mfrankovi:
This refactor besides optimizing dag blocks processing fixes issue found on testnet for a race condition between removing expired DAG block from DAG and inserting a DAG block that points to a pivot/tip that is about to expire.
Since parallel processing using multiple threads is done on the packets level having a separate worker thread and dag blocks queue for processing dag blocks is not needed. DAG blocks order is guaranteed by the gossip protocol so valid dag blocks 
can be verified and inserted on the packet handler thread. 
DagBlockManager class is removed and small leftover of its functionality for verifying dag blocks is moved into DagManager. DagManager no longer has a worker thread and does not need start/stop methods.


libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
```diff
@@ -102,18 +101,26 @@ void DagSyncPacketHandler::process(const PacketData& packet_data, const std::sha
 
     received_dag_blocks_str += block.getHash().abridged() + " ";
 
-    auto status = checkDagBlockValidation(block);
-    if (!status.first) {
-      // This should only happen with a malicious node or a fork
-      std::ostringstream err_msg;
-      err_msg << "DagBlock " << block.getHash() << " validation failed: " << status.second;
+    if (block.getLevel() > peer->dag_level_) peer->dag_level_ = block.getLevel();
```
MatusKysel, line 104: I guess we should set this after the verification, as when verification fails we will disconnect

libraries/core_libs/consensus/include/dag/dag.hpp
```diff
@@ -24,8 +24,8 @@
 #include "common/types.hpp"
```
MatusKysel, line 24: can we split this file to dag and dag_mngr? I do not really like combine those two things into one cpp/hpp

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
```diff
@@ -112,43 +90,60 @@ void DagBlockPacketHandler::sendBlock(dev::p2p::NodeID const &peer_id, taraxa::D
 }
 
 void DagBlockPacketHandler::onNewBlockReceived(DagBlock &&block, const std::shared_ptr<TaraxaPeer> &peer) {
-  if (dag_blk_mgr_) [[likely]] {
+  if (dag_mgr_) [[likely]] {
```
MatusKysel, line 93: does this need to be separate function? process is doing almost nothing and if we are not calling this from somewhere else I would remove it and move this code to process function

MatusKysel (approved):

libraries/core_libs/consensus/src/dag/dag.cpp
```diff
@@ -422,20 +392,26 @@ bool DagManager::addDagBlock(DagBlock &&blk, SharedTransactions &&trxs, bool pro
...
-          LOG(log_nf_) << "Dropping old block: " << blk_hash << ". Expiry level: " << dag_expiry_level
+
+        if (blk.getLevel() < dag_expiry_level_) {
+          LOG(log_nf_) << "Dropping old block: " << blk_hash << ". Expiry level: " << dag_expiry_level_
                        << ". Block level: " << blk.getLevel();
-          return false;
+          return {false, std::nullopt};
+        }
+
+        auto res = pivotAndTipsAvailable(blk);
```
kstdl, line 406: Why is it changed to return a pair, but second is not used?

kstdl: Ok, I see that we use it to add to disconnect reason for peer later. But would it be useful? This pair looks pretty weird for me here and we aren't sending all missing hashes, but only first one we found(we could miss a few, right?). So it is not very useful as I see

kstdl (approved):

libraries/core_libs/consensus/src/dag/dag.cpp
kstdl: Maybe it would be cleaner to change this method to `missingPivotOrTips` and return collection with missing hashes. If it is empty we know that we have all of them

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
mfrankovi: We call it from one old test, maybe this test should be refactored

libraries/core_libs/consensus/include/dag/dag.hpp
```diff
@@ -161,35 +174,46 @@ class DagManager : public std::enable_shared_from_this<DagManager> {
...
 
+  void setNetwork(std::weak_ptr<Network> network) { network_ = std::move(network); }
+
   /**
-   * @brief Start the worker thread processing new DAG blocks
+   * @param hash
+   * @return true in case block was already seen or is part of dag structure
    */
-  void start();
+  bool isDagBlockKnown(const blk_hash_t &hash);
```
JakubFornadel, line 183: should be const probably ?

JakubFornadel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
mfrankovi: Done

libraries/core_libs/consensus/include/dag/dag.hpp
mfrankovi: Done

libraries/core_libs/consensus/src/dag/dag.cpp
mfrankovi: Done

libraries/core_libs/consensus/include/dag/dag.hpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
mfrankovi: TODO: Fix in https://github.com/Taraxa-project/taraxa-node/issues/1895

Issue #1889, opened 2022-07-26
Lower number of snapshots to 1 for testnet/mainnet


MatusKysel:
As community was complaining about disk space we created this tutorial: https://docs.taraxa.io/faq/testnet#why-is-my-node-eating-up-so-much-disk-space

I do not really like users touch blockchain data, so my suggestion is to lower number of snapshots to 1 for testnet/mainnet. I believe we do not need more and one snapshot 10k blocks back will not take much data just few hundred MBs

PR #1892, opened 2022-07-27
chore: lower number of snapshots

MatusKysel:
## Purpose
Fixes: #1889


mfrankovi (commented):
Maybe this could be a separate task but we have this thing that we do not create snapshots when we are syncing and that is something I often found problematic in my testing because if I would sync and get somehwere stuck and there was no snapshots. With creating only 1 snapshot is creating snapshots while syncing expensive?

mfrankovi (approved):

MatusKysel:
@mfrankovi hmm good point maybe we should create task for that as we need to measure it

JakubFornadel (approved):

PR #1894, opened 2022-07-28
chore: unify network & chain id

### LINKED ISSUE #1812: Leave only one of network_id and chain_id
Right now we have network_id in network config and chain_id in chain config. But it should be the same thing, or it should be totally separated if not. 
Right now we [have this](https://github.com/Taraxa-project/taraxa-node/blob/develop/libraries/config/src/config.cpp#L257), but we can have situation when network_id will be changed in the config, but not chain id and it could mess something.

MatusKysel:
## Purpose
Fixes #1812

### DIFF
--- a/libraries/cli/include/cli/config_jsons/default.json
+++ b/libraries/cli/include/cli/config_jsons/default.json
@@ -1,4 +1,5 @@
 {
+  "chain_id": 4,
   "node_secret": "",
   "vrf_secret": "",
   "data_path": "",
@@ -143,7 +144,6 @@
     ]
   },
   "chain_config": {
-    "chain_id": "0x4",
     "dag_genesis_block": {
       "level": "0x0",
       "pivot": "0x0000000000000000000000000000000000000000000000000000000000000000",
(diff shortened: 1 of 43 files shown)


libraries/cli/include/cli/config_jsons/default.json
```diff
@@ -143,7 +144,6 @@
     ]
   },
   "chain_config": {
-    "chain_id": "0x4",
```
JakubFornadel, line 146: hmm chain_id sounds like a chain_config param ?

JakubFornadel (approved):

libraries/cli/include/cli/config_jsons/default.json
MatusKysel: as we unified network_id and chain_id it can not be chain config param anymore

MatusKysel: and one more thing chain config is only optional

justinsnapp:
Does having these be split and not unified serve the purpose of being able to more easily fork Taraxa.  So that you could have taraxafork mainnet and testnet and devnet etc.

MatusKysel:
> Does having these be split and not unified serve the purpose of being able to more easily fork Taraxa. So that you could have taraxafork mainnet and testnet and devnet etc.

using 2 different values is from eth, where network_id is for p2p and chain_id for signing trxs (so they can not use trxs from testnet on mainnet) more here:  https://besu.hyperledger.org/en/stable/Concepts/NetworkID-And-ChainID/ . I do not think there will be scenario where we change chain_id or network_id on real network, but if we planning to have this feature we should not unify them

mfrankovi (approved):

Issue #1880, opened 2022-07-21
Differentiate between standard & reward vote through flag


JakubFornadel:
There is ton of occasions when we wrongly identify standard vote as a reward vote, which leads to validation fails, etc...

The reason is that we identify vote as reward vote based on vote pbft round:
`if (vote_round < current_pbft_round) {  // reward vote`

That is problem because some out of sync nodes might be sending us standard votes with older round and we identify it as reward votes... These votes could be dropped right away if we knew they are supposed to be standard votes...

### Solution

Add flag into the vote, which will explicitly say "this is reward vote"

To reproduce this problem, run:
`PbftManagerTest.check_get_eligible_vote_count  `

justinsnapp:
@JakubFornadel , I would have naively thought a standard vote could be used as a reward vote?   Remind me to ask for more insight into this in a meeting (unless its easy to just answer as a comment here.)

Issue #1896, opened 2022-07-28
Create cache for dpos contract fetched data


JakubFornadel:
Create cache (probably in final_chain) and cache fetched data from dpos contract for last N blocks. For example these calls are called hundreds of times each round due to votes validation and most of the time they should return the same values:

```
  try {
    // TODO: final_chain_->dpos... ret values should be cached as it is heavily used and most of the time it returns the
    // same value!
    voter_dpos_votes_count = final_chain_->dpos_eligible_vote_count(period, node_addr_);
    total_dpos_votes_count = final_chain_->dpos_eligible_total_vote_count(period);
    pbft_sortition_threshold = getPbftSortitionThreshold(vote_type, period);

  } catch (state_api::ErrFutureBlock &e) {
    LOG(log_er_) << "Unable to place vote for round: " << round << ", period: " << period << ", step: " << step
                 << ", voted block hash: " << blockhash.abridged() << ". "
                 << "Period  is too far ahead of actual finalized pbft chain size ("
                 << final_chain_->last_block_number() << "). Err msg: " << e.what();

    return 0;
  }
```

Thanks to this, we could also get rid of the current cache, which is holding dpos values just for the latest period -> it will become redundant:

```
size_t getDposTotalVotesCount() const;
size_t getDposWeightedVotesCount() const;
uint64_t getCurrentPbftSortitionThreshold(PbftVoteTypes vote_type) const;
etc...
```

PR #1900, opened 2022-07-29
Votes ddos protection

JakubFornadel:
## Purpose

This PR is removing concept of unverified votes. Previously we would accept every vote and without any validation we would put it into the unverified queue, which could lead to memory grow...

Many possible ddos attacks are now eliminated by these changes, but not all of them. Anyway we could probably merge this PR so Dmitry can use it and then we can have separate PR with additional changes...

In the new version:

- each vote contains also pbft period so we can validate it against dpos contract as soon as we receive it
- each vote is fully validated in packet handlers and if validation fails, vote is dropped
- there is no unverified queue for votes anymore
- each voter can create only 1 unique vote per round & step & voter address. Exception is next_vote - we allow 2 unique votes - 1 for some specific block hash and 1 for NULL_BLOCK_HASH
- if some voter already sent standard vote with round N, any new vote from him with round N-1 is dropped (reward & sync next) votes are exceptions

##### Standard vote:
`vote_round >= current_pbft_round`
// Here should be vote_period > current_pbft_period but it breaks some tests due to missing rewards vote -> rewards votes need to be revisited
`vote_period >= current_pbft_period && vote_period <= current_pbft_period + dpos_delay(5)`

##### Reward vote:
`vote_round < current_pbft_round`
`vote_period == current_pbft_period`

##### Sync next vote:
`vote_round >= current_pbft_round - 1`
`vote_period >= current_pbft_period && vote_period <= current_pbft_period + dpos_delay(5)`

## Problems

- We differentiate standard & reward simply by checking round, `if vote_round < current_pbgt_round` we identify vote as reward vote and use validation for rewards votes -> this might be a problem as some out of sync node might send us standard vote and we process it as reward vote.... https://github.com/Taraxa-project/taraxa-node/issues/1880
- when running local-net I observed many errors on reward votes validation, when `vote_round < current_pbft_round` but `vote_period > current_pbft_period` -> this does not make sense: how can vote has older round than current_pbft_round and newer period than current_pbft_period ? 
- Standard & sync next vote -> as the rule is `vote_round >= current_pbft_round`, attacker might just keep increasing pbft round and these votes would pass validation and fill up verified queues. How to solve this ? We could put here at least some upper boundary, e.g.  `vote_round <= current_pbft_round + 1000` or some big number ?  

## Things to review properly:
I am now putting period into each placeVote_ call -> this period is then in the vote. In some votes like soft/cert votes I am using period from proposes pbft block, in other votes when we for example vote for NULL_BLOCK_HASH I am using current pbft period -> we should think about this approach if it is 100% valid !

### DIFF
--- a/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
+++ b/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
@@ -324,6 +324,7 @@ class VoteManager {
   blk_hash_t reward_votes_pbft_block_hash_;
   uint64_t last_pbft_block_cert_round_;
   std::unordered_map<vote_hash_t, std::shared_ptr<Vote>> reward_votes_;
+  std::unordered_map<addr_t, vote_hash_t> reward_votes_unique_authors_;
   mutable std::shared_mutex reward_votes_mutex_;
 
   // <PBFT round, <PBFT step, <voter address, pair<vote 1, vote 2>>>>
--- a/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
+++ b/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
@@ -137,7 +137,7 @@ class NextVotesManager {
  */
 class VoteManager {
  public:
-  VoteManager(size_t pbft_committee_size, const addr_t& node_addr, std::shared_ptr<DbStorage> db,
+  VoteManager(const addr_t& node_addr, std::shared_ptr<DbStorage> db,
               std::shared_ptr<PbftChain> pbft_chain, std::shared_ptr<FinalChain> final_chain,
               std::shared_ptr<NextVotesManager> next_votes_mgr, std::shared_ptr<KeyManager> key_manager);
   ~VoteManager();
@@ -202,8 +202,10 @@ class VoteManager {
   /**
    * @brief Add a vote to the verified votes map
    * @param vote vote
+   *
+   * @return true if vote was successfully added, otherwise false
    */
-  void addVerifiedVote(std::shared_ptr<Vote> const& vote);
+  bool addVerifiedVote(std::shared_ptr<Vote> const& vote);
 
   /**
    * @brief Check if the vote has been in the verified votes map
@@ -276,12 +278,17 @@ class VoteManager {
 
   /**
    * @brief Add last period cert vote to reward_votes_ after the cert vote voted block finalized
-   *
    * @param cert vote voted to last period PBFT block
    *
-   * @return true if add successful
+   * @return true if vote was successfully added, otherwise false
    */
   bool addRewardVote(const std::shared_ptr<Vote>& vote);
+
+  /**
+   * @param vote_hash
+   * @return true if vote_hash is already in rewards_votes, otheriwse false
+   */
+  bool isKnownRewardVote(const vote_hash_t& vote_hash);
 
   /**
    * @brief Check reward_votes_ if including all reward votes for the PBFT block
@@ -378,9 +385,6 @@ class VoteManager {
   mutable boost::shared_mutex unverified_votes_access_;
   mutable boost::shared_mutex verified_votes_access_;
 
-  const size_t pbft_committee_size_;
-  const addr_t node_addr_;
-
   std::shared_ptr<DbStorage> db_;
   std::shared_ptr<PbftChain> pbft_chain_;
   std::shared_ptr<FinalChain> final_chain_;
@@ -388,6 +392,7 @@ class VoteManager {
   std::shared_ptr<KeyManager> key_manager_;
   std::weak_ptr<Network> network_;
 
+  // TODO: this seems like something we dont we dont need
   blk_hash_t reward_votes_pbft_block_hash_;
   uint64_t last_pbft_block_cert_round_;
   std::unordered_map<vote_hash_t, std::shared_ptr<Vote>> reward_votes_;
--- a/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
+++ b/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
@@ -150,12 +150,21 @@ class VoteManager {
    * @param vote vote
    * @return true if exist
    */
-  bool voteInVerifiedMap(std::shared_ptr<Vote> const& vote);
+  bool voteInVerifiedMap(std::shared_ptr<Vote> const& vote) const;
 
   /**
-   * @brief Clear the verified votes map
+   * @brief Inserts unique vote
+   * @param vote
+   * @return true if vote was successfully inserted(it was unique) or this specific vote was already inserted, otherwise
+   * false
    */
-  void clearVerifiedVotesTable();
+  bool insertUniqueVote(const std::shared_ptr<Vote>& vote);
+
+  /**
+   * @param vote
+   * @return <true, ""> if vote is unique per round & step & voter, otherwise <false, "err msg">
+   */
+  std::pair<bool, std::string> isUniqueVote(const std::shared_ptr<Vote>& vote) const;
 
   /**
    * @brief Get all verified votes
@@ -200,6 +209,10 @@ class VoteManager {
   uint64_t roundDeterminedFromVotes(size_t two_t_plus_one);
 
   // reward votes
+  /**
+   * @return current rewards votes pbft block hash
+   */
+  blk_hash_t getCurrentRewardsVotesBlock() const;
 
   /**
    * @brief Add last period cert vote to reward_votes_ after the cert vote voted block finalized
@@ -213,7 +226,7 @@ class VoteManager {
    * @param vote_hash
    * @return true if vote_hash is already in rewards_votes, otheriwse false
    */
-  bool isKnownRewardVote(const vote_hash_t& vote_hash);
+  bool isInRewardsVotes(const vote_hash_t& vote_hash);
 
   /**
    * @brief Check reward_votes_ if including all reward votes for the PBFT block
@@ -298,8 +311,6 @@ class VoteManager {
                                           std::pair<uint64_t, std::unordered_map<vote_hash_t, std::shared_ptr<Vote>>>>>>
       verified_votes_;
 
-  std::unordered_map<addr_t, uint64_t> max_received_round_for_address_;
-
   std::unique_ptr<std::thread> daemon_;
 
   mutable boost::shared_mutex verified_votes_access_;
@@ -310,11 +321,17 @@ class VoteManager {
   std::shared_ptr<NextVotesManager> next_votes_manager_;
   std::weak_ptr<Network> network_;
 
-  // TODO: this seems like something we dont we dont need
   blk_hash_t reward_votes_pbft_block_hash_;
   uint64_t last_pbft_block_cert_round_;
   std::unordered_map<vote_hash_t, std::shared_ptr<Vote>> reward_votes_;
   mutable std::shared_mutex reward_votes_mutex_;
+
+  // <PBFT round, <PBFT step, <voter address, pair<vote 1, vote 2>>>>
+  // For next votes we enable 2 votes per round & step, one of which must be vote for NULL_BLOCK_HASH
+  std::map<uint64_t, std::unordered_map<
+                         size_t, std::unordered_map<addr_t, std::pair<std::shared_ptr<Vote>, std::shared_ptr<Vote>>>>>
+      voters_unique_votes_;
+  mutable std::shared_mutex voters_unique_votes_mutex_;
 
   LOG_OBJECTS_DEFINE
 };
--- a/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
+++ b/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
@@ -137,9 +137,9 @@ class NextVotesManager {
  */
 class VoteManager {
  public:
-  VoteManager(const addr_t& node_addr, std::shared_ptr<DbStorage> db,
-              std::shared_ptr<PbftChain> pbft_chain, std::shared_ptr<FinalChain> final_chain,
-              std::shared_ptr<NextVotesManager> next_votes_mgr, std::shared_ptr<KeyManager> key_manager);
+  VoteManager(const addr_t& node_addr, std::shared_ptr<DbStorage> db, std::shared_ptr<PbftChain> pbft_chain,
+              std::shared_ptr<FinalChain> final_chain, std::shared_ptr<NextVotesManager> next_votes_mgr,
+              std::shared_ptr<KeyManager> key_manager);
   ~VoteManager();
   VoteManager(const VoteManager&) = delete;
   VoteManager(VoteManager&&) = delete;
--- a/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
+++ b/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
@@ -90,8 +90,6 @@ class NextVotesManager {
  private:
   using UniqueLock = boost::unique_lock<boost::shared_mutex>;
   using SharedLock = boost::shared_lock<boost::shared_mutex>;
-  using UpgradableLock = boost::upgrade_lock<boost::shared_mutex>;
-  using UpgradeLock = boost::upgrade_to_unique_lock<boost::shared_mutex>;
 
   /**
    * @brief Assertion happens when there is more than 1 voting value for a non-NULL block hash
@@ -117,6 +115,7 @@ class NextVotesManager {
   LOG_OBJECTS_DEFINE
 };
 
+// TODO[1907]: refactor vote manager
 /**
  * @brief VoteManager class manage votes for PBFT consensus
  */
@@ -170,7 +169,7 @@ class VoteManager {
    * @brief Get all verified votes
    * @return all verified votes
    */
-  std::vector<std::shared_ptr<Vote>> getVerifiedVotes();
+  std::vector<std::shared_ptr<Vote>> getVerifiedVotes() const;
 
   /**
    * @brief Get the total size of all verified votes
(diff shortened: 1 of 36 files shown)


justinsnapp (commented):
@JakubFornadel  thanks for the excellent write-up on this PR.     I think everything you have done should definitely allow us to remove the queue for unvalidated votes.    But perhaps it still allows for a node to send us too many validatable votes, by sending us votes for wildly in the future rounds?   It seems a node can still can send votes from many rounds in the future, even if they get to speak only sometimes by churning through lots of rounds they could still amass a lot of votes to send?   

Unless we are out of sync they should be at most one round ahead of us.   So any node sending us votes for a round far ahead of local nodes round, the local node could request to sync to their round by getting 2t+1 next votes from previous round or else consider such a node malicious ... ie by getting a sync next vote packet.    Alteneratively since we send status packets (still), we could put our step and round in that and if the round was "wildly" (allowing for the status packet update interval) behind the local node we would send them a sync next votes packet and not relay votes way ahead.  And on the receiving side if we got votes for a round "wildly" ahead of our round we would mark that node as malicious.

You mentioned being a bit hazy as how to define the correct period for vote for each round to send.  I think the proper definition is that it **should be that it doesn't change during the round... so should be the same throughout the round, and be the period at the start of the round, before any possible growth in the chain by commit of a new pbft block**

JakubFornadel:
> @JakubFornadel thanks for the excellent write-up on this PR. I think everything you have done should definitely allow us to remove the queue for unvalidated votes. But perhaps it still allows for a node to send us too many validatable votes, by sending us votes for wildly in the future rounds? It seems a node can still can send votes from many rounds in the future, even if they get to speak only sometimes by churning through lots of rounds they could still amass a lot of votes to send?
> 
> Unless we are out of sync they should be at most one round ahead of us. So any node sending us votes for a round far ahead of local nodes round, the local node could request to sync to their round by getting 2t+1 next votes from previous round or else consider such a node malicious ... ie by getting a sync next vote packet. Alteneratively since we send status packets (still), we could put our step and round in that and if the round was "wildly" (allowing for the status packet update interval) behind the local node we would send them a sync next votes packet and not relay votes way ahead. And on the receiving side if we got votes for a round "wildly" ahead of our round we would mark that node as malicious.
> 
> You mentioned being a bit hazy as how to define the correct period for vote for each round to send. I think the proper definition is that it **should be that it doesn't change during the round... so should be the same throughout the round, and be the period at the start of the round, before any possible growth in the chain by commit of a new pbft block**

-  Yes the problem with future rounds/steps is still not resolved but that could separate PR. This PR does not make the node more vulnerable to any new attack compared to previous version so even if we merge this it is going to be huge improvement & Dmitry can resolve conflicts while we work on fix for future rounds & steps. 
- I actually used this design before and the problem I had was that at the beginning of a new round I set period based on node pbft chain size, which is a problem because nodes might have different chain sizes... When I think about it now maybe we could set it based on periods from next votes. Currently we determine round based on 2t+1 next votes, what if we dont check just rounds, but also periods from these 2t+1 next votes and determine  round & period based on that ?

kstdl (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1110,14 +1123,50 @@ blk_hash_t PbftManager::generatePbftBlock(const blk_hash_t &prev_blk_hash, const
...
+      err << "No vrf key mapped for vote author " << vote->getVoterAddr();
+      return {false, err.str()};
+    }
+
+    // TODO[1896]: final_chain_->dpos... ret values should be cached as it is heavily used and most of the time it
+    // returns the same value!
+    const uint64_t voter_dpos_votes_count = final_chain_->dpos_eligible_vote_count(vote_period, vote->getVoterAddr());
+    const uint64_t total_dpos_votes_count = final_chain_->dpos_eligible_total_vote_count(vote_period);
+    const uint64_t pbft_sortition_threshold = getPbftSortitionThreshold(vote->getType(), vote_period);
+
```
MatusKysel, line 1150: I think this should be called only when `vote_period != dpos_period_` as we currently have some kind of caching we should use it

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1129,17 +1178,56 @@ uint64_t PbftManager::getThreshold(PbftVoteTypes vote_type) const {
...
+  switch (vote_type) {
+    case propose_vote_type:
+      return std::min<uint64_t>(NUMBER_OF_PROPOSERS, total_dpos_votes_count);
+    case soft_vote_type:
+    case cert_vote_type:
+    case next_vote_type:
+    default:
+      return std::min<uint64_t>(COMMITTEE_SIZE, total_dpos_votes_count);
+  }
+}
```
MatusKysel, line 1193: Having getCurrentPbftSortitionThreshold and getPbftSortitionThreshold does not make sense as it shares 90% of code logic, this could lead to errors where fix switch case just for one of the functions

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1129,17 +1178,56 @@ uint64_t PbftManager::getThreshold(PbftVoteTypes vote_type) const {
...
+  uint64_t total_dpos_votes_count = 0;
+  uint64_t pbft_sortition_threshold = 0;
+
+  try {
+    // TODO[1896]: final_chain_->dpos... ret values should be cached as it is heavily used and most of the time it
+    // returns the same value!
+    voter_dpos_votes_count = final_chain_->dpos_eligible_vote_count(period, node_addr_);
+    total_dpos_votes_count = final_chain_->dpos_eligible_total_vote_count(period);
+    pbft_sortition_threshold = getPbftSortitionThreshold(vote_type, period);
+
```
MatusKysel, line 1212: again this could be check if it is not already pre-cached

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1398,17 +1485,18 @@ blk_hash_t PbftManager::identifyLeaderBlock_() {
       continue;
     }
 
-    leader_candidates.emplace_back(std::make_pair(getProposal(v), proposed_block_hash));
+    leader_candidates.emplace_back(std::make_pair(getProposal(v), v));
```
MatusKysel, line 1488: why don't just use `getPeriod()` ? this should be more efficient than whole vote object

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1628,7 +1716,6 @@ void PbftManager::pushSyncedPbftBlocksIntoChain() {
       net->setSyncStatePeriod(period);
 
       if (executed_pbft_block_) {
-        vote_mgr_->removeVerifiedVotes();
```
MatusKysel, line 1631: we do not need to clean those votes?

libraries/core_libs/network/src/network.cpp
```diff
@@ -13,32 +13,33 @@
 
 namespace taraxa {
 
-Network::Network(NetworkConfig const &config, const h256 &genesis_hash,
+Network::Network(const FullNodeConfig &config, const h256 &genesis_hash,
```
MatusKysel, line 16: sorry I missed why did you change this? I did not see where it is used

tests/pbft_manager_test.cpp
```diff
@@ -397,24 +383,8 @@ TEST_F(PbftManagerTest, check_get_eligible_vote_count) {
...
-      expected_eligible_total_vote++;
-    }
-    EXPECT_HAPPENS({120s, 1s}, [&](auto &ctx) {
-      for (auto &node : nodes) {
-        if (ctx.fail_if(node->getDB()->getNumTransactionExecuted() != trxs_count)) {
-          return;
-        }
-      }
-    });
-  }
```
MatusKysel, line 417: same here

libraries/types/vote/include/vote/vrf_sortition.hpp
```diff
@@ -17,44 +17,35 @@ enum PbftVoteTypes : uint8_t { propose_vote_type = 0, soft_vote_type, cert_vote_
 /**
  * @brief VrfPbftMsg struct uses vote type, PBFT round, and PBFT step to generate a message for doing VRF sortition.
  */
-struct VrfPbftMsg {
+class VrfPbftMsg {
```
MatusKysel, line 20: if you changing it to class you should also rename member variables

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
```diff
@@ -335,39 +271,35 @@ class VoteManager {
   using UpgradableLock = boost::upgrade_lock<boost::shared_mutex>;
   using UpgradeLock = boost::upgrade_to_unique_lock<boost::shared_mutex>;
```
MatusKysel, line 272: you should remove those as we are no longer using them

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -480,102 +403,34 @@ uint64_t VoteManager::roundDeterminedFromVotes(size_t two_t_plus_one) {
...
-  if (!verifyRewardVote(vote)) {
-    return false;
-  }
-
-  {
-    std::unique_lock lock(reward_votes_mutex_);
-    reward_votes_.insert({vote_hash, vote});
-  }
-
-  db_->saveLastBlockCertVote(vote);
```
MatusKysel, line 526: If you do not use it anywhere we should remove this function too

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -480,102 +403,34 @@ uint64_t VoteManager::roundDeterminedFromVotes(size_t two_t_plus_one) {
...
-    LOG(log_er_) << e.what();
-    LOG(log_er_) << "DPOS period " << dpos_period << ", voter " << voter_account_addr << " votes count "
-                 << voter_dpos_votes_count << ", total votes count " << dpos_total_votes_count
-                 << ", pbft sortition threshold " << pbft_sortition_threshold;
-
-    return false;
-  }
-
-  return true;
+bool VoteManager::isInRewardsVotes(const vote_hash_t& vote_hash) {
```
MatusKysel, line 431: const ?

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
```diff
@@ -152,70 +136,35 @@ class VoteManager {
...
-  bool voteInVerifiedMap(std::shared_ptr<Vote> const& vote);
+  bool voteInVerifiedMap(std::shared_ptr<Vote> const& vote) const;
+
+  /**
+   * @brief Inserts unique vote
+   * @param vote
+   * @return true if vote was successfully inserted(it was unique) or this specific vote was already inserted, otherwise
+   * false
+   */
+  bool insertUniqueVote(const std::shared_ptr<Vote>& vote);
```
MatusKysel, line 161: Why `insertUniqueVote` is separate function from `addVerifiedVote` ? is there a case where would you call just one? Because what I can see is that someone will call just one of those and break whole functionality

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -2,17 +2,147 @@
...
+      vote_mgr_(std::move(vote_mgr)) {}
+
+std::pair<bool, std::string> ExtVotesPacketHandler::validateStandardVote(const std::shared_ptr<Vote> &vote) const {
+  const uint64_t current_pbft_period = pbft_chain_->getPbftChainSize();
+  const auto current_pbft_round = pbft_mgr_->getPbftRound();
+
+  // Old vote or vote from too far in the future, can be dropped
+  // TODO[1880]: should be vote->getPeriod() <= current_pbft_period - if <=, some tests are failing due to missing
+  // reward votes -> whole rewards votes gossiping need to be checked...
+  if (vote->getPeriod() < current_pbft_period || vote->getPeriod() > current_pbft_period + kDposDelay_) {
```
MatusKysel, line 28: `kDposDelay_` should be without _

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -47,7 +176,7 @@ void ExtVotesPacketHandler::sendPbftVotes(const dev::p2p::NodeID &peer_id, std::
     dev::RLPStream s(count);
     for (auto i = index; i < index + count; i++) {
       const auto &v = votes[i];
-      s.appendRaw(v->rlp(true));
+      s.appendRaw(v->rlp(true, true));
```
MatusKysel, line 179: why are you include weight?

MatusKysel (approved):

tests/pbft_manager_test.cpp
```diff
@@ -46,22 +46,6 @@ void check_2tPlus1_validVotingPlayers_activePlayers_threshold(size_t committee_s
...
-      trxs_count++;
-    }
-    EXPECT_HAPPENS({120s, 1s}, [&](auto &ctx) {
-      for (auto &node : nodes) {
-        if (ctx.fail_if(node->getDB()->getNumTransactionExecuted() != trxs_count)) {
-          return;
-        }
-      }
-    });
-  }
```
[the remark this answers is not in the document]
JakubFornadel: - main purpose of function `check_2tPlus1_validVotingPlayers_activePlayers_threshold` is to test dpos transactions ? Either something is terribly wrong with the function name or the main purpose is something totally different
- if we wanna be sure dpos transactions works, there should be separate test for it

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: https://github.com/Taraxa-project/taraxa-node/issues/1896

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: will be solved by https://github.com/Taraxa-project/taraxa-node/issues/1896

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: will be solved by https://github.com/Taraxa-project/taraxa-node/issues/1896

tests/pbft_manager_test.cpp
JakubFornadel: https://github.com/Taraxa-project/taraxa-node/pull/1900#discussion_r939946599

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: not anymore. Purpose of this function: @brief Move all verified votes to the unverified map. Since PBFT chain has pushed new blocks, that will affect DPOS. I deleted this function

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
JakubFornadel: yeah well it was used just here, but I have no idea why... Any possible clue why we would save block cert vote when adding reward vote ?

libraries/core_libs/network/src/network.cpp
JakubFornadel: because in multiple packet handlers I am using "delegation_delay" to know if it even makes sense to process e.g. votes and validate them against dpos contract

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
JakubFornadel: because the node did not work anymore due to some assert related to empty weight, but I did not spend too much tie thinking about other possible fixes so I can revisit this one

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
MatusKysel: So please add TODO as we agreed for all things that will be solved ..

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
JakubFornadel: the reason is that `insertUniqueVote` is called also when adding reward votes and next votes in NextVotesManager. The other solution is to separate this unique votes functionality into separate class and call it inside `addVerifiedVote`, `addRewardVote` and `updateWithSyncedVotes`

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: sure, will do

tests/pbft_manager_test.cpp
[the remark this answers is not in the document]
MatusKysel: sorry, I meant that idea of that test was to increase number of eligible nodes in runtime not statistically on start

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
MatusKysel: I guess it was because it could be different cert vote we did not have saved before ?

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
JakubFornadel: https://github.com/Taraxa-project/taraxa-node/issues/1935

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -346,39 +348,46 @@ void PbftManager::resetStep_() {
...
 
 bool PbftManager::resetRound_() {
-  // Jump to round
-  auto consensus_pbft_round = vote_mgr_->roundDeterminedFromVotes(TWO_T_PLUS_ONE);
+  auto determined_round_and_period = vote_mgr_->determineRoundAndPeriodFromVotes(TWO_T_PLUS_ONE);
+  if (!determined_round_and_period.has_value()) {
+    return false;
+  }
+
+  auto [determined_round, determined_previous_round_period] = determined_round_and_period.value();
```
MatusKysel, line 356: if you do check for optional value, do not use `.value()` but * as it has perf improvements

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -509,39 +524,40 @@ void PbftManager::initialState() {
   pbft_step_last_requested_sync_ = 0;
 
   auto own_starting_value = db_->getPbftMgrVotedValue(PbftMgrVotedValue::OwnStartingValueInRound);
-  if (own_starting_value) {
+  if (own_starting_value.has_value()) {
     // From DB
-    own_starting_value_for_round_ = *own_starting_value;
+    own_starting_value_for_round_ = own_starting_value.value();
```
MatusKysel, line 529: again, this is wrong. if there is check for the value, never use `.value()`

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -666,84 +682,100 @@ bool PbftManager::stateOperations_() {
...
-    }
-    try {
-      v->validate(dposEligibleVoteCount_(v->getVoterAddr()), getDposTotalVotesCount(), getThreshold(v->getType()), *pk);
-    } catch (const std::logic_error &e) {
-      LOG(log_wr_) << e.what();
-      return false;
+  auto [round, previous_round_period] = getPbftRoundAndPeriod();
+  LOG(log_dg_) << "PBFT current round(r): " << round << ", r-1 period: " << previous_round_period << ", step " << step_;
+
+  static bool is_new_round = false;
```
MatusKysel, line 688: why static?

libraries/core_libs/network/src/network.cpp
```diff
@@ -13,32 +13,33 @@
...
                  std::filesystem::path const &network_file_path, dev::KeyPair const &key, std::shared_ptr<DbStorage> db,
                  std::shared_ptr<PbftManager> pbft_mgr, std::shared_ptr<PbftChain> pbft_chain,
                  std::shared_ptr<VoteManager> vote_mgr, std::shared_ptr<NextVotesManager> next_votes_mgr,
                  std::shared_ptr<DagManager> dag_mgr, std::shared_ptr<TransactionManager> trx_mgr)
-    : kConf(config), tp_(config.network.network_num_threads, false) {
+    : tp_(config.network.network_num_threads, false) {
   auto const &node_addr = key.address();
   LOG_OBJECTS_CREATE("NETWORK");
-  LOG(log_nf_) << "Read Network Config: " << std::endl << kConf.network << std::endl;
+  LOG(log_nf_) << "Read Network Config: " << std::endl << config.network << std::endl;
```
MatusKysel, line 25: why did you renamed constant member?

mfrankovi (approved):

libraries/core_libs/network/src/network.cpp
JakubFornadel: it is not a member anymore, it is ctor argument

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: because I need it to hold the value throughout multiple calls. I could either have class member or static variable - as this variable is needed only in this single function I used static rather than class member

MatusKysel: ok, but I do not think it is good idea, as now you have some state, that it is shared between all instances of pbft manager and it should be bind to specific instance not to all. Doesn't it influence tests?

JakubFornadel: true, fixed

PR #1902, opened 2022-07-29
rewrite config initial phase in python

rjonczy:
## Purpose

Generate config for non-standard networks in python (initial phase)

Changed from bash into python part of config generation.

## Changes made

Changed config generation for:
- initial validators

### DIFF
--- a/charts/taraxa-node/templates/initconfig-consensus-node.yaml
+++ b/charts/taraxa-node/templates/initconfig-consensus-node.yaml
@@ -11,6 +11,85 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
 data:
+  genconfig.py: |-
+    import json
+    import sys
+    import subprocess
+
+    def get_vrf_public(vrf_prv_key):
+        process = subprocess.Popen(['taraxad', '--command', 'vrf', vrf_prv_key],stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+
+        for line in process.stdout:
+            l = line.decode("utf-8")
+
+            if "vrf_public" in l:
+                vrf_public = l.split(':')[1].replace("\"", "").strip()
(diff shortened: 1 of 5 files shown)


charts/taraxa-node/templates/initconfig-consensus-node.yaml
```diff
@@ -11,6 +11,85 @@ metadata:
     app.kubernetes.io/instance: {{ .Release.Name }}
     app.kubernetes.io/managed-by: {{ .Release.Service }}
 data:
+  genconfig.py: |-
+    import json
+    import sys
+    import subprocess
+
+    def get_vrf_public(vrf_prv_key):
```
MatusKysel, line 19: vrf public key is just a part of private key, so you could just split private key. 
```
	"vrf_public" : "42f6022cce105e9f251e8054378cf1f435ea1e7bd0ba12b5b22c48a8c68d7db0",
	"vrf_secret" : "84117a0e78a6caadf2ba67414d4310136feddff3abbf7acc80b6ed9912a8ec8942f6022cce105e9f251e8054378cf1f435ea1e7bd0ba12b5b22c48a8c68d7db0"
```

you can see that public key private_key[64:]

MatusKysel (approved):

charts/taraxa-node/templates/initconfig-consensus-node.yaml
rattrap: Nice. Didn't know that

rattrap (approved):
LGTM

charts/taraxa-node/templates/initconfig-consensus-node.yaml
rjonczy: @MatusKysel thanks. Now i merge this as I have another issue:

https://github.com/Taraxa-project/taraxa-node/issues/1906

In that issue I will completely change config generation into python and change the way how i obtain public key.

PR #1903, opened 2022-08-01
Minor fixes

kstdl:
During some tests I found a few issues:
1. In `local-net` script `address` and `vrf_key` was reordering sometimes and was leading to incorrect behaviour. For example if that was happening for 1 boot-node then it just can't finalize blocks because of not enought votes(incorrect address was specified in config). If few became incorrect then it wasn't starting because of `Validator already exist` error
2. Found that `str2bytes` was working incorrectly in my new test, so removed this and replaced it with `dev::fromHex` method that we already have. Did the same thing with `bytes2str` just because we already have `dev::toHex`

### DIFF
--- a/tests/util_test/node_dag_creation_fixture.hpp
+++ b/tests/util_test/node_dag_creation_fixture.hpp
@@ -53,7 +53,7 @@ struct NodeDagCreationFixture : BaseTest {
   }
 
   void deployContract() {
-    auto trx = std::make_shared<Transaction>(nonce, 100, 0, TEST_TX_GAS_LIMIT,
+    auto trx = std::make_shared<Transaction>(nonce, 0, 0, TEST_TX_GAS_LIMIT,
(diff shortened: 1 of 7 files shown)


tests/util_test/node_dag_creation_fixture.hpp
```diff
@@ -53,7 +53,7 @@ struct NodeDagCreationFixture : BaseTest {
   }
 
   void deployContract() {
-    auto trx = std::make_shared<Transaction>(nonce, 100, 0, TEST_TX_GAS_LIMIT,
+    auto trx = std::make_shared<Transaction>(nonce, 0, 0, TEST_TX_GAS_LIMIT,
```
MatusKysel, line 56: why did you remove value? isn't it payable contract?

mfrankovi (approved):

PR #1908, opened 2022-08-02
chore: light node test improvement

mfrankovi:
Light node test was sometimes failing. I have found and fixed two such issues and now it is passing 100/100 on my machine. I did not find anything wrong with the loght node implementation, only with the test itself.
Issues fixed:
1. Due to small light_node_history in test sometimes data would be deleted before it was synced to the other node. To prevent this transactions are now inserted into a non-light node and light node is the one that is syncing
2. There was a race condition between creating a new pbft block and deleting old blocks in a light node. Fix is simply allowing one extra pbft block to be present in a light node by increasing the limit that we verify in line 1609

### DIFF
--- a/tests/full_node_test.cpp
+++ b/tests/full_node_test.cpp
@@ -1576,12 +1576,12 @@ TEST_F(FullNodeTest, transaction_validation) {
 
 TEST_F(FullNodeTest, light_node) {
   auto node_cfgs = make_node_cfgs<10>(2);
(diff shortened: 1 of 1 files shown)


MatusKysel (approved):

kstdl (approved):

justinsnapp:
@mfrankovi maybe mostly for posterity's sake but also for people following along on your great work you should make more clear in the PR that these were only changes to the test and not to the light node implementation.

Issue #1885, opened 2022-07-25
New transaction gossip proposal


MatusKysel:
This is related to [#1869](https://github.com/Taraxa-project/taraxa-node/issues/1869). If we are planning to limit pool size we should improve transaction gossiping. Main idea is to have new way to ask for missing transaction and drop transaction with lower fees from pool when it is full

Issue #1869, opened 2022-07-18
Transaction gossip overload


mfrankovi:
Transaction gossip depends on the known transactions cache for each peer. If transactions pool grows larger that known transactions cache same transactions will be sent over and over again spamming entire network.

Quick fix on teestnet was increasing the cache. Smarter solution is needed.

MatusKysel:
Transaction pool improvements proposal:
- introducing limit per account (e.g. one account can only have 30% of total size of the pool) 
- introducing expiration time limit (e.g. transaction older then 3h hours, should be removed)
- known transactions per peer object will depend on pool size (e.g. if pool size is 30k then known cache could be like 110% so 33k)

PR #1912, opened 2022-08-04
Limit trx pool

MatusKysel:
## Purpose
Fixes: #1885  #1869

### DIFF
--- a/libraries/common/include/common/constants.hpp
+++ b/libraries/common/include/common/constants.hpp
@@ -26,6 +26,8 @@ const uint32_t kPeriodicEventsThreadCount{2};
 
 const uint64_t kMinTxGas{21000};
 
+constexpr uint32_t kMinTransactionPoolSize{30000};
+
 // The various denominations; here for ease of use where needed within code.
 static const u256 kOneTara = dev::exp10<18>();
 // static const u256 kFinney = exp10<15>();
--- a/libraries/common/include/common/constants.hpp
+++ b/libraries/common/include/common/constants.hpp
@@ -27,6 +27,7 @@ const uint32_t kPeriodicEventsThreadCount{2};
 const uint64_t kMinTxGas{21000};
 
 constexpr uint32_t kMinTransactionPoolSize{30000};
+constexpr uint32_t kDefaultTransactionPoolSize{200000};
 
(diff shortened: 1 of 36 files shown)


libraries/common/include/common/constants.hpp
```diff
@@ -26,6 +26,8 @@ const uint32_t kPeriodicEventsThreadCount{2};
 
 const uint64_t kMinTxGas{21000};
 
+constexpr uint32_t kMinTransactionPoolSize{30000};
```
mfrankovi, line 29: For any kind of high load tps testing this should be much bigger. If we are going for 10k tps, my recommendation would be that max size of the pool should be at least 200k. @justinsnapp what do you think?

mfrankovi: Also maybe we should have two constants, one for default value and one for min value, min can stay 30k but for default I would put 200k

MatusKysel: That's the idea, right now 30k is min but we can specify any above that

JakubFornadel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
```diff
@@ -104,7 +104,12 @@ void DagBlockPacketHandler::onNewBlockReceived(DagBlock &&block, const std::shar
         throw MaliciousPeerException(err_msg.str());
       }
       case DagManager::VerifyBlockReturnType::MissingTransaction:
-        if (peer->peer_dag_synced_) {
+        if (trx_mgr_->isTransactionPoolFull()) [[unlikely]] {
```
mfrankovi, line 107: I think in some edge cases we might not be full but still get into this state because we were recently full. Maybe we should do this even if we are 80% or 90% full?

kstdl (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/dag_block_packet_handler.cpp
MatusKysel: good point, let's set it to 90%

mfrankovi (approved):

PR #1913, opened 2022-08-04
fix: corrupt reward votes

mfrankovi:
Votes need to be verified for finalization but if the period is already pushed to pbft chain reward votes should not be added to vote manager because this corrupts the node since it adds reward votes for previous period while at the same time vote manager can already contain reward votes from current period which than incorrectly mixes votes from multiple periods.

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -119,12 +119,9 @@ void PbftManager::run() {
     }
     // We need this section because votes need to be verified for reward distribution
     for (const auto &v : period_data.previous_block_cert_votes) {
-      vote_mgr_->addRewardVote(v);
+      vote_mgr_->verifyRewardVote(v);
     }
-    if (!vote_mgr_->checkRewardVotes(period_data.pbft_blk)) {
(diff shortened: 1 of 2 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -119,12 +119,9 @@ void PbftManager::run() {
     }
     // We need this section because votes need to be verified for reward distribution
     for (const auto &v : period_data.previous_block_cert_votes) {
-      vote_mgr_->addRewardVote(v);
-    }
-    if (!vote_mgr_->checkRewardVotes(period_data.pbft_blk)) {
-      LOG(log_er_) << "Invalid reward votes in block " << period_data.pbft_blk->getBlockHash() << " in DB.";
-      assert(false);
+      vote_mgr_->verifyRewardVote(v);
```
kstdl, line 122: Shouldn't we add vote after this verification?

MatusKysel (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: no, the reason for verification is that verification sets some internal structure within the votes, vote manager holds reward votes for the pbft block that is not yet part of the chain and this is not the case, here pbft block is already pushed to chain just not finalized/executed

kstdl (approved):

Issue #1905, opened 2022-08-01
Fix RPC calls


JakubFornadel:
#### 1st part
The RPC Server implementation lacks authentication or level-based access
control, allow anyone with network level access to create raw transactions
via one of the following RPC calls
- eth_sendTransaction
- eth_call
- eth_sendRawTransaction
- send_coin_transaction

#### 2nd part
Get rid of Test.jsonrpc.json file - we can either put it into the Taraxa.jsonrpc.json file if those RPC are useful or somehow make it possible to disable test rpcs ? 

## Solution

- Disable eth_sendTransaction as it is using node's secret to sign tx on node - either return "unimplemented method" or "deprecated method" err  
- Discuss (in team as well as with halborn) why the other calls need authentication if they just read some data from node ?

PR #1914, opened 2022-08-04
Fix rpc calls

kstdl:
#1905 
1. Removed `eth_sendTransaction` method. 
2. Added flag `--enable-test-rpc` to node which enables Test API that is now disabled by default.

### DIFF
--- a/libraries/core_libs/network/rpc/eth/Eth.cpp
+++ b/libraries/core_libs/network/rpc/eth/Eth.cpp
@@ -81,16 +81,6 @@ class EthImpl : public Eth, EthParams {
 
   Json::Value eth_getUncleCountByBlockNumber(string const&) override { return toJS(0); }
 
-  string eth_sendTransaction(Json::Value const& _json) override {
-    auto t = toTransactionSkeleton(_json);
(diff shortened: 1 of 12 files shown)


libraries/core_libs/network/rpc/eth/Eth.cpp
```diff
@@ -81,16 +81,6 @@ class EthImpl : public Eth, EthParams {
 
   Json::Value eth_getUncleCountByBlockNumber(string const&) override { return toJS(0); }
 
-  string eth_sendTransaction(Json::Value const& _json) override {
-    auto t = toTransactionSkeleton(_json);
-    set_transaction_defaults(t, final_chain->last_block_number());
-    auto trx = std::make_shared<Transaction>(t.nonce.value_or(0), t.value, t.gas_price.value_or(0), t.gas.value_or(0),
-                                             t.data, secret, t.to ? optional(t.to) : nullopt, chain_id);
```
MatusKysel, line 88: I suggest to remove `secret` from eth struct

MatusKysel (approved):

JakubFornadel (approved):

PR #1919, opened 2022-08-08
chore: verify reward votes round

mfrankovi:
When verifying reward vote verify that this is a cert vote which has the same round as the round that previous pbft block pushed in the chain was cert voted on

### DIFF
--- a/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
+++ b/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
@@ -483,6 +483,7 @@ uint64_t VoteManager::roundDeterminedFromVotes(size_t two_t_plus_one) {
 bool VoteManager::addRewardVote(const std::shared_ptr<Vote>& vote) {
   const auto vote_hash = vote->getHash();
   const auto voted_block_hash = vote->getBlockHash();
+  const auto vote_round = vote->getRound();
   const auto vote_step = vote->getStep();
   const auto vote_type = vote->getType();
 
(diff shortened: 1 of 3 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1679,6 +1679,17 @@ bool PbftManager::pushPbftBlock_(PeriodData &&period_data, std::vector<std::shar
...
+  // It is not possible to have apbft block without cert votes
+  assert(cert_votes.size());
+  const auto first_vote_round = cert_votes[0]->getRound();
+  for (uint64_t i = 1; i < cert_votes.size(); i++) {
+    if (first_vote_round != cert_votes[i]->getRound()) {
+      LOG(log_er_) << "Cert vote " << cert_votes[i]->getHash() << " round " << cert_votes[i]->getRound()
+                   << " mismatching round with: " << cert_votes[0]->getHash() << "" << first_vote_round;
+      return false;
+    }
+  }
```
MatusKysel, line 1691: this is already done one cert votes validation no need to have it here

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -533,13 +533,20 @@ bool VoteManager::verifyRewardVote(const std::shared_ptr<Vote>& vote) {
...
     LOG(log_dg_) << "Drop vote " << vote_hash << ". PBFT chain last block is " << pbft_chain_last_block_hash
                  << ", but vote for PBFT block is " << voted_block_hash;
     return false;
   }
 
+  if (vote->getRound() != pbft_chain_last_block_round) {
+    LOG(log_dg_) << "Drop vote " << vote_hash << ". PBFT chain last block round is " << pbft_chain_last_block_round
+                 << ", but vote for PBFT block is round " << vote->getRound();
+    return false;
+  }
```
MatusKysel, line 548: and this should be assert as this can not happened or? how we can have different rounds? then it is not cert vote

MatusKysel (approved):

JakubFornadel (approved):

PR #1930, opened 2022-08-11
chore: allow rewards votes with different rounds

mfrankovi:
As noticed on devnet although this happens rarely it is possible that the same pbft block can be cert voted and pushed into chain on different nodes in different rounds. This means that it is possible that reward votes for next proposed pbft chain which are cert voted of the previous block can be in different rounds.

The changes in this PR allow us to save and gossip reward votes from different rounds as long as they point to the correct hash of the previous pbft block.

However on proposing a pbft block we only choose to include reward votes from the round that the proposer has cert voted previous block on. So reward votes included in pbft block will always have the same round.

### DIFF
--- a/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
+++ b/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
@@ -507,9 +507,9 @@ bool VoteManager::addRewardVote(const std::shared_ptr<Vote>& vote) {
     }
 
     if (reward_votes_pbft_block_round_ != vote_round) {
-      LOG(log_wr_) << "The reward vote " << vote_hash << " round " << vote_round << " is different than "
+      // This can happen if some nodes cert voted previous block in different rounds
+      LOG(log_dg_) << "The reward vote " << vote_hash << " round " << vote_round << " is different than "
                    << reward_votes_pbft_block_round_;
-      return false;
     }
 
     if (reward_votes_.contains(vote_hash)) {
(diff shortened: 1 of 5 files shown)


libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -610,6 +610,10 @@ void VoteManager::replaceRewardVotes(std::vector<std::shared_ptr<Vote>>&& cert_v
   std::unique_lock lock(reward_votes_mutex_);
   reward_votes_ = {};
   reward_votes_pbft_block_hash_ = cert_votes[0]->getBlockHash();
+
+  // It is possible that incoming reward votes might have another round because it is possible that same block was cert
+  // voted in different rounds on different nodes but this is a reference round for any pbft block this node might
+  // propose
   reward_votes_pbft_block_round_ = cert_votes[0]->getRound();
```
MatusKysel, line 617: do we need this ? I would remove it as it does nothing anymore

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -628,6 +632,35 @@ std::vector<std::shared_ptr<Vote>> VoteManager::getRewardVotes() {
...
+std::vector<std::shared_ptr<Vote>> VoteManager::getRewardVotes(const std::vector<vote_hash_t>& vote_hashes) {
+  std::vector<std::shared_ptr<Vote>> reward_votes;
+
+  std::shared_lock lock(reward_votes_mutex_);
+  for (auto vote_hash : vote_hashes) {
+    auto it = reward_votes_.find(vote_hash);
+    if (it != reward_votes_.end()) {
+      reward_votes.emplace_back(it->second);
+    } else {
+      LOG(log_dg_) << "Missing reward vote: " << vote_hash;
```
MatusKysel, line 644: I would return immediately with empty array. No need to continue

MatusKysel (approved):

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: Fixed

kstdl (approved):

JakubFornadel (commented):
@mfrankovi I dont think this will work. If I push block at round 1 and save cert votes from this round as reward votes and others push this block at round 2 - they will save cert votes from round 2 so they dont have cert votes from round 1. Then when I try to regossip my reward votes(cert votes from round 1) it is ignored as these votes are marked as already known for others...

PR #1944, opened 2022-08-16
chore: next vote after block pushed

mfrankovi:
Not giving up on next voting a value that is already cert voted and pushed to the chain


JakubFornadel:
So the actual bug here is this condition because we wrongly set last_cert_voted_value_ to NULL_BLOCK_HASH right after we push block into the chain....
`if (last_cert_voted_value_ != NULL_BLOCK_HASH) { `

in first finish step. This is fixed in my branch so I think it will fix the issue

justinsnapp:
@mfrankovi  this PR is outdated, having been superseded by all the pbft simplification changes.  So I will close this PR.

PR #1949, opened 2022-08-17
chore: on null block vote use current period

mfrankovi:


justinsnapp (commented):
I don't think the term "propose period" is appropriate outside of the proposal step.   Seems like this is just "current period" or vote period.

MatusKysel (approved):

kstdl (approved):

PR #1953, opened 2022-08-18
chore: unique cert value in period

mfrankovi:
Only reset cert vote value if it is pushed in chain.
Do not give up on cert vote value

### DIFF
--- a/tests/pbft_chain_test.cpp
+++ b/tests/pbft_chain_test.cpp
@@ -85,7 +85,7 @@ TEST_F(PbftChainTest, pbft_db_test) {
 }
 
 TEST_F(PbftChainTest, proposal_block_broadcast) {
-  auto node_cfgs = make_node_cfgs<5>(3);
+  auto node_cfgs = make_node_cfgs<1>(3);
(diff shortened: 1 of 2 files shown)


MatusKysel (approved):

justinsnapp (approved):
I guess @JakubFornadel  should weigh in on why this was changed to be this way, but certainly to me this looks like a restoration to correct behavior.

tests/pbft_chain_test.cpp
```diff
@@ -85,7 +85,7 @@ TEST_F(PbftChainTest, pbft_db_test) {
 }
 
 TEST_F(PbftChainTest, proposal_block_broadcast) {
-  auto node_cfgs = make_node_cfgs<5>(3);
+  auto node_cfgs = make_node_cfgs<1>(3);
```
justinsnapp, line 88: So compared to @MatusKysel 's change of just adding more time to our weight before doing an equality check within the test, here we are just running this test in "real time" as opposed to 5x acceleration.    Thats fine, but I am curious as to what changed to make broadcast "slower"?

PR #1956, opened 2022-08-19
Fix different periods

JakubFornadel:
## Purpose

Always use chain_size + 1 as period in any next vote. Thanks to this we should be able to determine period in new round:

- if node sees 2t+1 next votes with period 10 in round 20, it moves to the round 21 with determined period 10. Node's pbft consensus is blocked until it's chain size is >= 9...

### Potential problems:
if half of the nodes pushed block into the chain in round 20 and other half did not, 50% of the next votes will have period N and other 50% N+1, which might cause the network to not advance into the next round until some of the 50% of nodes that did not push block sync up and create next vote with N+1 period so we will actually have 2t+1 next votes with some value...

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -922,8 +922,8 @@ void PbftManager::identifyBlock_() {
   if (auto vote_weight = placeVote_(previous_round_next_voted_value_.first, soft_vote_type,
                                     previous_round_next_voted_value_.second, round, step_);
       vote_weight) {
-    LOG(log_nf_) << "Placed soft vote for block from previous round " << previous_round_next_voted_value_
-                 << ", vote weight " << vote_weight << ", round " << round << ", period " << period - 1 << ", step "
+    LOG(log_nf_) << "Placed soft vote for block from previous round " << previous_round_next_voted_value_.first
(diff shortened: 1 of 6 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -915,29 +914,29 @@ void PbftManager::identifyBlock_() {
...
 
-  if (auto vote_weight =
-          placeVote_(previous_round_next_voted_value_, soft_vote_type, previous_round_period, round, step_);
+  if (auto vote_weight = placeVote_(previous_round_next_voted_value_.first, soft_vote_type,
+                                    previous_round_next_voted_value_.second, round, step_);
       vote_weight) {
     LOG(log_nf_) << "Placed soft vote for block from previous round " << previous_round_next_voted_value_
-                 << ", vote weight " << vote_weight << ", round " << round << ", period " << previous_round_period
-                 << ", step " << step_;
+                 << ", vote weight " << vote_weight << ", round " << round << ", period " << period - 1 << ", step "
```
MatusKysel, line 926: period - 1 ? not `previous_round_next_voted_value_.second`

MatusKysel (approved):

justinsnapp (approved):

kstdl (approved):

Issue #1947, opened 2022-08-17
Votes DDOS protection


kstdl:
We should limit processing of future votes. Votes with `vote_round` in it bigger than node current `node_round + Nr` or if `vote_step` is bigger than `node_step + Ns` should be dropped and sync should be requested from the peer that sent us this vote.

PR #1957, opened 2022-08-19
feat: not accept votes with round or step bigger then expected. check sync votes bundle has same step and round

kstdl:
#1947

### DIFF
--- a/libraries/config/include/config/config.hpp
+++ b/libraries/config/include/config/config.hpp
@@ -46,6 +46,9 @@ struct NetworkConfig {
   uint16_t network_peer_blacklist_timeout = kBlacklistTimeoutDefaultInSeconds;
   bool disable_peer_blacklist = false;
   uint16_t deep_syncing_threshold = 10;
+  uint16_t vote_accepting_periods = 5;
+  uint16_t vote_accepting_rounds = 5;
+  uint16_t vote_accepting_steps = 25;
 
   void validate() const;
 };
--- a/libraries/config/include/config/state_api_config.hpp
+++ b/libraries/config/include/config/state_api_config.hpp
@@ -52,8 +52,8 @@ struct DPOSConfig {
   uint16_t max_block_author_reward = 0;
   uint16_t commission_change_delta = 0;
   uint32_t commission_change_frequency = 0;  // number of blocks
-  uint32_t delegation_delay = 0;             // number of blocks
-  uint32_t delegation_locking_period = 0;    // number of blocks
+  uint32_t delegation_delay = 5;             // number of blocks
+  uint32_t delegation_locking_period = 5;    // number of blocks
   uint32_t blocks_per_year = 0;              // number of blocks - it is calculated from lambda_ms_min
   uint16_t yield_percentage = 0;             // [%]
   std::vector<ValidatorInfo> initial_validators;
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/vote_packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/vote_packet_handler.hpp
@@ -13,7 +13,7 @@ class VotePacketHandler final : public ExtVotesPacketHandler {
  public:
   VotePacketHandler(std::shared_ptr<PeersState> peers_state, std::shared_ptr<PacketsStats> packets_stats,
                     std::shared_ptr<PbftManager> pbft_mgr, std::shared_ptr<PbftChain> pbft_chain,
-                    std::shared_ptr<VoteManager> vote_mgr, const uint32_t dpos_delay, const addr_t& node_addr);
+                    std::shared_ptr<VoteManager> vote_mgr, const NetworkConfig& net_config, const addr_t& node_addr);
 
   // Packet type that is processed by this handler
   static constexpr SubprotocolPacketType kPacketType_ = SubprotocolPacketType::VotePacket;
@@ -21,8 +21,13 @@ class VotePacketHandler final : public ExtVotesPacketHandler {
  private:
   void validatePacketRlpFormat(const PacketData& packet_data) const override;
   void process(const PacketData& packet_data, const std::shared_ptr<TaraxaPeer>& peer) override;
+  bool shouldProcessVote(const std::shared_ptr<Vote>& vote, const std::shared_ptr<TaraxaPeer>& peer);
 
+  const uint16_t kVoteAcceptingRounds;
+  const uint16_t kVoteAcceptingSteps;
+  constexpr static std::chrono::seconds kSyncRequestInterval = std::chrono::seconds(10);
   ExpirationCache<vote_hash_t> seen_votes_;
+  std::chrono::system_clock::time_point round_sync_request_time_;
 };
 
 }  // namespace taraxa::network::tarcap
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/vote_packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/vote_packet_handler.hpp
@@ -21,7 +21,7 @@ class VotePacketHandler final : public ExtVotesPacketHandler {
  private:
   void validatePacketRlpFormat(const PacketData& packet_data) const override;
   void process(const PacketData& packet_data, const std::shared_ptr<TaraxaPeer>& peer) override;
-  bool shouldProcessVote(const std::shared_ptr<Vote>& vote, const std::shared_ptr<TaraxaPeer>& peer);
+  bool checkVoteMaxPeriodRoundStep(const std::shared_ptr<Vote>& vote, const std::shared_ptr<TaraxaPeer>& peer);
 
   const uint16_t kVoteAcceptingRounds;
   const uint16_t kVoteAcceptingSteps;
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/votes_sync_packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/votes_sync_packet_handler.hpp
@@ -16,7 +16,7 @@ class VotesSyncPacketHandler final : public ExtVotesPacketHandler {
   VotesSyncPacketHandler(std::shared_ptr<PeersState> peers_state, std::shared_ptr<PacketsStats> packets_stats,
                          std::shared_ptr<PbftManager> pbft_mgr, std::shared_ptr<PbftChain> pbft_chain,
                          std::shared_ptr<VoteManager> vote_mgr, std::shared_ptr<NextVotesManager> next_votes_mgr,
-                         std::shared_ptr<DbStorage> db, const uint32_t dpos_delay, const addr_t& node_addr);
+                         std::shared_ptr<DbStorage> db, uint32_t vote_accepting_periods, const addr_t& node_addr);
 
   void broadcastPreviousRoundNextVotesBundle();
 
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
@@ -8,9 +8,12 @@ namespace taraxa::network::tarcap {
 VotePacketHandler::VotePacketHandler(std::shared_ptr<PeersState> peers_state,
                                      std::shared_ptr<PacketsStats> packets_stats, std::shared_ptr<PbftManager> pbft_mgr,
                                      std::shared_ptr<PbftChain> pbft_chain, std::shared_ptr<VoteManager> vote_mgr,
-                                     const uint32_t dpos_delay, const addr_t &node_addr)
+                                     const NetworkConfig &net_config, const addr_t &node_addr)
     : ExtVotesPacketHandler(std::move(peers_state), std::move(packets_stats), std::move(pbft_mgr),
-                            std::move(pbft_chain), std::move(vote_mgr), dpos_delay, node_addr, "PBFT_VOTE_PH"),
+                            std::move(pbft_chain), std::move(vote_mgr), net_config.vote_accepting_periods, node_addr,
+                            "PBFT_VOTE_PH"),
+      kVoteAcceptingRounds(net_config.vote_accepting_rounds),
+      kVoteAcceptingSteps(net_config.vote_accepting_steps),
       seen_votes_(1000000, 1000) {}
 
 void VotePacketHandler::validatePacketRlpFormat([[maybe_unused]] const PacketData &packet_data) const {
@@ -18,6 +21,67 @@ void VotePacketHandler::validatePacketRlpFormat([[maybe_unused]] const PacketDat
   if (items == 0 || items > kMaxVotesInPacket) {
     throw InvalidRlpItemsCountException(packet_data.type_str_, items, kMaxVotesInPacket);
   }
+}
+
+bool VotePacketHandler::shouldProcessVote(const std::shared_ptr<Vote> &vote, const std::shared_ptr<TaraxaPeer> &peer) {
+  const auto [current_pbft_round, current_pbft_period] = pbft_mgr_->getPbftRoundAndPeriod();
+  const auto current_pbft_step = pbft_mgr_->getPbftStep();
+  const auto vote_hash = vote->getHash();
+
+  if (vote->getPeriod() - 1 > pbft_chain_->getPbftChainSize() + kVoteAcceptingPeriods) {
+    // Do not request round sync too often here
+    if (std::chrono::system_clock::now() - round_sync_request_time_ > kSyncRequestInterval) {
+      // request PBFT chain sync from this node
+      sealAndSend(peer->getId(), SubprotocolPacketType::GetPbftSyncPacket,
+                  std::move(dev::RLPStream(1) << std::max(vote->getPeriod() - 1, peer->pbft_chain_size_.load())));
+      round_sync_request_time_ = std::chrono::system_clock::now();
+    }
+    LOG(log_dg_) << "Skip vote " << vote_hash.abridged()
+                 << ". Failed period validation against current state. Vote: {period: " << vote->getPeriod()
+                 << ", round: " << vote->getRound() << ", step: " << vote->getStep()
+                 << "}. Current state: {period: " << current_pbft_period << ", round: " << current_pbft_round
+                 << ", step: " << pbft_mgr_->getPbftStep() << "}";
+    return false;
+  }
+
+  auto checking_round = current_pbft_round;
+  // If period is not the same we assuming current round is equal to 1
+  // So we won't accept votes for future period with round bigger than kVoteAcceptingSteps
+  if (current_pbft_period != vote->getPeriod()) {
+    checking_round = 1;
+  }
+  if (vote->getRound() >= checking_round + kVoteAcceptingRounds) {
+    LOG(log_dg_) << "Skip vote " << vote_hash.abridged()
+                 << ". Failed round validation against current state. Checking round " << checking_round
+                 << ". Vote: {period: " << vote->getPeriod() << ", round: " << vote->getRound()
+                 << ", step: " << vote->getStep() << "}. Current state: {period: " << current_pbft_period
+                 << ", round: " << current_pbft_round << ", step: " << current_pbft_step << "}";
+    // Do not request round sync too often here
+    if (std::chrono::system_clock::now() - round_sync_request_time_ > kSyncRequestInterval) {
+      // request round votes sync from this node
+      sealAndSend(peer->getId(), GetVotesSyncPacket,
+                  std::move(dev::RLPStream(3) << current_pbft_period << current_pbft_round << 0));
+      round_sync_request_time_ = std::chrono::system_clock::now();
+    }
+    return false;
+  }
+
+  auto checking_step = pbft_mgr_->getPbftStep();
+  // If round is not the same we assuming current step is equal to 1
+  // So we won't accept votes for future rounds with step bigger than kVoteAcceptingSteps
+  if (current_pbft_round != vote->getRound()) {
+    checking_step = 1;
+  }
+  if (vote->getStep() >= checking_step + kVoteAcceptingSteps) {
+    LOG(log_dg_) << "Skip vote " << vote_hash.abridged()
+                 << ". Failed step validation against current state. Checking step " << checking_step
+                 << ". Vote: {period: " << vote->getPeriod() << ", round: " << vote->getRound()
+                 << ", step: " << vote->getStep() << "}. Current state: {period: " << current_pbft_period
+                 << ", round: " << current_pbft_round << ", step: " << current_pbft_step << "}";
+    return false;
+  }
+
+  return true;
 }
 
 void VotePacketHandler::process(const PacketData &packet_data, const std::shared_ptr<TaraxaPeer> &peer) {
@@ -36,22 +100,26 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
                    << ") already seen.";
       continue;
     }
-    peer->markVoteAsKnown(vote_hash);
 
     // Standard vote
     if (vote->getPeriod() >= current_pbft_period) {
       if (vote_mgr_->voteInVerifiedMap(vote)) {
         LOG(log_dg_) << "Vote " << vote_hash.abridged() << " already inserted in verified queue";
-      }
+        // We don't need to validate and process it, but should add to votes for gossiping
+      } else {
+        if (!shouldProcessVote(vote, peer)) {
+          continue;
+        }
 
-      if (auto vote_is_valid = validateStandardVote(vote); vote_is_valid.first == false) {
-        LOG(log_wr_) << "Vote " << vote_hash.abridged() << " validation failed. Err: " << vote_is_valid.second;
-        continue;
-      }
+        if (auto vote_is_valid = validateStandardVote(vote); vote_is_valid.first == false) {
+          LOG(log_wr_) << "Vote " << vote_hash.abridged() << " validation failed. Err: " << vote_is_valid.second;
+          continue;
+        }
 
-      if (!vote_mgr_->addVerifiedVote(vote)) {
-        LOG(log_dg_) << "Vote " << vote_hash << " already inserted in verified queue(race condition)";
-        continue;
+        if (!vote_mgr_->addVerifiedVote(vote)) {
+          LOG(log_dg_) << "Vote " << vote_hash << " already inserted in verified queue(race condition)";
+          continue;
+        }
       }
 
     } else if (vote->getPeriod() == current_pbft_period - 1 && vote->getType() == PbftVoteTypes::cert_vote_type) {
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
@@ -23,12 +23,13 @@ void VotePacketHandler::validatePacketRlpFormat([[maybe_unused]] const PacketDat
   }
 }
 
-bool VotePacketHandler::shouldProcessVote(const std::shared_ptr<Vote> &vote, const std::shared_ptr<TaraxaPeer> &peer) {
+bool VotePacketHandler::checkVoteMaxPeriodRoundStep(const std::shared_ptr<Vote> &vote,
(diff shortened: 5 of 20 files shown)


libraries/core_libs/network/include/network/tarcap/packets_handlers/vote_packet_handler.hpp
```diff
@@ -13,7 +13,8 @@ class VotePacketHandler final : public ExtVotesPacketHandler {
  public:
   VotePacketHandler(std::shared_ptr<PeersState> peers_state, std::shared_ptr<PacketsStats> packets_stats,
                     std::shared_ptr<PbftManager> pbft_mgr, std::shared_ptr<PbftChain> pbft_chain,
-                    std::shared_ptr<VoteManager> vote_mgr, const uint32_t dpos_delay, const addr_t& node_addr);
+                    std::shared_ptr<VoteManager> vote_mgr, const uint32_t dpos_delay,
+                    const uint16_t vote_accepting_rounds, const uint16_t vote_accepting_steps, const addr_t& node_addr);
```
MatusKysel, line 17: please remove `const` it is irrelevant

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -59,6 +64,31 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
...
       if (vote_mgr_->voteInVerifiedMap(vote)) {
         LOG(log_dg_) << "Vote " << vote_hash.abridged() << " already inserted in verified queue";
+        // We don't need to process this vote because it is already in verified map
+        continue;
+      }
+
+      if (vote->getRound() > pbft_mgr_->getPbftRound() + kVoteAcceptingRounds) {
+        LOG(log_nf_) << "Skip. Round too big " << pbft_mgr_->getPbftRound() << " " << vote->getRound();
+        // Do not request round sync too often here
+        if (std::chrono::system_clock::now() - round_sync_request_time_ > std::chrono::seconds(10)) {
```
MatusKysel, line 74: those 10s could be another const variable

MatusKysel: and is this ok, to mark this vote at this time as seen? line `if (!seen_votes_.insert(vote_hash)) {` , because if this vote comes again, it won't be processed

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
```diff
@@ -35,21 +35,37 @@ void VotesSyncPacketHandler::process(const PacketData &packet_data, const std::s
     return;
   }
 
+  std::map<blk_hash_t, std::pair<uint64_t, uint64_t>> voted_value_round_step;
```
MatusKysel, line 38: unordered?

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
```diff
@@ -35,21 +35,37 @@ void VotesSyncPacketHandler::process(const PacketData &packet_data, const std::s
...
+
   std::vector<std::shared_ptr<Vote>> next_votes;
   const auto next_votes_count = packet_data.rlp_.itemCount();
   for (size_t i = 0; i < next_votes_count; i++) {
     auto next_vote = std::make_shared<Vote>(packet_data.rlp_[i].data().toBytes());
     const auto next_vote_hash = next_vote->getHash();
+
+    // votes for the same value should be from one round and step
+    auto [previous_votes_round, previous_votes_step] = voted_value_round_step[next_vote->getBlockHash()];
```
MatusKysel, line 47: I do not understand this logic. First you check here but you insert few lines under? so is this check even working?

MatusKysel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
kstdl: We could have votes for 2 values here. One of this values should be NULL_HASH. So yeah, I missed one more check here

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -59,6 +64,31 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
     } else {  // standard vote -> vote_round >= current_pbft_round
       if (vote_mgr_->voteInVerifiedMap(vote)) {
         LOG(log_dg_) << "Vote " << vote_hash.abridged() << " already inserted in verified queue";
+        // We don't need to process this vote because it is already in verified map
+        continue;
+      }
+
+      if (vote->getRound() > pbft_mgr_->getPbftRound() + kVoteAcceptingRounds) {
+        LOG(log_nf_) << "Skip. Round too big " << pbft_mgr_->getPbftRound() << " " << vote->getRound();
```
JakubFornadel, line 72: Add vote hash to the msg + "Skip vote"

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -59,6 +64,31 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
...
+      }
+
+      auto current_step = pbft_mgr_->getPbftStep();
+      // If round is not the same we assuming current step is equal to 0
+      // So we won't accept votes for future rounds with step bigger than kVoteAcceptingSteps
+      if (pbft_mgr_->getPbftRound() != vote->getRound()) {
+        current_step = 0;
+      }
+      if (vote->getStep() > current_step + kVoteAcceptingSteps) {
+        LOG(log_er_) << "Skip. Step too big " << pbft_mgr_->getPbftStep() << " " << vote->getStep();
```
JakubFornadel, line 90: https://github.com/Taraxa-project/taraxa-node/pull/1957/files#r952244501

libraries/core_libs/network/include/network/tarcap/packets_handlers/votes_sync_packet_handler.hpp
```diff
@@ -24,6 +24,7 @@ class VotesSyncPacketHandler final : public ExtVotesPacketHandler {
   static constexpr SubprotocolPacketType kPacketType_ = SubprotocolPacketType::VotesSyncPacket;
 
  private:
+  bool validatePacket(const PacketData& packet_data) const;
```
JakubFornadel, line 27: where is this used ?

libraries/core_libs/network/include/network/tarcap/packets_handlers/vote_packet_handler.hpp
JakubFornadel, line 17: just an optional note: if there is more than 1 pram from config that you need to pass here, you could simply pass the whole config instead of 3 separate params...

libraries/config/include/config/config.hpp
```diff
@@ -46,6 +46,8 @@ struct NetworkConfig {
   uint16_t network_peer_blacklist_timeout = kBlacklistTimeoutDefaultInSeconds;
   bool disable_peer_blacklist = false;
   uint16_t deep_syncing_threshold = 10;
+  uint16_t vote_accepting_rounds = 5;
+  uint16_t vote_accepting_steps = 50;
```
JakubFornadel, line 50: Did you try some smaller number ? did it affect the tests ?

kstdl: Accepting only 1 future round didn't worked, so I decided that this values should be ok. It shouldn't be possible to save too much votes with this config. And even this affected some tests that was speeded-up(related to first param in `make_configs` function), for example `FullNodeTest.sync_five_nodes` needed 4x time to complete because it was dropping some votes. But it was passing. So I introduced this thing https://github.com/Taraxa-project/taraxa-node/pull/1957/files#diff-b8695b851212c29d15116200602b07a1de6e81b2e9d49ec2760dc31b7261ddd1R198

JakubFornadel: sure, 1 round cannot work. I think 5 rounds is ok, I am asking about those 50 steps - did you try to use 10, 20, etc... ? If that did not affect the tests I think the less the better

JakubFornadel (approved):

libraries/config/include/config/config.hpp
kstdl: I will test it with less steps

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -30,50 +33,76 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
...
     LOG(log_dg_) << "Received PBFT vote " << vote_hash;
 
+    // Check this before we are marking vote as seen to have an opportunity to process it after
+    if (vote->getRound() > pbft_mgr_->getPbftRound() + kVoteAcceptingRounds) {
+      LOG(log_nf_) << "Skip vote " << vote_hash.abridged() << ". Round too big " << pbft_mgr_->getPbftRound() << " "
+                   << vote->getRound();
+      // Do not request round sync too often here
+      if (std::chrono::system_clock::now() - round_sync_request_time_ > kSyncRequestInterval) {
+        // request sync from this node
+        sealAndSend(peer->getId(), GetVotesSyncPacket, std::move(dev::RLPStream(2) << pbft_mgr_->getPbftRound() << 0));
```
JakubFornadel, line 43: I guess this requesting mechanism cannot work anymore without providing also period...

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -30,50 +33,76 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
     const auto vote_hash = vote->getHash();
     LOG(log_dg_) << "Received PBFT vote " << vote_hash;
 
+    // Check this before we are marking vote as seen to have an opportunity to process it after
+    if (vote->getRound() > pbft_mgr_->getPbftRound() + kVoteAcceptingRounds) {
```
JakubFornadel, line 37: this will also not work in our new design with periods & rounds

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
```diff
@@ -54,44 +86,53 @@ void VotesSyncPacketHandler::process(const PacketData &packet_data, const std::s
...
-
-      // CONCERN: Huh? Race condition...
-
-      if (!vote_mgr_->insertUniqueVote(vote)) {
-        LOG(log_dg_) << "Non unique vote " << next_vote_hash.abridged() << " (race condition)";
-        continue;
-      }
-    }
+  // Remove invalid votes from the next_votes bundle. Should we do something with it if
+  auto remove_it = std::remove_if(
```
JakubFornadel, line 92: what ? why do we need to remove anything ?

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
```diff
@@ -36,11 +37,42 @@ void VotesSyncPacketHandler::process(const PacketData &packet_data, const std::s
...
+  std::unordered_map<blk_hash_t, uint64_t> voted_value_round;
   std::vector<std::shared_ptr<Vote>> next_votes;
+
   const auto next_votes_count = packet_data.rlp_.itemCount();
+  // Parse and mark all votes as known for sender peer to avoid resends to it
   for (size_t i = 0; i < next_votes_count; i++) {
     auto next_vote = std::make_shared<Vote>(packet_data.rlp_[i].data().toBytes());
-    const auto next_vote_hash = next_vote->getHash();
+    peer->markVoteAsKnown(next_vote->getHash());
+    next_votes.push_back(next_vote);
```
JakubFornadel, line 48: I guess pushing all votes here and then removing is not ideal

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
JakubFornadel: previous logic was quite straight forward, in loop either use standard vote validation or next vote validation and then push it into the next votes manager if those were next votes or just regossip if they were standard vote

kstdl: We have checks for whole bundle. And if we see that something wrong with it, then we shouldn't process any votes from it. We can introduce new collection and move processed(valid) elements to it, but I thought that it is cleaner to just remove invalid elements from the first collection that we already have

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
kstdl: What is better way to do it then? I thought it is the best one :)

JakubFornadel: what was here before

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
JakubFornadel: I guess I dont follow

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
justinsnapp: I guess we should properly implement a class to store (period, round) and then we can implement comparison operators > , < , and such    But in this case logic would be I think if period is equal then this check is ok, else if period is one higher, then vote->getRound() > kVoteAcceptingRounds, or if period is higher by more than one.

MatusKysel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -30,28 +62,54 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
...
-
-      if (auto vote_is_valid = validateStandardVote(vote); vote_is_valid.first == false) {
-        LOG(log_wr_) << "Vote " << vote_hash.abridged() << " validation failed. Err: " << vote_is_valid.second;
-        continue;
-      }
-
-      if (!vote_mgr_->addVerifiedVote(vote)) {
-        LOG(log_dg_) << "Vote " << vote_hash << " already inserted in verified queue(race condition)";
-        continue;
+        // We don't need to validate and process it, but should add to votes for gossiping
```
JakubFornadel, line 102: we could just do:

```
if (!vote_mgr_->voteInVerifiedMap(vote)) {
  // validate....

}
```

without any else ?

kstdl: But do we need to call `validateStandardVote` and `addVerifiedVote` if this vote is already in verified map?

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -30,28 +62,54 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
...
+      if (std::chrono::system_clock::now() - round_sync_request_time_ > kSyncRequestInterval) {
+        // request sync from this node
+        sealAndSend(peer->getId(), SubprotocolPacketType::GetPbftSyncPacket,
+                    std::move(dev::RLPStream(1) << vote->getPeriod()));
+        round_sync_request_time_ = std::chrono::system_clock::now();
+      }
+      continue;
+    }
+    // Check this before we are marking vote as seen to have an opportunity to process it after
+    if (!isVoteRoundStepRelevant(vote)) {
```
JakubFornadel, line 79: this should be called only for standard vote as it does not make sense for reward vote

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -20,6 +23,35 @@ void VotePacketHandler::validatePacketRlpFormat([[maybe_unused]] const PacketDat
   }
 }
 
+bool VotePacketHandler::isVoteRoundStepRelevant(const std::shared_ptr<Vote> &vote) const {
+  auto checking_round = pbft_mgr_->getPbftRound();
+  // If period is not the same we assuming current round is equal to 0
+  // So we won't accept votes for future period with round bigger than kVoteAcceptingSteps
+  if (pbft_chain_->getPbftChainSize() != vote->getPeriod()) {
```
JakubFornadel, line 30: I think this is wrong logic, you should be checking pbft_chain_->getPbftChainSize() + 1, but dont use chainSize,  use:

const auto [checking_round, current_pbft_period] = pbft_mgr_->getPbftRoundAndPeriod();
if (current_pbft_period != vote->getPeriod()) {
  checking_round = 0
}

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -20,6 +23,35 @@ void VotePacketHandler::validatePacketRlpFormat([[maybe_unused]] const PacketDat
...
+  if (vote->getRound() > checking_round + kVoteAcceptingRounds) {
+    LOG(log_nf_) << "Skip vote " << vote_hash.abridged() << ". Round too big " << pbft_mgr_->getPbftRound() << " "
+                 << vote->getRound();
+    return false;
+  }
+
+  auto checking_step = pbft_mgr_->getPbftStep();
+  // If round is not the same we assuming current step is equal to 0
+  // So we won't accept votes for future rounds with step bigger than kVoteAcceptingSteps
+  if (pbft_mgr_->getPbftRound() != vote->getRound()) {
```
JakubFornadel, line 43: dont call this obver and over, you should get it just once as I mentioned in one of the comments

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -20,6 +23,35 @@ void VotePacketHandler::validatePacketRlpFormat([[maybe_unused]] const PacketDat
...
+bool VotePacketHandler::isVoteRoundStepRelevant(const std::shared_ptr<Vote> &vote) const {
+  auto checking_round = pbft_mgr_->getPbftRound();
+  // If period is not the same we assuming current round is equal to 0
+  // So we won't accept votes for future period with round bigger than kVoteAcceptingSteps
+  if (pbft_chain_->getPbftChainSize() != vote->getPeriod()) {
+    checking_round = 0;
+  }
+  const auto vote_hash = vote->getHash();
+  if (vote->getRound() > checking_round + kVoteAcceptingRounds) {
+    LOG(log_nf_) << "Skip vote " << vote_hash.abridged() << ". Round too big " << pbft_mgr_->getPbftRound() << " "
```
JakubFornadel, line 35: no for pbft_mgr_->getPbftRound(). Also it would be nice to see in log which round is related to vote and which round is current pbft round

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -20,6 +23,35 @@ void VotePacketHandler::validatePacketRlpFormat([[maybe_unused]] const PacketDat
...
+
+  auto checking_step = pbft_mgr_->getPbftStep();
+  // If round is not the same we assuming current step is equal to 0
+  // So we won't accept votes for future rounds with step bigger than kVoteAcceptingSteps
+  if (pbft_mgr_->getPbftRound() != vote->getRound()) {
+    checking_step = 0;
+  }
+  if (vote->getStep() > checking_step + kVoteAcceptingSteps) {
+    LOG(log_er_) << "Skip vote " << vote_hash.abridged() << ". Step too big " << pbft_mgr_->getPbftStep() << " "
+                 << vote->getStep();
```
JakubFornadel, line 48: this log wont tell me anything if vote is from different round as pbft_mgr_->getPbftStep() is totally irrelevant then and the message will be confusing

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
JakubFornadel: also https://github.com/Taraxa-project/taraxa-node/pull/1957/files#r954845005

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
JakubFornadel: also this is checking/validating vote max round & step, so I guess something like validateVoteMaxRoundAndStep would be more clear

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
```diff
@@ -81,27 +108,28 @@ void VotesSyncPacketHandler::process(const PacketData &packet_data, const std::s
...
       if (vote_mgr_->voteInVerifiedMap(next_vote)) {
         LOG(log_dg_) << "Vote " << next_vote_hash.abridged() << " already inserted in verified queue";
-      }
-
-      if (auto vote_is_valid = validateStandardVote(next_vote); vote_is_valid.first == false) {
-        LOG(log_wr_) << "Vote " << next_vote_hash.abridged() << " validation failed. Err: " << vote_is_valid.second;
-        continue;
-      }
+        // We don't need to perform other checks, but it shouldn't be removed as it is used for next calls
+      } else {
```
JakubFornadel, line 112: https://github.com/Taraxa-project/taraxa-node/pull/1957/files#r954829859

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
```diff
@@ -44,11 +45,37 @@ void VotesSyncPacketHandler::process(const PacketData &packet_data, const std::s
...
+        LOG(log_er_) << "Votes sync bundle has votes " << next_vote_hash.abridged() << " with different rounds("
+                     << next_vote->getRound() << ", " << previous_votes_round
+                     << "). The peer may be a malicious player, will be disconnected";
+        disconnect(packet_data.from_node_id_, dev::p2p::UserReason);
+        return;
+      }
+    } else {
+      voted_value_round[next_vote->getBlockHash()] = next_vote->getRound();
+      // We could have votes only for two values. And one of these should be NULL_HASH
+      if (voted_value_round.size() > 2 ||
```
JakubFornadel, line 71: I dont follow whole logic around voted_value_round. If we wanna check that there are only two values, we can simply add this check:

```
const auto peer_pbft_period = vote->getPeriod();
const auto peer_pbft_round = vote->getRound();
const auto votes_voted_block = vote->getBlockHash();

if (next_vote->getBlockHash() != votes_voted_block || next_vote->getBlockHash() != NULL_BLOCK_HASH) {
  LOG(log_er_) << "Received next votes bundle with unmatched periods from " << packet_data.from_node_id_
               << ". The peer may be a malicious player, will be disconnected";
  disconnect(packet_data.from_node_id_, dev::p2p::UserReason);
  return;
}
```

kstdl: That is overcomplicated right now for no reason, but original idea was also to check that we have voted for one voted value from the same step

libraries/config/include/config/state_api_config.hpp
```diff
@@ -52,8 +52,8 @@ struct DPOSConfig {
   uint16_t max_block_author_reward = 0;
   uint16_t commission_change_delta = 0;
   uint32_t commission_change_frequency = 0;  // number of blocks
-  uint32_t delegation_delay = 0;             // number of blocks
-  uint32_t delegation_locking_period = 0;    // number of blocks
+  uint32_t delegation_delay = 5;             // number of blocks
+  uint32_t delegation_locking_period = 5;    // number of blocks
```
kstdl, line 56: Not sure why we was setting this to zero by default, but I think it should be more correct to set it to 5 here. At least for some reason configs for py_tests is generated from this default values, but not from some json config 
@JakubFornadel what do you think about this?

JakubFornadel: no idea

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -36,22 +100,26 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
...
-        LOG(log_wr_) << "Vote " << vote_hash.abridged() << " validation failed. Err: " << vote_is_valid.second;
-        continue;
-      }
-
-      if (!vote_mgr_->addVerifiedVote(vote)) {
-        LOG(log_dg_) << "Vote " << vote_hash << " already inserted in verified queue(race condition)";
-        continue;
+        // We don't need to validate and process it, but should add to votes for gossiping
+      } else {
+        if (!shouldProcessVote(vote, peer)) {
```
JakubFornadel, line 110: - this should be called inside validateStandardVote as it is check related to standard votes and validateStandardVote is also called in votes sync packet handler and right now you dont call it there even though we should...
- also I still dont get the generic name shouldProcessVote. This is pretty simple and specific function, which is checking max bounds for period, round and step and with proper name I would not need to open that function at all when going through code. When I see shouldProcessVote I have no idea what is going on inside so I have to check it if I am not that familiar with the code

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
JakubFornadel: true, we dont - there is !

```
if (!vote_mgr_->voteInVerifiedMap(vote)) {
  // validate....

}
```

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -20,6 +23,67 @@ void VotePacketHandler::validatePacketRlpFormat([[maybe_unused]] const PacketDat
   }
 }
 
+bool VotePacketHandler::shouldProcessVote(const std::shared_ptr<Vote> &vote, const std::shared_ptr<TaraxaPeer> &peer) {
+  const auto [current_pbft_round, current_pbft_period] = pbft_mgr_->getPbftRoundAndPeriod();
+  const auto current_pbft_step = pbft_mgr_->getPbftStep();
+  const auto vote_hash = vote->getHash();
+
+  if (vote->getPeriod() - 1 > pbft_chain_->getPbftChainSize() + kVoteAcceptingPeriods) {
```
JakubFornadel, line 31: I did create comment before, which you resolved by yourself and then created the same thing. Do not use  getPbftChainSize here, use current_pbft_period

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
JakubFornadel: I mean it does the same thing, I just find the if else not that useful as it can be done with single if but choose what you like more

kstdl: Yep. I missed that  added `!` in your comment, so I missed all the logic behind it. And this log inside shouldn't be very useful, so I will change this

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
kstdl: If we will also check this inside of `VotesSyncPacketHandler`,  then we wouldn't process votes with bigger period\round\step than our configured values, so for me it looks like syncing won't work in most of the cases 
Thats why it is only here and not inside of this `validateStandardVote`
Is `checkVoteMaxBounds` is good enough you think? Because I don't like something as `checkVoteMaxPeriodRoundStep` is kind of too big

PR #1959, opened 2022-08-22
chore: optimize dag verify

mfrankovi:
When verifying received DAG block all transactions need to be retrieved from pool or database to be able to verify block weight. If there is a lot of transactions in the block that are already finalized these transactions are retrieved from the database from period data which is very slow. A block with 10000 transactions took almost 3 minutes to verify when transactions are pulled from period data.

Following optimizations are made:
Check both the pool and nonfinalized_transactions_in_dag_ where transactions are stored in memory before checking the db for finalized transactions.
Keep last finalized block transactions in memory since DAG block might often have transactions which have just been finalized
When actually retrieving transactions from database, group transactions per period so that multiple finalized transactions from same period are all fetched in the same period data retrieval

### DIFF
--- a/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
@@ -370,7 +370,7 @@ std::optional<std::map<trx_hash_t, std::shared_ptr<Transaction>>> TransactionMan
     auto finalizedTransactions = db_->getFinalizedTransactions(finalized_trx_hashes);
     if (finalizedTransactions.first.has_value()) {
       for (auto trx : *finalizedTransactions.first) {
-        transactions.emplace(trx->getHash(), trx);
+        transactions.emplace(trx->getHash(), std::move(trx));
       }
     } else {
       LOG(log_er_) << "Block " << blk.getHash() << " has missing transaction " << finalizedTransactions.second;
--- a/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
@@ -21,12 +21,6 @@ TransactionManager::TransactionManager(FullNodeConfig const &conf, std::shared_p
   }
 }
 
-uint64_t TransactionManager::estimateTransactionGasByHash(const trx_hash_t &hash,
-                                                          std::optional<uint64_t> proposal_period) const {
-  const auto &trx = getTransaction(hash);
-  return estimateTransactionGas(trx, proposal_period);
-}
-
 uint64_t TransactionManager::estimateTransactionGas(std::shared_ptr<Transaction> trx,
                                                     std::optional<uint64_t> proposal_period) const {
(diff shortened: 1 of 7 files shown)


libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -341,24 +337,48 @@ void TransactionManager::moveNonFinalizedTransactionsToTransactionsPool(std::uno
 }
 
 // Verify all block transactions are present
-bool TransactionManager::checkBlockTransactions(DagBlock const &blk) {
+std::optional<std::map<trx_hash_t, std::shared_ptr<Transaction>>> TransactionManager::getBlockTransactions(
```
MatusKysel, line 340: why it needs to be map? and does it need to be optional? can we return empty container?

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -341,24 +337,48 @@ void TransactionManager::moveNonFinalizedTransactionsToTransactionsPool(std::uno
...
+          } else {
+            finalized_trx_hashes.emplace_back(tx_hash);
+          }
+        }
       }
     }
+    auto finalizedTransactions = db_->getFinalizedTransactions(finalized_trx_hashes);
+    if (finalizedTransactions.first.has_value()) {
+      for (auto trx : *finalizedTransactions.first) {
+        transactions.emplace(trx->getHash(), trx);
```
MatusKysel, line 373: we can do move on trx

MatusKysel (approved):

libraries/core_libs/storage/src/storage.cpp
```diff
@@ -552,6 +552,35 @@ std::shared_ptr<Transaction> DbStorage::getTransaction(trx_hash_t const& hash) {
   return nullptr;
 }
 
+std::pair<std::optional<SharedTransactions>, trx_hash_t> DbStorage::getFinalizedTransactions(
+    std::vector<trx_hash_t> const& trx_hashes) const {
+  // Map of period to position of transactions within a period
+  std::map<uint64_t, std::set<uint32_t>> period_map;
+  SharedTransactions transactions;
```
MatusKysel, line 559: you can reserve size

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp

libraries/core_libs/storage/src/storage.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
mfrankovi: It is a map because later it is needed as a map to match the estimates within the dag block. I think it is cleaner to have an optional to point that it is an error and not just no transactions found. But either way is good

JakubFornadel (approved):

PR #1961, opened 2022-08-23
Rewrite PBFT to simplify algorithm

justinsnapp:
## Purpose

The introduction of periods in to PBFT votes created both challenges and a major opportunity to solve PBFT complexity.

## Changes made

This PR (draft) is attempting to make the following changes...
1. Votes in vote manager are stored by period, round, step not round, period, step
2. 2t+1 votes requirement is stored in db by period not round
3. As soon as 2t+1 cert votes are observed a block is pushed into the chain for execution and period is advanced 1 and round is reset to 1
4. We only identify leader value if we have the block and have verified it. (Having done so we should not do it again in cert voting, but that change hasn't been made)
5. Remove giveUpNextVoted and giveUpSoftVoted functions and need for such logic
6. Remove storing "own starting value" in the db.  
7. Should never get a next voted value that is not a valid value.
8. Period used for all votes within a round should be unchanged (only advance period upon pushing block into chain)

Tests are definitely not passing, so probably both problems with implementation and with tests, particularly the vote creation in the tests often seems to assume no longer valid reasoning about period and round relationship.

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -766,14 +766,16 @@ void PbftManager::proposeBlock_() {
     proposed_block_ = proposePbftBlock_();
 
     if (proposed_block_) {
-      if (auto vote_weight = placeVote_(proposed_block_->getBlockHash(), propose_vote_type, period, round, step_);
-          vote_weight) {
+      if (auto vote = generateVoteWithWeight(proposed_block_->getBlockHash(), propose_vote_type, period, round, step_);
+          vote) {
         LOG(log_nf_) << "Placed propose vote for new block " << proposed_block_->getBlockHash() << ", vote weight "
-                     << vote_weight << ", period " << period << ", round " << round << ", step " << step_;
+                     << *vote->getWeight() << ", period " << period << ", round " << round << ", step " << step_;
 
         // broadcast pbft block
         if (auto net = network_.lock()) {
+          // In propose step, send block before vote as we require block to be present during vote validation
           net->getSpecificHandler<network::tarcap::PbftBlockPacketHandler>()->onNewPbftBlock(proposed_block_);
+          net->getSpecificHandler<network::tarcap::VotePacketHandler>()->onNewPbftVotes({std::move(vote)});
         }
       }
     }
@@ -799,15 +801,18 @@ void PbftManager::proposeBlock_() {
       }
     }
 
-    if (auto vote_weight = placeVote_(previous_round_next_voted_value_.first, propose_vote_type, period, round, step_);
-        vote_weight) {
+    if (auto vote =
+            generateVoteWithWeight(previous_round_next_voted_value_.first, propose_vote_type, period, round, step_);
+        vote) {
       LOG(log_nf_) << "Placed propose vote for previous round next voted value "
-                   << previous_round_next_voted_value_.first << ", vote weight " << vote_weight << ", period " << period
-                   << ", round " << round << ", step " << step_;
+                   << previous_round_next_voted_value_.first << ", vote weight " << *vote->getWeight() << ", period "
+                   << period << ", round " << round << ", step " << step_;
 
       // broadcast pbft block
       if (auto net = network_.lock()) {
+        // In propose step, send block before vote as we require block to be present during vote validation
         net->getSpecificHandler<network::tarcap::PbftBlockPacketHandler>()->onNewPbftBlock(pbft_block);
+        net->getSpecificHandler<network::tarcap::VotePacketHandler>()->onNewPbftVotes({std::move(vote)});
       }
     }
     return;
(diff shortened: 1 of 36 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -688,55 +700,36 @@ bool PbftManager::stateOperations_() {
...
+      advancePeriod_();
+
+      // Restart while loop  CONCERN: still need to do that?
+      return true;
+    
+    } else {
+      LOG(log_er_) << "PBFT block " << certified_block->voted_block_hash
+                   << " certified but not pushed into chain in round " << round;
+
+      // CONCERN: How would we get here?
```
JakubFornadel, line 726: very easily, you saw 2t+1 cert votes but you are missing 1 some data (dag block, tx, reward vote....) so block was not pushed

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -445,6 +435,23 @@ bool PbftManager::resetRound_() {
...
+void PbftManager::advancePeriod_() {
+  auto new_period = pbft_chain_->getPbftChainSize();
+  
+  // Update in DB first
+  auto batch = db_->createWriteBatch();
+  db_->addPbftMgrFieldToBatch(PbftMgrRoundStep::PbftPeriod, new_period, batch);
+  db_->addPbft2TPlus1ToBatchForPeriod(period_, TWO_T_PLUS_ONE, batch);
+  db_->commitWriteBatch(batch);
+
+  assert(new_period == period_ + 1);
```
JakubFornadel, line 447: this is bug: advancePeriod might be called even if block was not pushed

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
justinsnapp: Ok you are quite right @JakubFornadel

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -19,12 +19,13 @@ ExtVotesPacketHandler::ExtVotesPacketHandler(std::shared_ptr<PeersState> peers_s
...
 std::pair<bool, std::string> ExtVotesPacketHandler::validateStandardVote(const std::shared_ptr<Vote> &vote) const {
-  const uint64_t current_pbft_period = pbft_chain_->getPbftChainSize();
-  const auto current_pbft_round = pbft_mgr_->getPbftRound();
+  const auto [current_pbft_round, current_pbft_period] = pbft_mgr_->getPbftRoundAndPeriod();
 
   // Old vote or vote from too far in the future, can be dropped
   // TODO[1880]: should be vote->getPeriod() <= current_pbft_period - if <=, some tests are failing due to missing
   // reward votes -> whole rewards votes gossiping need to be checked...
+
+  // CONCERN: Why the minus one on the vote period?
```
JakubFornadel, line 28: becasue votes are usually for future period (not yet pushed into the chain) so when validating vote in dpos contract, we use vote_period -  1

libraries/core_libs/network/include/network/tarcap/taraxa_peer.hpp
```diff
@@ -86,6 +86,8 @@ class TaraxaPeer : public boost::noncopyable {
   std::atomic<bool> syncing_ = false;
   std::atomic<uint64_t> dag_level_ = 0;
   std::atomic<uint64_t> pbft_chain_size_ = 0;
+  // CONCERN: period different from chain size? how?
```
JakubFornadel, line 89: actually period should always be: `chain_size + 1 == period`

libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
kstdl: Wouldn't it be more correct to check it on `vote_period - delefation_delay`?

libraries/core_libs/network/include/network/tarcap/taraxa_peer.hpp
justinsnapp: @JakubFornadel  is that true if we have seen 2t+1 cert votes for a block but don't yet have the block to add to the chain?   Don't we now advance to the next period without getting the block?    ... And I think actually we still need to add in check that (period == chain_size + 1) within pbft mgr before participating within the current period by voting.

justinsnapp: @JakubFornadel  I added 

`assert(pbft_chain_->getPbftChainSize() + 1 == period_);`

into stateOperations_() in pbft and this assertion does not always pass.

JakubFornadel: @justinsnapp 

- no we never advance to the next period until we actually push the block into the chain...
- that assert should not be there: pbft manager runs independently in its own thread and increasing chain / increasing pbft period is not atomic operation. We increment chain size, which does some db writes etc... and then we increment pbft period so there is always a chance that stateOperations_() is called between these 2 operations and such assert would fail

JakubFornadel: Actually, I just realized we dont need `period_` at all as it is always incremented only if chain_size is increment, we can simply use chain_size + 1 as current_pbft_period and thats it -> in such case even the race condition I mentioned above will disappear

JakubFornadel (approved):

kstdl (approved):

PR #1965, opened 2022-08-23
chore: optimize dag block processing

mfrankovi:
Currently once we receive same block from multiple nodes we are connected to we often verify the same block multiple times which significantly affects performance.
At the same time each thread that does that should only complete processing after the dag block has been added to the DAG. Returning before this would create disorder in ordering of incoming dag blocks and might lead to a block missing a tip or pivot.

This change makes a thread wait if the block is already being processed but at the same time allows parallel processing of different DAG blocks.


MatusKysel:
I think this should be handled as dag levels by creating similar solution as  `isDagBlockPacketBlockedByLevel` , because this solution is blocking your TP and if it is implemented on layer lower, we could still use those threads for something

mfrankovi:
> I think this should be handled as dag levels by creating similar solution as `isDagBlockPacketBlockedByLevel`, because this solution is blocking our TP and if it is implemented on a lower layer , we could still use those threads for something

True. I will try to implement it this way

PR #1968, opened 2022-08-24
chore: optimize dag block processing

mfrankovi:
Block processing of same dag blocks concurrently.
For large dag blocks this was causing a major bottleneck with the same block being verified multiple times in parallel on multiple threads since all connected nodes might send it in short time frame.

### DIFF
--- a/libraries/core_libs/network/src/tarcap/threadpool/packets_blocking_mask.cpp
+++ b/libraries/core_libs/network/src/tarcap/threadpool/packets_blocking_mask.cpp
@@ -76,6 +76,27 @@ std::optional<taraxa::level_t> PacketsBlockingMask::getSmallestDagLevelBeingProc
   return {};
 }
 
+void PacketsBlockingMask::setDagBlockBeingProcessed(const PacketData& packet) {
+  // Signature is used as id for the dag block since it is cheaper to get signature than to calculate hash at this point
+  // and signature should be unique per dag block
+  sig_t sig = DagBlock::extract_signature_from_rlp(packet.rlp_);
+
+  // If blocking is working correctly it should not be possible that we already are processing the same block
+  assert(processing_dag_blocks_.find(sig) == processing_dag_blocks_.end());
+
+  processing_dag_blocks_.emplace(std::move(sig), packet.id_);
+}
+
+void PacketsBlockingMask::unsetDagBlockBeingProcessed(const PacketData& packet) {
+  sig_t sig = DagBlock::extract_signature_from_rlp(packet.rlp_);
+
+  // There must be existing block inside processing_dag_blocks_
+  const auto processing_dag_block = processing_dag_blocks_.find(sig);
+  assert(processing_dag_block != processing_dag_blocks_.end());
+
+  processing_dag_blocks_.erase(processing_dag_block);
+}
+
 void PacketsBlockingMask::setDagBlockLevelBeingProcessed(const PacketData& packet) {
   level_t dag_level = DagBlock::extract_dag_level_from_rlp(packet.rlp_);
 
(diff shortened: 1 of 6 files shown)


libraries/core_libs/network/src/tarcap/threadpool/packets_blocking_mask.cpp
```diff
@@ -76,6 +76,28 @@ std::optional<taraxa::level_t> PacketsBlockingMask::getSmallestDagLevelBeingProc
...
+}
+
+void PacketsBlockingMask::unsetDagBlockBeingProcessed(const PacketData& packet) {
+  sig_t sig = DagBlock::extract_signature_from_rlp(packet.rlp_);
+
+  // There must be existing block inside processing_dag_blocks_
+  const auto processing_dag_block = processing_dag_blocks_.find(sig);
+  assert(processing_dag_block != processing_dag_blocks_.end());
+
+  processing_dag_blocks_.erase(sig);
```
MatusKysel, line 98: just a small speed up, I would use iterator `processing_dag_block` for erase as you do not need to search again

libraries/core_libs/network/src/tarcap/threadpool/packets_blocking_mask.cpp
```diff
@@ -76,6 +76,28 @@ std::optional<taraxa::level_t> PacketsBlockingMask::getSmallestDagLevelBeingProc
...
 }
 
+void PacketsBlockingMask::setDagBlockBeingProcessed(const PacketData& packet) {
+  // Signature is used as id for the dag block since it is cheaper to get signature than to calculate hash at this point
+  // and signature should be unique per dag block
+  sig_t sig = DagBlock::extract_signature_from_rlp(packet.rlp_);
+
+  // If blocking is working correctly it should not be possible that we already are processing the same block
+  const auto processing_dag_block = processing_dag_blocks_.find(sig);
+  assert(processing_dag_block == processing_dag_blocks_.end());
```
MatusKysel, line 86: also I would move find inside assert as for Release it will need to result of find

MatusKysel (approved):

kstdl (approved):

libraries/types/dag_block/include/dag/dag_block.hpp
```diff
@@ -33,6 +33,8 @@ class DagBlock {
   vdf_sortition::VdfSortition vdf_;
   mutable addr_t cached_sender_;  // block creater
   mutable util::DefaultConstructCopyableMovable<std::mutex> cached_sender_mu_;
+  static const size_t kLevelPosInRlp{1};
+  static const size_t kSigPosInRlp{6};
```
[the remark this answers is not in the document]
mfrankovi: There is already a big warning in description of the class about this. I think this should be enough:

class DagBlock {
  // !!! Important: Any change in order or types of DagBlock class members will result in broken RLP ctor and some
  // methods
  //                as it is parsed with strong assumption of which member is at which position in rlp bytes
  //                representation. See "DagBlock(dev::RLP const &_rlp)" or "extract_dag_level_from_rlp". If some change
  //                needs to be done in class members, please check carefully all possible dependencies it has like for
  //                example in mentioned methods...

libraries/core_libs/network/src/tarcap/threadpool/packets_blocking_mask.cpp
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/threadpool/packets_blocking_mask.cpp
mfrankovi: Fixed

PR #1971, opened 2022-08-25
chore: improve trx estimates

mfrankovi:
Block proposer will use transactions gas limit for estimates if the limit is under 200k. For any transaction above that the estimate will be run using dry_run_transaction.

This limit of 200k is not part of protocol. Proposer could change the limit and not break protocol. When verifying the estimates in a dag block, we allow for any transaction to have one of the two allowed values:
- transaction gas limit
- estimate of dry_run_transaction

Proposer could change this limit but lowering the limit might fit more transactions in the block but at the same time dag block will be created with a delay which might cause other nodes to create transactions faster so proposer would lose the benefit of having more transactions.

### DIFF
--- a/libraries/core_libs/consensus/src/dag/dag_manager.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag_manager.cpp
@@ -618,11 +618,17 @@ DagManager::VerifyBlockReturnType DagManager::verifyBlock(const DagBlock &blk) {
       return VerifyBlockReturnType::IncorrectTransactionsEstimation;
     }
     for (size_t i = 0; i < trxs_hashes.size(); ++i) {
-      const auto &e = trx_mgr_->estimateTransactionGas((*transactions)[trxs_hashes[i]], propose_period);
-      if (e != trxs_gas_estimations[i]) {
-        return VerifyBlockReturnType::IncorrectTransactionsEstimation;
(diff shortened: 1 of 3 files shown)


MatusKysel (approved):

libraries/core_libs/consensus/src/dag/dag_manager.cpp
```diff
@@ -618,11 +618,17 @@ DagManager::VerifyBlockReturnType DagManager::verifyBlock(const DagBlock &blk) {
       return VerifyBlockReturnType::IncorrectTransactionsEstimation;
     }
     for (size_t i = 0; i < trxs_hashes.size(); ++i) {
-      const auto &e = trx_mgr_->estimateTransactionGas((*transactions)[trxs_hashes[i]], propose_period);
-      if (e != trxs_gas_estimations[i]) {
-        return VerifyBlockReturnType::IncorrectTransactionsEstimation;
+      const auto trx = (*transactions)[trxs_hashes[i]];
+      // Transaction estimation should be either the transactions gas limit or the estimate
+      if (trx->getGas() != trxs_gas_estimations[i]) {
```
kstdl, line 623: We should also check `kEstimateGasLimit` here

kstdl (approved):

PR #1973, opened 2022-08-25
Pbft fixes

JakubFornadel:
## Purpose

Few fixes:

- current pbft period is always chain_size + 1
-  removed two_t_plus_one value from db as we dont need it - we can use current one
- send proposed pbft block before propose vote

### DIFF
--- a/libraries/core_libs/network/src/tarcap/stats/node_stats.cpp
+++ b/libraries/core_libs/network/src/tarcap/stats/node_stats.cpp
@@ -28,14 +28,6 @@ NodeStats::NodeStats(std::shared_ptr<PeersState> peers_state, std::shared_ptr<Pb
       packets_stats_(std::move(packets_stats)),
       thread_pool_(std::move(thread_pool)) {
   LOG_OBJECTS_CREATE("SUMMARY");
-
-  local_pbft_step_prev_interval_ = pbft_mgr_->getPbftStep();
-  const auto [local_pbft_round, local_pbft_period] = pbft_mgr_->getPbftRoundAndPeriod();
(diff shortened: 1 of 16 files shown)


justinsnapp (approved):

libraries/core_libs/network/src/tarcap/stats/node_stats.cpp
```diff
@@ -28,6 +28,14 @@ NodeStats::NodeStats(std::shared_ptr<PeersState> peers_state, std::shared_ptr<Pb
       packets_stats_(std::move(packets_stats)),
       thread_pool_(std::move(thread_pool)) {
   LOG_OBJECTS_CREATE("SUMMARY");
+
+  local_pbft_step_prev_interval_ = pbft_mgr_->getPbftStep();
```
JakubFornadel, line 32: @justinsnapp  you cant do this - some tests, e.g.
```
TEST_F(NetworkTest, transfer_block) {
  auto nw1 = std::make_unique<Network>(g_conf1);
  auto nw2 = std::make_unique<Network>(g_conf2);
```

create network without any managers so this segfaults. I am reverting this commit

justinsnapp: Ok problem was just including it in the constructor then at least without a check for pbft_mgr being initizalized.  We already call pbft_mgr in the logging function though so no reason to completely revert this commit @JakubFornadel

JakubFornadel: you can cherry pick it and fix it in your separate branch then. I wanted to continue with my changes and this broke my branch so I reverted it...

kstdl (approved):

PR #1975, opened 2022-08-26
fix: reward votes replace

mfrankovi:
There was a race condition where if some thread called addRewardVote between calls to:
db_->addLastBlockCertVotesToBatch and vote_mgr_->replaceRewardVotes(std::move(cert_votes))

reward vote from the old period would get saved into Columns::last_block_cert_votes and it would never be deleted.

The fix is that replaceRewardVotes now replace the votes in memory before deleting them from the db. After this only new reward votes are accepted into VoteMgr and old votes are deleted from the db with the return value of replaceRewardVotes.

On node restart last_block_cert_votes vote are verified that they point to the last block in the chain before being stored in memory as reward votes

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -123,7 +123,18 @@ void PbftManager::run() {
     }
     finalize_(std::move(period_data), db_->getFinalizedDagBlockHashesByPeriod(period), period == curr_period);
   }
-  vote_mgr_->replaceRewardVotes(db_->getLastBlockCertVotes());
+  // Verify that last block cert votes point to the last block hash
+  auto last_block_cert_votes = db_->getLastBlockCertVotes();
+  for (auto it = last_block_cert_votes.begin(); it != last_block_cert_votes.end();) {
+    if ((*it)->getBlockHash() != pbft_chain_->getLastPbftBlockHash()) {
+      LOG(log_er_) << "Found invalid last block cert vote: " << **it;
+      db_->removeLastBlockCertVotes((*it)->getHash());
+      it = last_block_cert_votes.erase(it);
+    } else {
+      ++it;
+    }
+  }
+  vote_mgr_->replaceRewardVotes(last_block_cert_votes);
   // Initialize PBFT status
   initialState();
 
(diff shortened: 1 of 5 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1915,7 +1924,8 @@ bool PbftManager::pushPbftBlock_(PeriodData &&period_data, std::vector<std::shar
   }
 
   db_->savePeriodData(period_data, batch);
-  db_->addLastBlockCertVotesToBatch(cert_votes, vote_mgr_->getRewardVotes(), batch);
+  auto reward_votes = vote_mgr_->replaceRewardVotes(cert_votes);
+  db_->addLastBlockCertVotesToBatch(std::move(cert_votes), reward_votes, batch);
```
MatusKysel, line 1928: this move to `addLastBlockCertVotesToBatch` does nothing

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -123,7 +123,16 @@ void PbftManager::run() {
...
     finalize_(std::move(period_data), db_->getFinalizedDagBlockHashesByPeriod(period), period == curr_period);
   }
-  vote_mgr_->replaceRewardVotes(db_->getLastBlockCertVotes());
+  // Verify that last block cert votes point to the last block hash
+  auto last_block_cert_votes = db_->getLastBlockCertVotes();
+  for (auto it = last_block_cert_votes.begin(); it != last_block_cert_votes.end(); it++) {
+    if ((*it)->getBlockHash() != pbft_chain_->getLastPbftBlockHash()) {
+      LOG(log_er_) << "Found invalid last block cert vote: " << **it;
+      db_->removeLastBlockCertVotes((*it)->getHash());
+      last_block_cert_votes.erase(it--);
```
MatusKysel, line 132: is this correct? shouldn't it be it = last_block_cert_votes.erase(it); ?

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -542,10 +542,15 @@ bool VoteManager::checkRewardVotes(const std::shared_ptr<PbftBlock>& pbft_block)
...
-void VoteManager::replaceRewardVotes(std::vector<std::shared_ptr<Vote>>&& cert_votes) {
-  if (cert_votes.empty()) return;
+std::vector<std::shared_ptr<Vote>> VoteManager::replaceRewardVotes(
+    const std::vector<std::shared_ptr<Vote>>& cert_votes) {
+  if (cert_votes.empty()) return {};
 
   std::unique_lock lock(reward_votes_mutex_);
+  std::vector<std::shared_ptr<Vote>> reward_votes;
+  for (const auto& v : reward_votes_) {
+    reward_votes.push_back(v.second);
```
MatusKysel, line 552: here we should do move as we are replacing it later, but not like this is should be `reward_votes = std::move(reward_votes_);`

MatusKysel (approved):

justinsnapp (approved):

PR #1976, opened 2022-08-29
chore: improve trx retrieval

mfrankovi:
On high tps tests on devnet, there is still some slow processing when verifying dag blocks. This happens when verifying DAG blocks which include a lot of transactions that are already finalized.

last_finalized_block_transactions_ was cleared even for blocks that had no transactions or a small number of transactions so instead of keeping only last block transactions in memory we now keep up to 50k recently finalized transactions in memory before clearing it.

### DIFF
--- a/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
@@ -312,19 +312,23 @@ SharedTransactions TransactionManager::getAllPoolTrxs() {
  */
 void TransactionManager::updateFinalizedTransactionsStatus(PeriodData const &period_data) {
   // !!! There is no lock because it is called under std::unique_lock trx_lock(trx_mgr_->getTransactionsMutex());
-  last_finalized_block_transactions_.clear();
-  for (auto const &trx : period_data.transactions) {
-    last_finalized_block_transactions_[trx->getHash()] = trx;
-    if (!nonfinalized_transactions_in_dag_.erase(trx->getHash())) {
-      trx_count_++;
-    } else {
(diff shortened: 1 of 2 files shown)


JakubFornadel (approved):

JakubFornadel:
just wondering why 50k ? Seems pretty big number

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -312,19 +312,23 @@ SharedTransactions TransactionManager::getAllPoolTrxs() {
...
   // !!! There is no lock because it is called under std::unique_lock trx_lock(trx_mgr_->getTransactionsMutex());
-  last_finalized_block_transactions_.clear();
-  for (auto const &trx : period_data.transactions) {
-    last_finalized_block_transactions_[trx->getHash()] = trx;
-    if (!nonfinalized_transactions_in_dag_.erase(trx->getHash())) {
-      trx_count_++;
-    } else {
-      LOG(log_dg_) << "Transaction " << trx->getHash() << " removed from nonfinalized transactions";
+  if (period_data.transactions.size() > 0) {
+    if (period_data.transactions.size() + recently_finalized_transactions_.size() > kRecentlyFinalizedTransactionsMax) {
```
kstdl, line 316: Won't ExpirationCacheMap work better here?

mfrankovi: ExpirationCacheMap is thread safe and since all of this is already under transactions mutex we would be double locking. Probably some kind of more specialized cache would be better here since the current solution at some poiint completely clears the cache but this is just something I wanted to test on devnet to confirm that this is actually the bottleneck for high TPS and then once I confirm it I would try to implement better caching

kstdl (approved):

PR #1980, opened 2022-08-29
fix: sporadic transaction_pool_overflow fails

kstdl:
Fix sporadic fails of `transaction_pool_overflow` test by increasing count of validators. Without this change all consensus is done by one node and other is permanently syncing from it. Because of this we could be in situation when nodes has different chain size and dag syncing isn't working because we are still in pbft syncing process


JakubFornadel:
I just rebased pbft_simplify branch to the develop and this PR seems to be in some strange state where it re creates all of the changed, which are already in pbft_simplify ?

kstdl:
@JakubFornadel should be ok right now

JakubFornadel (approved):

PR #1983, opened 2022-08-30
chore: fix period advancing + synchronous vs asynchronous processing …

JakubFornadel:
## Purpose

There was a bug for networks with delegation_delay set to 0:
advancePeriod() must be called after chain size is increased(updatePbftChain) as well as finalize_ function on final chain is called, otherwise consesnus might get stuck due to infinite waiting inside updateDposState_(), which is waiting for block to be finalized...

This should not affect devnet though

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -989,8 +989,8 @@ void PbftManager::firstFinish_() {
       placeVote(std::move(vote));
     }
   } else {
-    if (auto vote = generateVoteWithWeight(previous_round_next_voted_value_.first, next_vote_type, period,
-                                           round, step_);
+    if (auto vote =
+            generateVoteWithWeight(previous_round_next_voted_value_.first, next_vote_type, period, round, step_);
         vote) {
       LOG(log_nf_) << "Placed first finish next vote for " << previous_round_next_voted_value_.first.abridged()
                    << ", vote weight " << *vote->getWeight() << ", round " << round << ", period " << period
@@ -1732,19 +1732,12 @@ void PbftManager::pushSyncedPbftBlocksIntoChain() {
       }
 
       net->setSyncStatePeriod(period);
-
-      if (executed_pbft_block_) {
-        updateDposState_();
-        // update sortition_threshold and TWO_T_PLUS_ONE
-        updateTwoTPlusOneAndThreshold_();
-        db_->savePbftMgrStatus(PbftMgrStatus::ExecutedBlock, false);
-        executed_pbft_block_ = false;
-      }
     }
   }
 }
 
-void PbftManager::finalize_(PeriodData &&period_data, std::vector<h256> &&finalized_dag_blk_hashes, bool sync) {
+void PbftManager::finalize_(PeriodData &&period_data, std::vector<h256> &&finalized_dag_blk_hashes,
+                            bool synchronous_processing) {
   const auto anchor = period_data.pbft_blk->getPivotDagBlockHash();
 
   auto result = final_chain_->finalize(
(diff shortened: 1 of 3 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1835,16 +1829,17 @@ bool PbftManager::pushPbftBlock_(PeriodData &&period_data, std::vector<std::shar
...
                << ", round: " << getPbftRound();
 
-  finalize_(std::move(period_data), std::move(dag_blocks_order));
+  finalize_(std::move(period_data), std::move(dag_blocks_order), synchronous_processing);
 
   db_->savePbftMgrStatus(PbftMgrStatus::ExecutedBlock, true);
   executed_pbft_block_ = true;
+
+  // Advance pbft consensus period
+  advancePeriod();
```
kstdl, line 1841: maybe this change should be enough? 
If I understand correctly what is happening now, we won't be waiting indefinitely in `updateDposState_` because finalization is occuring in other thread. So when block would be finalized, we will get new values inside of `updateDposState_` and move on

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1701,7 +1701,8 @@ bool PbftManager::pushCertVotedPbftBlockIntoChain_(taraxa::blk_hash_t const &cer
     return false;
   }
 
-  if (!pushPbftBlock_(std::move(period_data_), std::move(current_round_cert_votes), std::move(dag_blocks_order))) {
+  if (!pushPbftBlock_(std::move(period_data_), std::move(current_round_cert_votes), std::move(dag_blocks_order),
+                      true)) {
```
kstdl, line 1705: After this change we are always waiting for block execution before moving to next period, right? It is not good because we are adding some pause to consensus, while we applying block. Even worse, the block execution time could differ from node to node. With this we are desyncronize the nodes

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
MatusKysel, line 1841: I guess better design would be to have all this in call back function of `finalize`

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: true, I looked at it from the perspective when there is no delegation_delay, in which case you simply have to wait for finalization and you can wait either here or in updateDpos...

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: any reason ?

JakubFornadel: actually, it cannot be in callback function if we want to take advantage of asynchronous processing and continue advance period asap

MatusKysel: instead of synchronous execution, you should execute it in callback function and use async exe

MatusKysel (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: ? please re review the code as I pushed some changes because I dont really understand how is what you suggest relevant to this code

JakubFornadel: in other words I dont get it so you push commit with change - that will be faster

PR #1989, opened 2022-08-31
chore: optimize pbft block proposal

mfrankovi:
Remove getting finalized transactions from proposal since it is not needed.
Fill seen_blocks_ cache on startup so that getting dag blocks in pbft block proposal is served from cache and not db
Cache sender in transactions on startup since getSender is expensive to avoid doing it on proposing pbft blocks

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -1404,16 +1404,10 @@ std::shared_ptr<PbftBlock> PbftManager::proposePbftBlock_() {
       }
     }
   }
-  std::vector<trx_hash_t> non_finalized_transactions;
-  auto trx_finalized = db_->transactionsFinalized(trx_hashes);
-  for (uint32_t i = 0; i < trx_finalized.size(); i++) {
-    if (!trx_finalized[i]) {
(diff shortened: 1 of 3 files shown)


kstdl (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1404,16 +1404,9 @@ std::shared_ptr<PbftBlock> PbftManager::proposePbftBlock_() {
...
-  auto trx_finalized = db_->transactionsFinalized(trx_hashes);
-  for (uint32_t i = 0; i < trx_finalized.size(); i++) {
-    if (!trx_finalized[i]) {
-      non_finalized_transactions.emplace_back(trx_hashes[i]);
-    }
-  }
 
-  const auto transactions = trx_mgr_->getNonfinalizedTrx(non_finalized_transactions, true /*sorted*/);
-  non_finalized_transactions.clear();
+  std::vector<trx_hash_t> non_finalized_transactions;
```
MatusKysel, line 1408: you can reserve size based on `transactions`

mfrankovi: Done

PR #1991, opened 2022-08-31
fix: pbft consensus fixes -> make it behave according to the paper

JakubFornadel:
## Purpose

Fixes for multiple bugs:
- we never cleaned previous round next votes when moving to the next period
- firstFinish_ step: 
a) else if condition would pass for round == 1, which it should not. 
b) wrongly used previous_round_next_voted_value_.first == NULL_BLOCK_HASH in conditions as previous_round_next_voted_value_ should never be NULL_BLOCK_HASH -> these condition probably did not work correctly as NULL_BLOCK_HASH was default, but this variable should have either some specific hash or nothing
- secondFInish step: 
a) wrongly used previous_round_next_voted_value_.first == NULL_BLOCK_HASH in conditions as previous_round_next_voted_value_ should never be NULL_BLOCK_HASH -> these condition probably did not work correctly
b) it used if else condition but it was supposed to be if if condition

### DIFF
--- a/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
+++ b/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
@@ -27,7 +27,7 @@ class NextVotesManager {
   /**
    * @brief Clear previous PBFT round next voting type votes
    */
-  void clear();
+  void clearVotes();
 
   /**
    * @brief Check if the vote has been in the map
@@ -54,7 +54,7 @@ class NextVotesManager {
    * @brief Get next voting type votes vote value
    * @return next voting type votes vote value and it's period
    */
-  std::pair<blk_hash_t, uint64_t> getVotedValue() const;
+  std::optional<std::pair<blk_hash_t, uint64_t>> getVotedValue() const;
 
   /**
    * @brief Get previous PBFT round all next voting type votes
@@ -107,8 +107,8 @@ class NextVotesManager {
   std::shared_ptr<FinalChain> final_chain_;
 
   bool enough_votes_for_null_block_hash_;
-  blk_hash_t voted_value_;  // For value is not null block hash
-  uint64_t voted_period_;
+  std::optional<std::pair<blk_hash_t, uint64_t /* period */>> voted_value_;
+
   // <voted PBFT block hash, next votes list that have exactly 2t+1 votes voted at the PBFT block hash>
   // only save votes == 2t+1 voted at same value in map and set
   std::unordered_map<blk_hash_t, std::vector<std::shared_ptr<Vote>>> next_votes_;
@@ -216,10 +216,11 @@ class VoteManager {
    * @brief Check if there are enough next voting type votes to set PBFT to a forward round within period
    * @param period is current pbft period
    * @param two_t_plus_one PBFT 2t+1 is 2/3 of PBFT sortition threshold and plus 1
-   * @return new round if there is enough next votes from prior round, otherwise returns 1 for being in initial round
-   * CONCERN: Was a std::optional, but why not just return 1 if we don't have any votes?  Now that we pass in period
+   * @return optional<pair<new round, 2t+1 next votes>> if there is enough next votes from prior round, otherwise
+   * returns empty optional
    */
-  uint64_t determineRoundFromPeriodAndVotes(uint64_t period, size_t two_t_plus_one);
+  std::optional<std::pair<uint64_t, std::vector<std::shared_ptr<Vote>>>> determineRoundFromPeriodAndVotes(
+      uint64_t period, size_t two_t_plus_one);
 
   // reward votes
   /**
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -378,9 +378,13 @@ bool PbftManager::tryPushCertVotesBlock() {
 bool PbftManager::advancePeriod() {
   resetPbftConsensus(1 /* round */);
 
-  // Move to a new period, cleanup previous period votes
-  const auto new_period = pbft_chain_->getPbftChainSize() + 1;
+  // Cleanup previous period votes in vote manager
+  const auto new_period = getPbftPeriod();
   vote_mgr_->cleanupVotesByPeriod(new_period);
+
+  // Cleanup previous round next votes in next votes manager
+  next_votes_manager_->clearVotes();
+
   LOG(log_nf_) << "Period advanced to: " << new_period << ", round reset to 1";
 
   // Restart while loop...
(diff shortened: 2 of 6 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -741,10 +756,8 @@ void PbftManager::proposeBlock_() {
   auto [round, period] = getPbftRoundAndPeriod();
   LOG(log_dg_) << "PBFT value proposal state in round: " << round << ", period: " << period;
 
-  if (round == 1 || next_votes_manager_->haveEnoughVotesForNullBlockHash()) {
-    if (round > 1) {
-      LOG(log_nf_) << "Previous round " << round - 1 << " had next voted NULL_BLOCK_HASH";
-    }
+  if (round == 1 || previous_round_next_voted_null_block_hash_) {
```
kstdl, line 759: Shouldn't we propose block here only if we doesn't have `previous_round_next_voted_value_`?

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
```diff
@@ -27,7 +27,7 @@ class NextVotesManager {
   /**
    * @brief Clear previous PBFT round next voting type votes
    */
-  void clear();
+  void clearVotes();
```
justinsnapp, line 30: Should this be called clearNextVotes() if only clears next votes?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -808,7 +824,7 @@ void PbftManager::identifyBlock_() {
   auto [round, period] = getPbftRoundAndPeriod();
   LOG(log_dg_) << "PBFT filtering state in round: " << round << ", period: " << period;
 
-  if (round == 1 || next_votes_manager_->haveEnoughVotesForNullBlockHash()) {
+  if (round == 1 || previous_round_next_voted_null_block_hash_) {
```
kstdl, line 827: Same thing here. Shouldn't other voted value(not NULL_HASH) be prioritized here?

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
JakubFornadel: coul be but as it is NextVotesManager it seemed to me clear that it does not have any other votes than next votes

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -741,10 +756,8 @@ void PbftManager::proposeBlock_() {
   auto [round, period] = getPbftRoundAndPeriod();
   LOG(log_dg_) << "PBFT value proposal state in round: " << round << ", period: " << period;
 
-  if (round == 1 || next_votes_manager_->haveEnoughVotesForNullBlockHash()) {
-    if (round > 1) {
-      LOG(log_nf_) << "Previous round " << round - 1 << " had next voted NULL_BLOCK_HASH";
-    }
+  if (round == 1 || previous_round_next_voted_null_block_hash_) {
+    LOG(log_nf_) << "Previous round " << round - 1 << " had next voted NULL_BLOCK_HASH";
```
justinsnapp, line 760: This log line is a bit misleading when round is 1.   Will print in round 0 had next voted null block hash...

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -756,42 +769,45 @@ void PbftManager::proposeBlock_() {
                        << *vote->getWeight() << ", period " << period << ", round " << round << ", step " << step_;
 
           // In propose step, send block before vote as we require block to be present during vote validation
+          // TODO: this is not enough - we should either block votes packets processing by pbft block packets processing
```
justinsnapp, line 772: This comment is vague.  Not enough for _what_?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
justinsnapp: You mean in terms of code execution, not in terms of algorithm logic?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -956,32 +965,29 @@ void PbftManager::firstFinish_() {
...
-    // If cert_voted_block_for_round_ pushed into chain vote for NULL value
-    if (cert_voted_block_for_round_->second < period) {
-      vote_value = NULL_BLOCK_HASH;
-    }
+    // It should never happen that node moved to the next period without cert_voted_block_for_round_ reset
+    assert(cert_voted_block_period == period);
 
-    if (auto vote = generateVoteWithWeight(vote_value, next_vote_type, period, round, step_); vote) {
-      LOG(log_nf_) << "Placed first finish next vote for " << vote_value << ", vote weight " << *vote->getWeight()
+    if (auto vote = generateVoteWithWeight(cert_voted_block, next_vote_type, cert_voted_block_period, round, step_);
```
justinsnapp, line 973: Why are we using this cert_voted_block_period and not just period?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -990,12 +996,24 @@ void PbftManager::firstFinish_() {
       placeVote(std::move(vote));
     }
   } else {
-    if (auto vote =
-            generateVoteWithWeight(previous_round_next_voted_value_.first, next_vote_type, period, round, step_);
+    // TODO: We should vote for any value that we first saw 2t+1 next votes for in previous round -> in current design
```
justinsnapp, line 999: Seems to me that previous implementation was correct, but that these changes are needed only because of the changes made to behavior of previous_round_next_voted_value_

I think new implementation is a bit more convoluted but should be correct.    Not sure how important it is to actually track which value was next voted first (could differ by node anyway).

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -484,14 +484,15 @@ std::optional<VotesBundle> VoteManager::getVotesBundle(uint64_t round, uint64_t
...
+std::optional<std::pair<uint64_t, std::vector<std::shared_ptr<Vote>>>> VoteManager::determineRoundFromPeriodAndVotes(
+    uint64_t period, size_t two_t_plus_one) {
   std::vector<std::shared_ptr<Vote>> votes;
 
   SharedLock lock(verified_votes_access_);
 
   const auto found_period_it = verified_votes_.find(period);
   if (found_period_it == verified_votes_.end()) {
-    return 1;
+    return {};
```
justinsnapp, line 495: I don't really understand why its better to use an optional here, as we can determine that we are in round 1 if we don't have next votes in the period for any round.   And do we we use the votes that are returned?

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -517,13 +518,12 @@ uint64_t VoteManager::determineRoundFromPeriodAndVotes(uint64_t period, size_t t
         LOG(log_nf_) << "Round determined for period " << period << ". Found " << count << " votes for voted value "
                      << voted_value.first << " for round " << round_rit->first << ", step " << step_rit->first;
 
-        next_votes_manager_->updateNextVotes(votes, two_t_plus_one);
-        return round_rit->first + 1;
+        return {std::make_pair(round_rit->first + 1, std::move(votes))};
```
justinsnapp, line 521: I see, you wanted to move this updating of next votes out of this function I guess so thats why you return votes?

justinsnapp (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: this condition does exactly as in paper: 
```
If p = 1 OR (p ≥ 2) AND (i has received 2t + 1 next-votes for ⊥ for period p − 1) ,
then i identifies his leader ` i,p for period p and soft-votes the value v proposed by ` i,p ;
```

previous_round_next_voted_null_block_hash_ is set to true if we saw 2t+1 next votes for NULL_BLOCK_HASH in previous round

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: it is the same, see `assert(cert_voted_block_period == period);`.

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
JakubFornadel: yes we do use votes, which are returned and thats why it is more clear to use optional. Otherwise we would have check if returned round  == 1 and based on that not touch returned votes(which are empty in such case)

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
JakubFornadel: yes exactly

kstdl (approved):

MatusKysel (approved):

PR #1995, opened 2022-09-02
chore: previous round next votes sync imrpovement

mfrankovi:
1. In VotePacketHandler::process process next votes from previous round same as it is done in VotesSyncPacketHandler::process
2. In broadcastPreviousRoundNextVotesBundle rebroadcast all votes even though we think that other nodes know about it
3. broadcastPreviousRoundNextVotesBundle is called in secondFinish each 100 steps, changed it to 20 steps

In first and second finish we rebroadcast pbft block, reward votes and these next votes from previous round. And this is done each N steps, I am not sure if it makes sense to configure how often we do it or we just stick with hard coded values?

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -1087,7 +1087,7 @@ void PbftManager::secondFinish_() {
     syncPbftChainFromPeers_(exceeded_max_steps, NULL_BLOCK_HASH);
   }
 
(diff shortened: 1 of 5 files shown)


kstdl (approved):

kstdl (changes requested):
changes in [vote_packet_handler.cpp](https://github.com/Taraxa-project/taraxa-node/pull/1995/files#diff-f67bfe16cf0f1aa914010fc5deae91e7b4f3cf7de887f036099122ee3028d303) should be inside of next if. So we don't need to validate it as standartVote if it is a nextVote

kstdl (approved):

justinsnapp (approved):

PR #1997, opened 2022-09-05
perf: call compareBlocksAndRewardVotes_ only on selected leader candi…

JakubFornadel:
…date


JakubFornadel:
hmm ok this is copy of https://github.com/Taraxa-project/taraxa-node/pull/1996/files, but it can be combined to be even better

PR #1999, opened 2022-09-05
chore: optimize pbft block proposal and verification

mfrankovi:
Pbft block proposal and verification is time sensitive for pbft steps.

Several recent changes were introduced to remove unneeded transactions processing in this process. To completely remove transactions processing from PbftManager::compareBlocksAndRewardVotes_ I have made changes in how we calculate gas limit for a pbft block. Previously gas limit was calculated for unique non-finalized transactions included in pbft block which required retrieving and processing these transactions for all transactions in pbft block. This is now changed and the gas limit check is done per weight of the dag block. This check is now no longer for unique transactions but for all transactions in dag block. I don't think we lose much with this change and we significantly improve the performance and we reduce the size of the dag block because now it only contains the gas estimation for the block and not for each individual transaction.

Besides this period_data_ was unnecessarily set within compareBlocksAndRewardVotes_ and it only needed to be set when pushing a block into chain.

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -900,7 +900,8 @@ void PbftManager::certifyBlock_() {
   }
   const auto [current_round_soft_voted_block, current_round_soft_votes_period] = *soft_voted_block;
 
-  if (!compareBlocksAndRewardVotes_(current_round_soft_voted_block)) {
+  auto block = pbft_chain_->getUnverifiedPbftBlock(current_round_soft_voted_block);
+  if (!compareBlocksAndRewardVotes_(block)) {
     LOG(log_dg_) << "Incomplete or invalid soft voted block " << current_round_soft_voted_block << ", round " << round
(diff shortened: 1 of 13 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1607,6 +1607,7 @@ std::pair<vec_blk_t, bool> PbftManager::compareBlocksAndRewardVotes_(std::shared
 
   auto dag_order_it = anchor_dag_block_order_cache.find(anchor_hash);
   if (dag_order_it != anchor_dag_block_order_cache.end()) {
+    period_data_.pbft_blk = std::move(pbft_block);
```
JakubFornadel, line 1610: no need to do anything also with period_data_.transactions & period_data_.dag_blocks?

mfrankovi: Refactored PbftManager::compareBlocksAndRewardVotes_, period_data_ is no longer set here

JakubFornadel (approved):

Issue #1977, opened 2022-08-29
Rework (votes) syncing to protect from DDOS


kstdl:
Right now we are accepting all sync packets and trying to process it, so malicious nodes can try to spam us with this packets.
One way to deal with it is to process only sync packets that we requested from peer otherwise mark it as malicious and disconnect from it

PR #2002, opened 2022-09-06
feat: process VotesSyncPacket only from the nodes we requested

kstdl:
#1977 
After this change we are sending previous round next votes from `PbftManager` with regular votes packet. 
Node will accept `VotesSyncPacket` only from the nodes we requested

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -1091,7 +1091,9 @@ void PbftManager::secondFinish_() {
     LOG(log_dg_) << "Node " << node_addr_ << " broadcast next votes for previous round. In period " << period
                  << ", round " << round << " step " << step_;
     if (auto net = network_.lock()) {
-      net->getSpecificHandler<network::tarcap::VotesSyncPacketHandler>()->broadcastPreviousRoundNextVotesBundle();
+      // Broadcast previous round next votes
+      net->getSpecificHandler<network::tarcap::VotePacketHandler>()->broadcastNextVotesBundle(
+          next_votes_manager_->getNextVotes());
     }
     pbft_round_last_broadcast_ = round;
     pbft_step_last_broadcast_ = step_;
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -1092,8 +1092,7 @@ void PbftManager::secondFinish_() {
                  << ", round " << round << " step " << step_;
     if (auto net = network_.lock()) {
       // Broadcast previous round next votes
-      net->getSpecificHandler<network::tarcap::VotePacketHandler>()->broadcastNextVotesBundle(
-          next_votes_manager_->getNextVotes());
+      net->getSpecificHandler<network::tarcap::VotePacketHandler>()->broadcastPreviousRoundNextVotesBundle();
     }
     pbft_round_last_broadcast_ = round;
     pbft_step_last_broadcast_ = step_;
--- a/libraries/core_libs/network/include/network/tarcap/packets_handlers/common/packet_handler.hpp
+++ b/libraries/core_libs/network/include/network/tarcap/packets_handlers/common/packet_handler.hpp
@@ -39,7 +39,7 @@ class PacketHandler {
    */
   void processPacket(const PacketData& packet_data);
 
-  void requestPbftNextVotesAtPeriodRound(dev::p2p::NodeID const& peerID, uint64_t pbft_period, uint64_t pbft_round,
+  void requestPbftNextVotesAtPeriodRound(const dev::p2p::NodeID& peerID, uint64_t pbft_period, uint64_t pbft_round,
                                          size_t pbft_previous_round_next_votes_size);
 
  private:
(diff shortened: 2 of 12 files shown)


libraries/core_libs/network/include/network/tarcap/packets_handlers/common/packet_handler.hpp
```diff
@@ -39,6 +39,9 @@ class PacketHandler {
    */
   void processPacket(const PacketData& packet_data);
 
+  void requestPbftNextVotesAtPeriodRound(dev::p2p::NodeID const& peerID, uint64_t pbft_period, uint64_t pbft_round,
```
JakubFornadel, line 42: const first

libraries/core_libs/network/src/tarcap/packets_handlers/common/packet_handler.cpp
```diff
@@ -117,4 +117,13 @@ void PacketHandler::disconnect(dev::p2p::NodeID const& node_id, dev::p2p::Discon
   }
 }
 
+void PacketHandler::requestPbftNextVotesAtPeriodRound(dev::p2p::NodeID const& peerID, uint64_t pbft_period,
```
JakubFornadel, line 120: const first

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
```diff
@@ -24,6 +24,14 @@ void VotesSyncPacketHandler::validatePacketRlpFormat([[maybe_unused]] const Pack
 }
 
 void VotesSyncPacketHandler::process(const PacketData &packet_data, const std::shared_ptr<TaraxaPeer> &peer) {
+  if (!peer->votes_sync_requested_) {
+    LOG(log_er_) << "Received VotesSyncPacket from peer(" << packet_data.from_node_id_
+                 << ") we did not request. The peer may be a malicious player, will be disconnected";
+    disconnect(packet_data.from_node_id_, dev::p2p::UserReason);
```
JakubFornadel, line 30: maybe we should use here `handleMaliciousSyncPeer` instead of just disconnect ?

libraries/core_libs/network/include/network/tarcap/packets_handlers/common/packet_handler.hpp
JakubFornadel: why is this function declared in PacketHandler, when it is used only in VotesPacketHandler ?

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1091,7 +1091,9 @@ void PbftManager::secondFinish_() {
     LOG(log_dg_) << "Node " << node_addr_ << " broadcast next votes for previous round. In period " << period
                  << ", round " << round << " step " << step_;
     if (auto net = network_.lock()) {
-      net->getSpecificHandler<network::tarcap::VotesSyncPacketHandler>()->broadcastPreviousRoundNextVotesBundle();
```
JakubFornadel, line 1094: but it actually does broadcast only previous round next votes or not ? broadcastNextVotesBundle evokes it sends also current round next votes

libraries/core_libs/network/include/network/tarcap/packets_handlers/common/packet_handler.hpp
kstdl: It is also used in `validateStandardVote` when we are requesting votes. I think it was just missed after merge.

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
kstdl: After this change it broadcasts bundle that we are passing to it, so I think new naming `broadcastNextVotesBundle` is better

libraries/core_libs/network/src/tarcap/packets_handlers/votes_sync_packet_handler.cpp
kstdl: This method is from other packet handler, so I just added `set_peer_malicious` here and changed log

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: yes so what was this change actually good for when we dont pass there anything else than previous round next votes ? + it is always called with `next_votes_manager_->getNextVotes()`, why not to call it inside the function ?

libraries/core_libs/network/include/network/tarcap/packets_handlers/common/packet_handler.hpp
JakubFornadel: then it should be in ExtVotesPacketHandler

kstdl: It is used in `StatusPacketHandler : ExtSyncingPacketHandler : PacketHandler` and in `ExtVotesPacketHandler : PacketHandler`. So it couldn't only be in `ExtVotesPacketHandler`

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
kstdl: There was no `NextVotesManager` in this packet handler before the latest rebase. And it wasn't worth to add it there just because of this one call. So I renamed it back and moved this call back inside of this method

PR #2003, opened 2022-09-07
fix: reward votes race condition

mfrankovi:
We have a following race condition which keeps happening all the time on devnet on every nodes:
Node received 2t+1 cert votes and it is ready to move to next period
Node receives additional cert vote and starts processing it in VotePacketHandler::process and starts the processing as a normal cert vote
While vote is being processed, node pushes the pbft block to the chain and moves to the next period
Vote validation now reports that cert vote is invalid since it started processing it as cert vote and not as reward vote

Example of logs:
PBFT_VOTE_PH [2022-09-07 06:34:39.196487] DEBUG: Received PBFT vote #56e8d471…
PBFT_MGR [2022-09-07 06:34:39.206854] DEBUG: Reset PBFT consensus to: round 1, period 28013, step 1, and resetting clock.
PBFT_VOTE_PH [2022-09-07 06:34:39.213300] WARN: Vote 56e8d471… validation failed. Err: Invalid period: Vote period: 28012, current pbft period: 28013
PBFT_VOTE_PH [2022-09-07 06:34:39.411146] DEBUG: Received PBFT vote #56e8d471…
PBFT_VOTE_PH [2022-09-07 06:34:39.414010] DEBUG: Received vote #56e8d471… (from 1755f430…) already seen.
VOTE_MGR [2022-09-07 06:34:40.530379] ERROR: Missing reward vote #56e8d471…
VOTE_MGR [2022-09-07 06:34:40.530825] ERROR: Missing reward vote #56e8d471…
VOTE_MGR [2022-09-07 06:34:41.030508] ERROR: Missing reward vote #56e8d471…

The fix consists of: 
1. having the packet handler and validate methods use the same period/round when validating votes
2. Inside of VoteManager::addVerifiedVote use case for handling reward votes is added when period has increased between vote validation and inserting the vote.  

This PR also includes a minor fix in PbftSyncPacketHandler::process which cause sync_five_nodes to fail sometimes and it could delay pbft syncing.

### DIFF
--- a/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
+++ b/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
@@ -316,6 +316,8 @@ class VoteManager {
                                          blk_hash_t,
                                          std::pair<uint64_t, std::unordered_map<vote_hash_t, std::shared_ptr<Vote>>>>>>>
       verified_votes_;
+  // Current period, it should only be used under verified_votes_access_ mutex
+  uint64_t period_;
   mutable boost::shared_mutex verified_votes_access_;
 
   // <PBFT period, <PBFT round, <PBFT step, <voter address, pair<vote 1, vote 2>>><>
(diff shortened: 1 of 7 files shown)


JakubFornadel:
Nice catch - hopefully we finally solved the problem with reward votes. Few notes:

- I think this design makes sense only fr standard vote
- for reward & next votes if you use old period (it was changed in the meantime) - it should not pass the validation
- I actually changed the votes processing and thanks to those changes (that would be merged as part of fix for new ddos attack) I was able to fix this race condition in like 5 lines of code (check this PR https://github.com/Taraxa-project/taraxa-node/pull/2004) so I would maybe merge that one

PR #2013, opened 2022-09-09
chore: pbft sync while block cert voted

mfrankovi:
Checking reward votes in PbftSyncPacketHandler can sometimes fail if we are moving to the next period at the same time which should not be handled as malicious behaviors. If it does fail, a check is added if we actually moved to next period and in that case we restart syncing because it is very likely that we are already synced and restartSyncingPbft will verify this.


JakubFornadel (approved):

justinsnapp (approved):
LGTM

PR #2015, opened 2022-09-09
do not force push chart (develop)

rjonczy:


justinsnapp (approved):
Looks ok.  Perhaps PR should explain the reason for the change though for general recollection. :-)

JakubFornadel (approved):

PR #2029, opened 2022-09-15
chore: optimize transactions processing

mfrankovi:
When transactions are gossiped around the network it is normal if a node is connected to lot of other nodes that it will receive the same transaction multiple times coming from different nodes. It is even probable that it will receive it from almost every node it is connected to. Deserializing transactions (converting transaction rlp to Transaction object) takes about 20 millisseconds per 1000 transactions which is not that small considering that even with 1000 new transactions per second appearing on the network we might receive the same transaction a lot of times, so for instance an RPC node on devnet connected to 20 other nodes will probably receive 10000 transactions per second although it will actually be only 1000 unique transactions per second.

To avoid this excessive processing we add transaction hash to the transaction packet. Deserializing just transactions hashes of all the transactions instead of deserializing the full transaction and calculating the hash, allows us to quickly remove all the known transactions from further processing and to quickly marked transaction as known for the peer we received it from so that some other thread might not unnecessarily send transactions back to the peer we received it from.

The downside of this approach is that we are increasing the size of the transaction packet by adding the hash, but with the faster processing we should reduce the actual number of transactions/packets being gossiped so the actual total size of data being sent might even be reduced but this would have to be proved by additional testing.

-------------------

On processing transaction packet that could contain up to 1000 transactions inside, so far node would first verified all the transactions and only then inserted them into transactions pool. This caused unneeded delay in transactions insertion which can cause us to process the same transactions on other threads in parallel and delay gossiping already processed transactions. There was no gain in inserting transactions in bunch. Now as soon as transaction is verified it is inserted in the pool. 

-------------------

Transaction packet max size is reduced from 1000 transactions to 500 to increase parallel processing. Multiple packets are processed in parallel on packet handler threads but transactions within the same packet are processed in series.

### DIFF
--- a/libraries/core_libs/consensus/include/transaction/transaction_manager.hpp
+++ b/libraries/core_libs/consensus/include/transaction/transaction_manager.hpp
@@ -30,7 +30,7 @@ class FullNode;
  * transactions and all transactions state change
  *
  * Incoming new transactions can be verified with verifyTransaction functions and than inserted in the transaction pool
- * with insertValidatedTransactions. Transactions are kept in transactions memory pool until they are included in a
+ * with insertValidatedTransaction. Transactions are kept in transactions memory pool until they are included in a
  * proposed dag block or received in an incoming dag block. Transaction verification consist of:
  * - Verifying the format
  * - Verifying signature
@@ -92,13 +92,14 @@ class TransactionManager : public std::enable_shared_from_this<TransactionManage
   void blockFinalized(uint64_t block_number);
 
   /**
-   * @brief Inserts batch of verified transactions to transaction pool
+   * @brief Inserts verified transaction to transaction pool
    *
-   * @note Some of the transactions might be already processed -> they are not processed and inserted again
-   * @param txs transactions to be processed
-   * @return number of successfully inserted unseen transactions
+   * @note transaction might be already processed -> they are not processed and inserted again
+   * @param tx transaction to be processed
+   * @param status transaction status
+   * @return true if successfully inserted unseen transactions
    */
-  uint32_t insertValidatedTransactions(std::vector<std::pair<std::shared_ptr<Transaction>, TransactionStatus>> &&txs);
+  bool insertValidatedTransaction(std::shared_ptr<Transaction> &&tx, const TransactionStatus status);
 
   /**
    * @param trx_hash transaction hash
--- a/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
@@ -97,7 +97,8 @@ std::pair<bool, std::string> TransactionManager::insertTransaction(const std::sh
     return {false, reason};
   }
 
-  if (insertValidatedTransactions({{trx, status}})) {
+  auto transaction = trx;
+  if (insertValidatedTransaction(std::move(transaction), status)) {
     return {true, "Can not insert transactions"};
   } else {
     const auto period = db_->getTransactionPeriod(trx->getHash());
@@ -109,18 +110,8 @@ std::pair<bool, std::string> TransactionManager::insertTransaction(const std::sh
   }
 }
 
-uint32_t TransactionManager::insertValidatedTransactions(
-    std::vector<std::pair<std::shared_ptr<Transaction>, TransactionStatus>> &&txs) {
-  std::vector<std::pair<std::shared_ptr<Transaction>, TransactionStatus>> unseen_txs;
-  std::vector<trx_hash_t> txs_hashes;
-
-  if (txs.empty()) {
-    return 0;
-  }
-
-  txs_hashes.reserve(txs.size());
-  std::transform(txs.begin(), txs.end(), std::back_inserter(txs_hashes),
-                 [](const auto &t) { return t.first->getHash(); });
+bool TransactionManager::insertValidatedTransaction(std::shared_ptr<Transaction> &&tx, const TransactionStatus status) {
+  const auto trx_hash = tx->getHash();
 
   // This lock synchronizes inserting and removing transactions from transactions memory pool with database insertion.
   // It is very important to lock checking the db state of transaction together with transaction pool checking to be
@@ -128,27 +119,14 @@ uint32_t TransactionManager::insertValidatedTransactions(
   std::unique_lock transactions_lock(transactions_mutex_);
   const auto last_block_number = final_chain_->last_block_number();
 
-  // Check the db with a multiquery if transactions are really new
-  auto db_seen = db_->transactionsInDb(txs_hashes);
-  for (uint32_t i = 0; i < txs_hashes.size(); i++) {
-    if (!db_seen[i]) {
-      unseen_txs.push_back(std::move(txs[i]));
-    } else {
-      // In case we received a new tx that is already in db, mark it as known in cache
-      transactions_pool_.markTransactionKnown(txs[i].first->getHash());
-    }
+  // Check the db with if transaction is really new
+  if (db_->transactionInDb(trx_hash)) {
+    transactions_pool_.markTransactionKnown(trx_hash);
+    return false;
   }
 
-  size_t trx_inserted_count = 0;
-  // Save transactions in memory pool
-  for (auto &trx : unseen_txs) {
-    auto tx_hash = trx.first->getHash();
-    LOG(log_dg_) << "Transaction " << tx_hash << " inserted in trx pool";
-    if (transactions_pool_.insert(std::move(trx), last_block_number)) {
-      trx_inserted_count++;
-    }
-  }
-  return trx_inserted_count;
+  LOG(log_dg_) << "Transaction " << trx_hash << " inserted in trx pool";
+  return transactions_pool_.insert(std::move(tx), status, last_block_number);
 }
 
 unsigned long TransactionManager::getTransactionCount() const {
@@ -344,7 +322,7 @@ void TransactionManager::moveNonFinalizedTransactionsToTransactionsPool(std::uno
     auto trx = nonfinalized_transactions_in_dag_.find(trx_hash);
     if (trx != nonfinalized_transactions_in_dag_.end()) {
       db_->removeTransactionToBatch(trx_hash, write_batch);
-      transactions_pool_.insert({std::move(trx->second), TransactionStatus::Verified});
+      transactions_pool_.insert(std::move(trx->second), TransactionStatus::Verified);
       nonfinalized_transactions_in_dag_.erase(trx);
     }
   }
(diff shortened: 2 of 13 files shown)


libraries/core_libs/consensus/include/transaction/transaction_manager.hpp
```diff
@@ -92,13 +92,13 @@ class TransactionManager : public std::enable_shared_from_this<TransactionManage
...
    *
-   * @note Some of the transactions might be already processed -> they are not processed and inserted again
-   * @param txs transactions to be processed
-   * @return number of successfully inserted unseen transactions
+   * @note transaction might be already processed -> they are not processed and inserted again
+   * @param tx transaction to be processed
+   * @return true if successfully inserted unseen transactions
    */
-  uint32_t insertValidatedTransactions(std::vector<std::pair<std::shared_ptr<Transaction>, TransactionStatus>> &&txs);
+  bool insertValidatedTransaction(std::pair<std::shared_ptr<Transaction>, TransactionStatus> &&tx);
```
kstdl, line 101: Why is it pair, but not 2 args?

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -109,46 +109,26 @@ std::pair<bool, std::string> TransactionManager::insertTransaction(const std::sh
...
-  size_t trx_inserted_count = 0;
-  // Save transactions in memory pool
-  for (auto &trx : unseen_txs) {
-    auto tx_hash = trx.first->getHash();
-    LOG(log_dg_) << "Transaction " << tx_hash << " inserted in trx pool";
-    if (transactions_pool_.insert(std::move(trx), last_block_number)) {
-      trx_inserted_count++;
-    }
+  LOG(log_dg_) << "Transaction " << trx_hash << " inserted in trx pool";
+  if (transactions_pool_.insert(std::move(tx), last_block_number)) {
```
JakubFornadel, line 128: you could use insert as return ? :
return transactions_pool_.insert(std::move(tx), last_block_number)

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
```diff
@@ -17,36 +17,57 @@ TransactionPacketHandler::TransactionPacketHandler(std::shared_ptr<PeersState> p
...
+    throw InvalidRlpItemsCountException(packet_data.type_str_, items, kTransactionPacketItemCount);
+  }
+  auto hashes_count = packet_data.rlp_[0].itemCount();
+  auto trx_count = packet_data.rlp_[1].itemCount();
+
+  if (hashes_count != trx_count) {
+    throw InvalidRlpItemsCountException(packet_data.type_str_, hashes_count, trx_count);
+  }
+  if (hashes_count == 0 || hashes_count > kMaxTransactionsInPacket) {
+    throw InvalidRlpItemsCountException(packet_data.type_str_, hashes_count, kMaxTransactionsInPacket);
```
JakubFornadel, line 30: So I find the rlp structure of TransactionPacket little bit weird. Why dont we simply have array of pairs<tx_hash, tx_object> so it would look like this:
[
[tx_hash, [tx_obj_var1, tx_obj_var2, ....]]
]

So when you validate it, you check the size of rlp to be <= kMaxTransactionsInPacket as before
Then when you start parsing insice packet handler, you iterate over the outer rlp list and for each item(tx pair), you check if tx_rlp.itemsCount() == 2 and then you parse only tx hash (tx_rlp[0]), check it against caches etc... and if it is not known you will parse actual object tx_rlp[1] and process it

mfrankovi: so the reason why I went for this structure is that it is costly to iterate back and forth over rlp, and it is common that we get a lot of transactions more than once so in many cases I will only be looking for transactions hashes and not the full transactions, so it is fastest to process the first chunk of only hashes first and go to the second chunk where transactions are only for those transactions that we need

mfrankovi: rlp does not have random access, it iterates to get to the position it needs

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
```diff
@@ -126,33 +141,40 @@ void TransactionPacketHandler::periodicSendTransactions(SharedTransactions &&tra
...
 
   for (auto &it : transactions_to_send) {
-    sendTransactions(it.first, it.second);
+    sendTransactions(it.first, it.second.first, it.second.second);
   }
-  for (auto &it : transactions_hash_to_send) {
-    for (auto &it2 : it.second) {
+  for (auto &it : transactions_to_send) {
+    for (auto &it2 : it.second.first) {
       peers[it.first]->markTransactionAsKnown(it2);
```
JakubFornadel, line 157: this marking should be inside sendTransactions function and only if sealAndSend returned true

libraries/core_libs/consensus/include/transaction/transaction_manager.hpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
```diff
@@ -126,33 +141,40 @@ void TransactionPacketHandler::periodicSendTransactions(SharedTransactions &&tra
...
     }
   }
 }
 
 void TransactionPacketHandler::sendTransactions(dev::p2p::NodeID const &peer_id,
+                                                std::vector<trx_hash_t> const &transactions_hashes,
                                                 std::vector<taraxa::bytes> const &transactions) {
```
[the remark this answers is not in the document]
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
```diff
@@ -17,36 +17,57 @@ TransactionPacketHandler::TransactionPacketHandler(std::shared_ptr<PeersState> p
...
+  trx_hashes.reserve(transaction_count);
 
-    try {
-      transaction = std::make_shared<Transaction>(packet_data.rlp_[tx_idx].data().toBytes());
-    } catch (const Transaction::InvalidSignature &e) {
-      throw MaliciousPeerException("Unable to parse transaction: " + std::string(e.what()));
-    }
+  // First extract only transaction hashes
+  for (const auto &trx_hash_rlp : packet_data.rlp_[0]) {
+    const auto trx_hash = trx_hash_rlp.toHash<trx_hash_t>();
```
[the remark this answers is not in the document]
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
```diff
@@ -106,8 +121,8 @@ void TransactionPacketHandler::onNewTransactions(
 }
 
 void TransactionPacketHandler::periodicSendTransactions(SharedTransactions &&transactions) {
-  std::unordered_map<dev::p2p::NodeID, std::vector<taraxa::bytes>> transactions_to_send;
-  std::unordered_map<dev::p2p::NodeID, std::vector<trx_hash_t>> transactions_hash_to_send;
+  std::unordered_map<dev::p2p::NodeID, std::pair<std::vector<trx_hash_t>, std::vector<taraxa::bytes>>>
+      transactions_to_send;
```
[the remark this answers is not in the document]
mfrankovi: we send the same transaction to multiple peers, if I would do it this way, would I not be serializing transaction to bytes for each peer so it would be done multiple times?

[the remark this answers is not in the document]
JakubFornadel: good and valid point. But we are doing it even now if you check the code -> this `transactions_to_send[peer.first].second.push_back(trx->rlp());` is called for each peer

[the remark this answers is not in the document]
mfrankovi: actually I see now that we cache the rlp inside of transaction object so your approach is good afterall. I will change it

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
JakubFornadel: It would look like this:
```
for (size_t tx_idx = 0; tx_idx < packet_data.rlp_.itemCount(); tx_idx++) {
  const auto& tx_rlp = packet_data.rlp_[tx_idx];
  if (tx_rlp.itemCount() != 2) {
  	throw MaliciousPeerException(msg);
  }
  
  const auto tx_hash = tx_rlp[0].toHash();
  
  // Do all the checks with tx hash, e.g.
  if (isKnownTx(tx_hash)) {
    continue;
  }
  
  const auto tx_obj = std::make_shared<Transaction>(tx_rlp_[1]);

  // Process tx obj
}
```

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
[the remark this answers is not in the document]
mfrankovi: Fixed

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
mfrankovi: In effort to reduce unnecessary gossip of transactions my goal was to mark transaction as known as soon as we receive it from the peer. So if we receive 1000 transactions, I want to mark these transactions as known for the peer we received it from because there is a great chance that a lot of these transactions we already have and we are about to send it to the same peer we just received it from. If I were to do it in the loop you proposed, deserializing transactions would slow this down. But if I were to use your format and only get hashes in one pass and then transactions in another pass that would be slower. This is what I just measured and for 1000 transactions and doing two passes the difference is this:
Elapsed method 1 : 955 microseconds
Elapsed method 2 : 1762 microseconds

But which ever format we choose I would definitely recommend first marking peer transction as known before doing the deserialization of transactions.

kstdl (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
mfrankovi: if needed we can address this in a separate PR, resolving to merge this PR

JakubFornadel: @mfrankovi  yeah so we do send votes periodically every 100ms, we do not gossip it with each incoming tx packet. So if the difference between these 2 methods is 800us, I think it is not worth it and we should choose the approach, which is better from design/flow point of view because saving 800us will have almost zero effect on marking txs as known if we send it txs every 100ms....

PR #2045, opened 2022-09-19
chore: votes logging and minor optimization

mfrankovi:
I have noticed while doing stress tests some propose votes are not gossiped to all nodes. My suspicion is that in ExtVotesPacketHandler::onNewPbftVote we skip sending this vote because of the state of peer.second->pbft_chain_size_ and peer.second->pbft_round_ so I added additional logging there to confirm this.

I noticed in the logs on devnet that we were logging received same votes multiple times. I moved the check for duplicate votes before the logs and before deserialization of pbft block.

### DIFF
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
@@ -268,6 +268,7 @@ void ExtVotesPacketHandler::onNewPbftVote(const std::shared_ptr<Vote> &vote, con
 
   for (const auto &peer : peers_state_->getAllPeers()) {
     if (peer.second->syncing_) {
+      LOG(log_dg_) << " PBFT vote " << vote->getHash() << " not sent to " << peer.first << " peer syncing";
       continue;
     }
 
@@ -279,10 +280,14 @@ void ExtVotesPacketHandler::onNewPbftVote(const std::shared_ptr<Vote> &vote, con
     if (!(vote->getType() == cert_vote_type && vote->getBlockHash() == rewards_votes_block.first)) {
       // CONCERN ... should we be using a period stored in peer state?
       if (peer.second->pbft_chain_size_ > vote->getPeriod() - 1) {
+        LOG(log_dg_) << " PBFT vote " << vote->getHash() << " not sent to " << peer.first
+                     << " peer chain size: " << peer.second->pbft_chain_size_;
         continue;
       }
 
       if (peer.second->pbft_round_ > vote->getRound()) {
+        LOG(log_dg_) << " PBFT vote " << vote->getHash() << " not sent to " << peer.first
+                     << " peer round: " << peer.second->pbft_round_;
         continue;
       }
     }
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
@@ -33,6 +33,7 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
 
   std::vector<std::shared_ptr<Vote>> votes;
   std::vector<std::shared_ptr<Vote>> previous_next_votes;
+  vote_hash_t vote_hash;
   for (size_t i = 0; i < packet_data.rlp_.itemCount(); i++) {
     // TODO[2031]: the way we differentiate now between receiving single vote with some optional data vs
     //             batch of simple votes is ugly now. We could change the way we use VotePacket vs VotesSyncPacket to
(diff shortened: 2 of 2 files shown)


libraries/core_libs/network/src/tarcap/packets_handlers/common/ext_votes_packet_handler.cpp
```diff
@@ -279,10 +279,14 @@ void ExtVotesPacketHandler::onNewPbftVote(const std::shared_ptr<Vote> &vote, con
     if (!(vote->getType() == cert_vote_type && vote->getBlockHash() == rewards_votes_block.first)) {
       // CONCERN ... should we be using a period stored in peer state?
       if (peer.second->pbft_chain_size_ > vote->getPeriod() - 1) {
+        LOG(log_dg_) << " PBFT vote " << vote->getHash() << " not sent to " << peer.first
```
kstdl, line 282: I think it is worth to add such logs to ifs above(especially syncing check). For me it looks like more probable to fail there

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -64,20 +74,19 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
...
     }
 
-    const auto vote_hash = vote->getHash();
-
-    // Synchronization point in case multiple threads are processing the same vote at the same time
-    if (!seen_votes_.insert(vote_hash)) {
-      LOG(log_dg_) << "Received vote " << vote_hash << " (from " << packet_data.from_node_id_.abridged()
-                   << ") already seen.";
-      continue;
-    }
```
JakubFornadel, line 79: why was this moved from here ?

JakubFornadel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -49,6 +50,15 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
...
 
       vote = std::make_shared<Vote>(vote_with_block_rlp[0]);
+      vote_hash = vote->getHash();
+
+      // Synchronization point in case multiple threads are processing the same vote at the same time
+      if (!seen_votes_.insert(vote_hash)) {
+        LOG(log_tr_) << "Received vote " << vote_hash << " (from " << packet_data.from_node_id_.abridged()
+                     << ") already seen.";
+        continue;
+      }
```
JakubFornadel, line 60: it is not worth it to put it here. The code is not clean right now but if we enter this if branch:
```
    // There is included also PBFT block as optional value in Vote packet
    if (packet_data.rlp_.itemCount() == 1 && packet_data.rlp_[0].itemCount() == kVotePacketWithBlockSize &&
        packet_data.rlp_[0][0].isList()) {
```

there is only single vote so parsing 1 vote does not delay marking it as seen

libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
mfrankovi: I noticed that we have a lot of these repeating logs. If we are connected to 20 nodes, we have 20 of these for each received vote:
Taraxa_N1_09192022_144827_06792.log:0x00007fd1b37fe640 10c405dc… PBFT_VOTE_PH [2022-09-19 14:48:53.200986] DEBUG: Received PBFT vote #1e53029a… with PBFT block #65eef53e…
Taraxa_N1_09192022_144827_06792.log:0x00007fd1b37fe640 10c405dc… PBFT_VOTE_PH [2022-09-19 14:48:53.201078] DEBUG: Received vote #1e53029a… (from d8c2b208…) already seen.
Taraxa_N1_09192022_144827_06792.log:0x00007fd1c9ffb640 10c405dc… PBFT_VOTE_PH [2022-09-19 14:48:53.265493] DEBUG: Received PBFT vote #1e53029a… with PBFT block #65eef53e…
Taraxa_N1_09192022_144827_06792.log:0x00007fd1c9ffb640 10c405dc… PBFT_VOTE_PH [2022-09-19 14:48:53.265692] DEBUG: Received vote #1e53029a… (from 58d0c916…) already seen.
Taraxa_N1_09192022_144827_06792.log:0x00007fd1c9ffb640 10c405dc… PBFT_VOTE_PH [2022-09-19 14:48:53.269122] DEBUG: Received PBFT vote #1e53029a… with PBFT block #65eef53e…
.........................

mfrankovi: Besides the logging if we receive a vote we already know, I think it is better not to construct a pbft block. Now we construct same pbft block 20 times

PR #2058, opened 2022-09-21
chore: step end time

mfrankovi:
Original calculation for end_time_for_step was not good because LAMBDA_ms is not constant which could make several steps move very quickly and then some steps taking a very long time. 

New calculation only uses current LAMBDA to increase the time spent in the current step.

Reason for this:
loop_back_finish_state_ = elapsed_time_in_step > (int64_t)(2 * (LAMBDA_ms - POLLING_INTERVAL_ms));
being two times lambda is because this covers two steps since the delay is only added in second finish.

Besides this vote_accepting_steps is increased to 100 to prevent issues when nodes are down for a while and steps are not synchronized.

### DIFF
--- a/libraries/cli/include/cli/config_jsons/default.json
+++ b/libraries/cli/include/cli/config_jsons/default.json
@@ -15,7 +15,7 @@
   "deep_syncing_threshold": 10,
   "vote_accepting_periods": 5,
   "vote_accepting_rounds": 5,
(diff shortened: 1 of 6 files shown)


JakubFornadel (changes requested):
merge this PR into v2.0.0-alpha, not develop

JakubFornadel (approved):

PR #2066, opened 2022-09-21
feat: rebroadcat soft, next and reward votes also byt non-consensus n…

JakubFornadel:
## Purpose

Maybe not ok - this change would rebroadcast soft votes & reward votes only if node gets stalled, steps > 13 but originally it was in each second finish step


mfrankovi:
> ## Purpose
> Maybe not ok - this change would rebroadcast soft votes & reward votes only if node gets stalled, steps > 13 but originally it was in each second finish step
> 
> ## How does the solution address the problem
> ## Changes made

So previously in second finish on each step we would call onNewPbftVotes for soft votes and sendrewardvotes for reward vote but this was not a real rebroadcast because this would only send soft and reward votes to peer that we did not send them before. In most cases this will actually not do anything, only if some new node is connected or some node just got synced we would actually send something.

The second part was this broadcastPreviousRoundNextVotesBundle which we only did every 20 steps which is a true rebroadcast because we send these votes to all peers including the ones we already sent it before.

 I think for v2.0.0 I would leave the functionality as is and use it this way as it was before and than later on develop we refactor this. So just send it in the same intervals as they were sent before.
 
As for later refactoring I do think we need these two functionalities:
1. Sending votes to blocks that did not originally receive it because they were syncing or because they just got connected to the node/network in the middle of the round. These nodes will not have votes marked as known, so we only send to those that are not marked as known and this is something that we can check often or we refactor it to check at the actual moment node gets connected or gets synced 
2. Backup solution if we are stalled after X steps (not too often) where we rebroadcast every votes (I would rebroadcast previous round next votes, current round soft votes and the pbft block) to every connected node even if we did send it before

This option 2. ideally if we fixed all the bugs and race condition should never be triggered but it is good to have it just in case

PR #2089, opened 2022-09-30
deployments in release cycle

rjonczy:
## Purpose

draft

draft

## Changes made

draft

### DIFF
--- a/doc/release_cycle.md
+++ b/doc/release_cycle.md
@@ -5,8 +5,10 @@ For now we decided to go with 6 week release cycle. At the beginning of each rel
 
 Versioning is based on [Semantic Versioning 2.0.0](https://semver.org/). \
 
-Release branch annotation: `release/vX.Y.Z` \
-Release tag annotation: `vX.Y.Z`
+Release branch annotation: `release/vX.Y` (in branch name we use only `MAJOR`.`MINOR`, we don't specify `PATCH`) \
+Release tag annotation: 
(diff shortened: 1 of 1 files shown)


doc/release_cycle.md
```diff
@@ -5,8 +5,10 @@ For now we decided to go with 6 week release cycle. At the beginning of each rel
 
 Versioning is based on [Semantic Versioning 2.0.0](https://semver.org/). \
 
-Release branch annotation: `release/vX.Y.Z` \
-Release tag annotation: `vX.Y.Z`
+Release branch annotation: `release/vX.Y` (in branch name we use only `MAJOR`.`MINOR`, we don't specify `PATCH`) \
```
JakubFornadel, line 8: why not ?

doc/release_cycle.md
```diff
@@ -5,8 +5,10 @@ For now we decided to go with 6 week release cycle. At the beginning of each rel
...
+Release tag annotation: 
+- testnet: `vX.Y.Z-beta` 
+- mainnet: `vX.Y.Z`
```
JakubFornadel, line 11: when and on which branch do you want to create vX.Y.Z-beta tag ? In proposal we have only temporary `release/vX.Y.Z`  branch -> I guess creating tag from any commit on temporary branch is not good as these commits might be squashed, etc... After it is merged into master, we will create `vX.Y.Z` tag, which will be the same for mainnet as well as testnet ?

PR #2098, opened 2022-10-03
fix: fix potential spurious wakeup during sleep_(), which could lead …

JakubFornadel:
## Purpose

This PR mainly fixes a potential bug:

In sleep mechanism
`stop_cv_.wait_for(lock, std::chrono::milliseconds(time_to_sleep_for_ms));`
could spuriously wake up and in such case, wait_for exits, which would case this waiting mechanism to fail and we would start next step sooner than we are supposed to - not good.

Solution is to put this waiting mechanism into loop

As a byproduct this PR also cleans up the code around these timeouts:

- multiple unnecessary class members were removed
- u_long was replace by std::chrono::milliseconds

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -481,7 +481,7 @@ std::chrono::milliseconds PbftManager::elapsedTimeInMs(const time_point &start_t
 
 void PbftManager::sleep_() {
   // Run "wait_for" sleep in loop due to potential spurious wakeup on lock
-  while (true) {
+  while (!stopped_) {
     const auto round_elapsed_time = elapsedTimeInMs(current_round_start_datetime_);
     if (next_step_time_ms_ <= round_elapsed_time) {
       return;
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -25,6 +25,9 @@
 namespace taraxa {
 using vrf_output_t = vrf_wrapper::vrf_output_t;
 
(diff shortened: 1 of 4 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -466,24 +466,27 @@ void PbftManager::resetPbftConsensus(uint64_t round) {
...
+
+    const auto time_to_sleep_for_ms = next_step_time_ms_ - round_elapsed_time;
+    const auto [round, period] = getPbftRoundAndPeriod();
+    LOG(log_tr_) << "Sleep " << time_to_sleep_for_ms.count() << " [ms] before going into the next step. Period "
+                 << period << ", round " << round << ", step " << step_;
     std::unique_lock<std::mutex> lock(stop_mtx_);
-    stop_cv_.wait_for(lock, std::chrono::milliseconds(time_to_sleep_for_ms));
-  } else {
-    LOG(log_tr_) << "Skipping sleep, running late...";
+    stop_cv_.wait_for(lock, time_to_sleep_for_ms);
```
mfrankovi, line 489: I think you could use the return value of stop_cv_.wait_for to check if the time expired and return based on that

mfrankovi (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
JakubFornadel: return what ? This function is supposed to wait for time_to_sleep_for_ms. If wait_for exits before that due to spurious wake up, then what do you want to do ?

JakubFornadel:
Successful circleCI build here:
https://app.circleci.com/pipelines/github/Taraxa-project/taraxa-node/7783/workflows/a13aa4af-e5a1-4afa-ac65-47593d7f458d/jobs/8373

PR #2109, opened 2022-10-05
Fix estimate gas rpc

MatusKysel:
## Purpose
as `set_transaction_defaults` is setting wrong gas limit we where always returning only gas <= 90k, now it should return correct estimation up to GAS_LIMIT

https://github.com/Taraxa-project/taraxa-evm/pull/103


JakubFornadel (approved):

JakubFornadel:
@MatusKysel create PR also to release branch

PR #2111, opened 2022-10-05
No entering consensus steps if not eligible to vote and improve rebroadcast votes

mfrankovi:
If node is not eligible to vote, pbft state machine now never enters specific consensus steps.

Broadcasting/rebroadcasting soft votes/previous round next votes and reward votes is now no longer done in consensus steps. This broadcasting/rebroadcasting is a fail-safe method in case consensus gets delyed/stuck. In normal operation we move to next period and/or round before there is a need to do this.

Broadcasting votes is done each kBroadcastVotesLambdaTime(20) lambda time. This will only send votes to nodes which we did not send votes already.

Rebroadcasting votes is done each kReBroadcastVotesLambdaTime (60) lambda time. This will resend votes to all connected nodes


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -708,6 +740,15 @@ bool PbftManager::stateOperations_() {
     return true;
   }
 
+  // If node is not eligible to vote, always return true so pbft state machine never enters specific consensus steps
+  // (propose, soft-vote, cert-vote, next-vote). Nodes that have no delegation should just observe 2t+1 cert votes
+  // to move to the next period or 2t+1 next votes to move to the next round
+  if (!final_chain_->dpos_eligible_vote_count(period - 1, node_addr_)) {
+    // Check 2t+1 cert/next votes every 20ms
+    std::this_thread::sleep_for(std::chrono::milliseconds(20));
```
JakubFornadel, line 748: when I think about this, maybe we should use kPollingIntervalMs as this is used as timeout when we run the same step function in loop...

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -686,6 +690,24 @@ void PbftManager::loopBackFinishState_() {
   current_step_clock_initial_datetime_ = std::chrono::system_clock::now();
 }
 
+void PbftManager::broadcastVotes(bool rebroadcast) {
+  auto [round, period] = getPbftRoundAndPeriod();
+  auto net = network_.lock();
+  if (net) {
```
JakubFornadel, line 696: readability is better if you dont put big chunk of code into if when not needed. Instead of this if, just use:
assert(net) or

if (!net) {
  return
}

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -686,6 +690,24 @@ void PbftManager::loopBackFinishState_() {
...
+    if (auto soft_voted_block_data = getTwoTPlusOneSoftVotedBlockData(period, round);
+        soft_voted_block_data.has_value()) {
+      net->getSpecificHandler<network::tarcap::VotePacketHandler>()->onNewPbftVotes(
+          std::move(soft_voted_block_data->soft_votes_), rebroadcast);
+      vote_mgr_->sendRewardVotes(getLastPbftBlockHash(), rebroadcast);
+    }
+
+    LOG(log_dg_) << "Node " << node_addr_ << " broadcast next votes for previous round. In period " << period
```
JakubFornadel, line 704: I guess this part `"Node " << node_addr_ << ` is pretty much useless

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -686,6 +690,24 @@ void PbftManager::loopBackFinishState_() {
   current_step_clock_initial_datetime_ = std::chrono::system_clock::now();
 }
 
+void PbftManager::broadcastVotes(bool rebroadcast) {
+  auto [round, period] = getPbftRoundAndPeriod();
+  auto net = network_.lock();
+  if (net) {
+    if (auto soft_voted_block_data = getTwoTPlusOneSoftVotedBlockData(period, round);
```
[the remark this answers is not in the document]
mfrankovi: I think we need to call it because there might be votes coming in the meantime. This will not be called that often. Maybe once each 30 seconds so I think it should be fine

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: Replaced with polling interval

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: Changed

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: Removed that part

JakubFornadel (approved):

PR #2121, opened 2022-10-10
chore: implement nonce - gas price sorting

mfrankovi:
Since sorting transactions for block proposal is an expensive operation, it is now only done when we propose a dag block and not on ever transaction insertion/removal. Algorithm for sorting is:
1. Order by gas price all the unique sender transactions with minimal nonce per account
2. Remove already ordered transactions from the set and move back to step 1 if there are still transactions remaining

On execution preserve the order of all the transactions that have correct nonce order. Any transactions with nonce order is reversed are moved to the back of the order and sorted by nonce.

### DIFF
--- a/libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
@@ -4,32 +4,17 @@
 
 namespace taraxa {
 
-auto priorityComparator = [](const std::shared_ptr<Transaction> &first, const std::shared_ptr<Transaction> &second) {
-  if (first->getSender() == second->getSender()) {
-    return first->getNonce() < second->getNonce() ||
-           (first->getNonce() == second->getNonce() && first->getGasPrice() > second->getGasPrice());
-  } else {
-    return first->getGasPrice() > second->getGasPrice();
-  }
-};
+TransactionQueue::TransactionQueue(size_t max_size) : known_txs_(max_size * 2, max_size / 5), kMaxSize(max_size) {}
 
-TransactionQueue::TransactionQueue(size_t max_size)
-    : priority_queue_{priorityComparator}, known_txs_(max_size * 2, max_size / 5), kMaxSize(max_size) {
-  // There are library limits on multiset size, we need to check if max size is not exceeding it
-  if (kMaxSize > priority_queue_.max_size()) {
-    throw std::invalid_argument("Transaction pool size is too large");
-  }
-}
-
-size_t TransactionQueue::size() const { return hash_queue_.size(); }
+size_t TransactionQueue::size() const { return queue_transactions_.size(); }
 
 bool TransactionQueue::contains(const trx_hash_t &hash) const {
-  return hash_queue_.contains(hash) || non_proposable_transactions_.contains(hash);
+  return queue_transactions_.contains(hash) || non_proposable_transactions_.contains(hash);
 }
 
 std::shared_ptr<Transaction> TransactionQueue::get(const trx_hash_t &hash) const {
-  if (const auto it = hash_queue_.find(hash); it != hash_queue_.end()) {
-    return *(it->second);
+  if (const auto it = queue_transactions_.find(hash); it != queue_transactions_.end()) {
+    return it->second;
   }
 
   if (const auto transaction = non_proposable_transactions_.find(hash);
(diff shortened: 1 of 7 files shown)


justinsnapp:
What about the idea of moving a user's transactions where nonce are reversed to that user's last occurring transaction rather than to the back of the entire block?

mfrankovi:
> What about the idea of moving a user's transactions where nonce are reversed to that user's last occurring transaction rather than to the back of the entire block?

This is implemented. I am putting it there:
https://github.com/Taraxa-project/taraxa-node/blob/c314ed62e81b555ddad58a481bf39e4d57e1d4a9/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp#L1584

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -77,25 +104,45 @@ bool TransactionQueue::insert(std::shared_ptr<Transaction> &&transaction, const
...
+            non_proposable_transactions_[nonce_it->second->getHash()] = {last_block_number, nonce_it->second};
+            queue_transactions_.erase(nonce_it->second->getHash());
+            account_nonce_transactions_[transaction->getSender()][transaction->getNonce()] = transaction;
+            queue_transactions_[tx_hash] = transaction;
+          }
+        }
+      }
+
+      auto queue_size = size();
+      // This check if priority_queue_ is not bigger than max size if so we delete 1% of transactions
```
JakubFornadel, line 132: I forgot why do we delete 1% == 2000 txs ?

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -40,31 +25,73 @@ std::shared_ptr<Transaction> TransactionQueue::get(const trx_hash_t &hash) const
...
+    const auto &nonce_it = account_it->second.find(it->second->getNonce());
+    if (nonce_it != account_it->second.end()) {
+      if (hash == nonce_it->second->getHash()) {
+        account_it->second.erase(nonce_it);
+        if (account_it->second.size() == 0) {
+          account_nonce_transactions_.erase(account_it);
+        }
+      }
+    }
+  }
```
JakubFornadel, line 92: no need for all those nested ifs here, which are hard to read:

```
const auto it = queue_transactions_.find(hash);
if (it == queue_transactions_.end()) {
  return false;
}
queue_transactions_.erase(it);

const auto &account_it = account_nonce_transactions_.find(it->second->getSender());
if (account_it == account_nonce_transactions_.end()) {
  return
}

const auto &nonce_it = account_it->second.find(it->second->getNonce());
if (nonce_it == account_it->second.end()) {
  return
}
```

etc...

JakubFornadel (approved):

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
mfrankovi: ordering is expensive so this way it is done every 2000 new transactions received and not for every transaction received if we are at the limit. Ordering is only done when:
1. We are proposing dag block
2. We are over the max size limit

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
JakubFornadel: ok lets see how that behaves during stress tests

kstdl (approved):

JakubFornadel:
Successful circleCI build here
https://app.circleci.com/pipelines/github/Taraxa-project/taraxa-node/7771/workflows/2f67c74c-ebe2-4e2e-9c88-04a5948645d1/jobs/8359

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -4,32 +4,17 @@
...
-};
-
-TransactionQueue::TransactionQueue(size_t max_size)
-    : priority_queue_{priorityComparator}, known_txs_(max_size * 2, max_size / 5), kMaxSize(max_size) {
-  // There are library limits on multiset size, we need to check if max size is not exceeding it
-  if (kMaxSize > priority_queue_.max_size()) {
-    throw std::invalid_argument("Transaction pool size is too large");
-  }
-}
+TransactionQueue::TransactionQueue(size_t max_size) : known_txs_(max_size * 2, max_size / 5), kMaxSize(max_size) {}
```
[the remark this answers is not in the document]
mfrankovi: Any idea what size to reserve, does this really helps, it is not like it will be reserved permanently?

PR #2138, opened 2022-10-13
Max gas limit

### LINKED ISSUE #2132: Do not consume all gas in case tx fails inside dpos contract
## Task Description

This Task will...

MatusKysel:
## Purpose

https://github.com/Taraxa-project/taraxa-evm/pull/106/
fixed #2132

### DIFF
--- a/libraries/core_libs/network/rpc/eth/Eth.h
+++ b/libraries/core_libs/network/rpc/eth/Eth.h
@@ -9,6 +9,7 @@ namespace taraxa::net::rpc::eth {
 struct EthParams {
   Address address;
   uint64_t chain_id = 0;
+  uint64_t gas_limit = ((uint64_t)1 << 53) - 1;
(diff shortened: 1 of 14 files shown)


JakubFornadel (approved):

mfrankovi (approved):

libraries/core_libs/network/rpc/eth/Eth.h
```diff
@@ -9,6 +9,7 @@ namespace taraxa::net::rpc::eth {
 struct EthParams {
   Address address;
   uint64_t chain_id = 0;
+  uint64_t gas_limit = ((uint64_t)1 << 53) - 1;
```
kstdl, line 12: Should be overwritten with dag block limit. So maybe better idea to have 0 default value. It would be easier to notice if we are not setting it somewhere

kstdl (approved):

PR #2139, opened 2022-10-14
chore: trx sorting minor cleanup

mfrankovi:
Minor changes related to transactions ordering

### DIFF
--- a/libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
@@ -4,7 +4,9 @@
 
 namespace taraxa {
 
-TransactionQueue::TransactionQueue(size_t max_size) : known_txs_(max_size * 2, max_size / 5), kMaxSize(max_size) {}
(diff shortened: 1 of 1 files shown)


JakubFornadel (approved):

libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -77,13 +83,9 @@ bool TransactionQueue::erase(const trx_hash_t &hash) {
...
-    return false;
-  }
+  assert(account_it != account_nonce_transactions_.end());
   const auto &nonce_it = account_it->second.find(it->second->getNonce());
-  if (nonce_it == account_it->second.end()) {
-    return false;
-  }
+  assert(nonce_it != account_it->second.end());
 
   if (hash == nonce_it->second->getHash()) {
```
JakubFornadel, line 90: btw should this be assert too ? Is it possible that these 2 values are different ?

mfrankovi: added the assert

PR #2140, opened 2022-10-14
fix: ordered transaction get when queue empty

mfrankovi:
When queue empty getOrderedTransactions would return an invalid transactions instead of 0 transactions. Fixed


libraries/core_libs/consensus/src/transaction/transaction_queue.cpp
```diff
@@ -48,6 +48,9 @@ SharedTransactions TransactionQueue::getOrderedTransactions(uint64_t count) cons
     it->second.first++;
   }
   while (true) {
```
JakubFornadel, line 50: why not put it here ? 
`while(!head_transactions.empty())`

mfrankovi: Done

kstdl (approved):

PR #2144, opened 2022-10-14
fix: prnet deploy current chart

rjonczy:
## Purpose

Use current helm chart of taraxa-node

Replaces repo of chart.

## Changes made

Uses chart from the same repo what main taraxa-node


rattrap (approved):
LGTM

PR #2153, opened 2022-10-18
fix: pbft syncing race condition

mfrankovi:
If while syncing our chain advances by consensus pushing the block to the chain, stop requesting new blocks if we are synced.

Previously in this case we might request a future block which peer that we are syncing from would consider a malicious request

### DIFF
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
@@ -74,14 +74,15 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
       peer->dag_level_ = block.getLevel();
     }
   }
+  const auto pbft_block_period = period_data.pbft_blk->getPeriod();
 
-  LOG(log_dg_) << "PbftSyncPacket received. Period: " << period_data.pbft_blk->getPeriod()
+  LOG(log_dg_) << "PbftSyncPacket received. Period: " << pbft_block_period
                << ", dag Blocks: " << received_dag_blocks_str << " from " << packet_data.from_node_id_;
 
   peer->markPbftBlockAsKnown(pbft_blk_hash);
   // Update peer's pbft period if outdated
-  if (peer->pbft_chain_size_ < period_data.pbft_blk->getPeriod()) {
-    peer->pbft_chain_size_ = period_data.pbft_blk->getPeriod();
+  if (peer->pbft_chain_size_ < pbft_block_period) {
+    peer->pbft_chain_size_ = pbft_block_period;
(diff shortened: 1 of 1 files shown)


libraries/core_libs/network/src/tarcap/packets_handlers/pbft_sync_packet_handler.cpp
```diff
@@ -194,6 +195,11 @@ void PbftSyncPacketHandler::process(const PacketData &packet_data, const std::sh
   }
 
   if (last_block) {
+    // If current sync period is actually bigger than the block we just received we are probably synced but verify with
+    // calling restartSyncingPbft
+    if (pbft_sync_period > pbft_block_period) {
+      return restartSyncingPbft(true);
+    }
```
JakubFornadel, line 202: I think this is not ideal fix - give anyone a week and I am pretty sure nobody will know that this if is supposed to be protecting calling `syncPeerPbft` with wrong argument. Proper fix must ensure that syncPeerPbft will never proceed in sending sync request if it not supposed to, not just on this particular place.

Solution:
1. Change syncingPeer to:
const std::shared_ptr<TaraxaPeer> PbftSyncingState::syncingPeer() const {
  std::shared_lock lock(peer_mutex_);
  return peer_;

    // This was here originally - find it super weird to generate some random nodeID in case there is no peer set
   // we should definitely return nullptr in case there is no syncing peer set
  // if (peer_) {
  //  return peer_->getId();
  // }
  // return dev::p2p::NodeID();
}

3. Change syncPeerPbft to:
```
bool ExtSyncingPacketHandler::syncPeerPbft(uint64_t height_to_sync) {
  const auto sync_peer = pbft_syncing_state_->syncingPeer();
  if (!sync_peer) {
     // log err...
     return false;
  }

  if (height_to_sync > sync_peer->pbft_chain_size_) {
     // log err...
     return false;
  }

  LOG(log_nf_) << "Sync peer node " << node_id << " from pbft chain height " << height_to_sync;
  return sealAndSend(sync_peer->->getId(), SubprotocolPacketType::GetPbftSyncPacket, std::move(dev::RLPStream(1) << height_to_sync));
}
```

mfrankovi: the problem with your approach is that sync_peer->pbft_chain_size_ might not be up to date and it might be smaller than the actual peer chain size. This is specially true in unit tests where we have very small lambda time. But it could always happen. The pbft syncing packet will send a pbft_chain_synced flag in the packet when it is sent its last block. So this flag is a good indicator if we are fully synced. The reason why we had this issue is only if we progress with consensus/votes and not by syncing at the same time. So I think with your fix we might conclude we are synced even though we might be one or two blocks behind.

PR #2155, opened 2022-10-19
fix: never send syncing request in case therequested period > syncing…

JakubFornadel:
…_peer chain size


libraries/core_libs/network/src/tarcap/stats/node_stats.cpp
```diff
@@ -259,8 +261,8 @@ Json::Value NodeStats::getStatus() const {
     }
   }
 
-  if (pbft_syncing_state_->isPbftSyncing()) {
-    res["syncing_from_node_id"] = pbft_syncing_state_->syncingPeer().toString();
+  if (pbft_syncing_state_->isPbftSyncing() && pbft_syncing_state_->syncingPeer()) {
```
kstdl, line 264: `pbft_syncing_state_->syncingPeer()` used twice, so maybe better to move it to variable above

kstdl (approved):

libraries/core_libs/network/src/tarcap/stats/node_stats.cpp
JakubFornadel: fixed

mfrankovi (approved):

PR #2157, opened 2022-10-19
fix: eth_call default gas and gas_price values

kstdl:
#2156


justinsnapp (approved):
Looks great to me

libraries/core_libs/network/rpc/eth/Eth.cpp
```diff
@@ -276,18 +273,19 @@ class EthImpl : public Eth, EthParams {
...
     if (!t.nonce) {
       t.nonce = transaction_count(blk_n, t.from);
     }
     if (!t.gas_price) {
-      t.gas_price = gas_pricer();
+      t.gas_price = 0;
     }
     if (!t.gas) {
-      t.gas = 90000;
+      t.gas = gas_limit;
```
JakubFornadel, line 288: lets use dag block gas limit as we talked - it is pretty easy and quick fix to pass this config param to Eth class

JakubFornadel (approved):

PR #2159, opened 2022-10-20
fix: do not soft vote missing block

JakubFornadel:
## Purpose

If (for any reason/bug) we lose block, there is an infinite loop in our pbft when:
1. nodes soft vote missing previous round next voted block
2. nodes next vote 2t+1 soft-voted block (from previous step)
3. in new round propose step is skipped as nodes are supposed to re-propose previous round next voted block but they cant as they dont have it
4. repeat the loop from step 1.

This is a state, from which we cant recover + we are saving proposed block in memory as well as db per period & round. So even if the block is actually saved on some node(e.g. period 1, round 2), we already moved forward multiple rounds so we are not able to get it due to that infinite loop I mentioned above - we might be trying to get some block hash for current period 1, round 10 but the block is saved for period 1, round 2

Vote only for block that node has and it is valid

Reason:
- Voting for missing blocks means we have no idea if the block is valid, etc... If we do not allow this for newly proposed blocks it does not make sense to allow it for previous round next voted blocks

This PR does not change current pbft behaviour - it would not start proposing new blocks, but rather get stalled on soft/next voting steps with error message - "Unable to soft/next vte for missing XYZ block", which would be printed repeatedly. It should be easier to spot and later debug without increasing round

### DIFF
--- a/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
+++ b/libraries/core_libs/consensus/include/pbft/pbft_manager.hpp
@@ -534,14 +534,15 @@ class PbftManager : public std::enable_shared_from_this<PbftManager> {
   const std::optional<TwoTPlusOneSoftVotedBlockData> &getTwoTPlusOneSoftVotedBlockData(uint64_t period, uint64_t round);
 
   /**
-   * @brief Get valid proposed pbft block. It will retrieved block from proposed_blocks and then validate it
+   * @brief Get valid proposed pbft block. It will retrieve block from proposed_blocks and then validate it if not
+   *        already validated
    *
    * @param period
    * @param round
    * @param block_hash
    * @return valid proposed pbft block or nullptr
    */
-  std::shared_ptr<PbftBlock> getValidPbftProposedBlock(uint64_t period, uint64_t round, const blk_hash_t& block_hash) const;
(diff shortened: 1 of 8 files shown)


mfrankovi (commented):
I fear that this could lead to a fork, if we allow proposing a new pbft block in this case it is possible that some nodes actually already observed enough cert votes for previous soft voted block and we might end up with a fork

justinsnapp:
I think we should fix the underlying bug as @mfrankovi  mentions and NOT try and add a protocol behavior to give up soft voted block in attempt to hide potential other bugs.  Especially at this juncture.

JakubFornadel:
> I think we should fix the underlying bug as @mfrankovi mentions and NOT try and add a protocol behavior to give up soft voted block in attempt to hide potential other bugs. Especially at this juncture.

I updated description, which explains why it would not change current pbft behaviour nor it would hife bugs - it would rather make them more obvious and easier to debug

justinsnapp (approved):
I woefully misinterpreted the description and failed to comprehend the last line in the PR description.

When it was stated that it "avoids the infinite loop" I thought it meant we would exit the loop when in point of fact we HANG and don't exit the round to start a new cycle of the loop when we have a data availability problem with a soft voted block.

So that seems good to me, the opposite of "masking" a bug, its really flagging it.   Since data availability after 2t+1 people observed it before is either a bug or a VERY powerful attack to knock entire committee from a given round offline.

mfrankovi (approved):

PR #2165, opened 2022-10-25
chore: verify vrf and node keys

mfrankovi:
On proposing dag blocks, verify the vrf secret key that was retrieved from wallet file on startup matches the public vrf key used when registering a validator. Do not propose a block if they are not matching.

Verify public keys and addresses from wallet file to match the secret keys in the wallet file.

### DIFF
--- a/libraries/config/src/config.cpp
+++ b/libraries/config/src/config.cpp
@@ -275,6 +275,19 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value 
 
   try {
     node_secret = dev::Secret(wallet["node_secret"].asString(), dev::Secret::ConstructFromStringType::FromHex);
+    if (!wallet["node_public"].isNull()) {
+      auto node_public = dev::Public(wallet["node_public"].asString(), dev::Public::ConstructFromStringType::FromHex);
+      if (node_public != dev::KeyPair(node_secret).pub()) {
+        throw ConfigException(std::string("Node secret key and public key in wallet do not match"));
+      }
+    }
+    if (!wallet["node_address"].isNull()) {
+      auto node_address =
+          dev::Address(wallet["node_address"].asString(), dev::Address::ConstructFromStringType::FromHex);
+      if (node_address != dev::KeyPair(node_secret).address()) {
+        throw ConfigException(std::string("Node secret key and address in wallet do not match"));
+      }
+    }
   } catch (const dev::Exception &e) {
     throw ConfigException(std::string("Could not parse node_secret: ") + e.what());
   }
@@ -283,6 +296,17 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value 
     vrf_secret = vrf_wrapper::vrf_sk_t(wallet["vrf_secret"].asString());
   } catch (const dev::Exception &e) {
     throw ConfigException(std::string("Could not parse vrf_secret: ") + e.what());
+  }
+
+  try {
+    if (!wallet["vrf_public"].isNull()) {
+      auto vrf_public = vrf_wrapper::vrf_pk_t(wallet["vrf_public"].asString());
+      if (vrf_public != taraxa::vrf_wrapper::getVrfPublicKey(vrf_secret)) {
+        throw ConfigException(std::string("Vrf secret key and public key in wallet do not match"));
+      }
+    }
+  } catch (const dev::Exception &e) {
+    throw ConfigException(std::string("Could not parse vrf_public: ") + e.what());
   }
   // TODO configurable
   opts_final_chain.expected_max_trx_per_block = 1000;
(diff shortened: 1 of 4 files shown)


libraries/core_libs/consensus/src/dag/block_proposer.cpp
```diff
@@ -107,6 +108,12 @@ bool SortitionPropose::propose() {
     return false;
   }
   LOG(log_nf_) << "VDF computation time " << vdf.getComputationTime() << " difficulty " << vdf.getDifficulty();
+  auto pk = key_manager_->get(*proposal_period, node_addr_);
+  if (pk && *pk != vrf_pk_) {
```
JakubFornadel, line 112: why it is here and not at the beginning of the function ? Some processing can be saved

libraries/core_libs/consensus/include/dag/block_proposer.hpp
```diff
@@ -21,8 +21,10 @@ class TransactionManager;
 class FullNode;
 class BlockProposer;
 class DagBlock;
+class KeyManager;
 
 using vrf_sk_t = vrf_wrapper::vrf_sk_t;
+using vrf_pk_t = vrf_wrapper::vrf_pk_t;
```
JakubFornadel, line 27: having such using in header is bad practice due to namespaces pollution... having it in cpp file is ok but in this case when implementation is in header, full types should be used

libraries/core_libs/consensus/src/dag/block_proposer.cpp
JakubFornadel: inst it even better when the message is printed all the time ? It happens when user somehow adjust his wallet in wrong way or he registers his validator in wrong way, in which case it is good to get his attention as soon as possible so he can change it

libraries/config/src/config.cpp
```diff
@@ -275,6 +275,19 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value
 
   try {
     node_secret = dev::Secret(wallet["node_secret"].asString(), dev::Secret::ConstructFromStringType::FromHex);
+    if (!wallet["node_public"].isNull()) {
```
kstdl, line 278: Why do we even have this fields in wallet file if we really need only `secret`?

kstdl (approved):

libraries/config/src/config.cpp
mfrankovi: I am not sure. It is not mandatory. It does not need to be there. But in a way I think it is useful for user to be able to get its address/public key this way. If we want community user to give us its public key/address, what would be better alternative?

libraries/core_libs/consensus/src/dag/block_proposer.cpp
mfrankovi: ok, moved it up

JakubFornadel (approved):

libraries/config/src/config.cpp
```diff
@@ -275,6 +275,19 @@ FullNodeConfig::FullNodeConfig(Json::Value const &string_or_object, Json::Value
...
+        throw ConfigException(std::string("Node secret key and public key in wallet do not match"));
+      }
+    }
+    if (!wallet["node_address"].isNull()) {
+      auto node_address =
+          dev::Address(wallet["node_address"].asString(), dev::Address::ConstructFromStringType::FromHex);
+      if (node_address != dev::KeyPair(node_secret).address()) {
+        throw ConfigException(std::string("Node secret key and address in wallet do not match"));
+      }
+    }
```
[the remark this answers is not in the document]
mfrankovi: Moving this to validate would mean that node_public and node_address would actually have to be a part of the config. I am not sure if that makes sense if they would not be used anywhere after validate?

libraries/core_libs/consensus/include/dag/block_proposer.hpp
```diff
@@ -100,7 +106,7 @@ class BlockProposer : public std::enable_shared_from_this<BlockProposer> {
         node_sk_(node_sk),
         vrf_sk_(vrf_sk) {
     LOG_OBJECTS_CREATE("PR_MDL");
-    propose_model_ = std::make_unique<SortitionPropose>(node_addr, db_, dag_mgr_, trx_mgr_);
+    propose_model_ = std::make_unique<SortitionPropose>(node_addr, db_, dag_mgr_, trx_mgr_, key_manager);
```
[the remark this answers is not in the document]
mfrankovi: Done

libraries/core_libs/consensus/include/dag/block_proposer.hpp
```diff
@@ -90,7 +95,8 @@ class BlockProposer : public std::enable_shared_from_this<BlockProposer> {
  public:
   BlockProposer(BlockProposerConfig const& bp_config, std::shared_ptr<DagManager> dag_mgr,
                 std::shared_ptr<TransactionManager> trx_mgr, std::shared_ptr<FinalChain> final_chain,
-                std::shared_ptr<DbStorage> db, addr_t node_addr, secret_t node_sk, vrf_sk_t vrf_sk)
+                std::shared_ptr<DbStorage> db, addr_t node_addr, secret_t node_sk, vrf_wrapper::vrf_sk_t vrf_sk,
+                std::shared_ptr<KeyManager> key_manager)
```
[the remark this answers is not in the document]
mfrankovi: Done

PR #2184, opened 2022-11-07
chore: handle transactions queue overflow

mfrankovi:
Once a peer has been dag synced, it should always send transactions and dag blocks in order so that dag block will never miss a transaction and dag block should never have a missing tip or pivot.

One exception to this rule is when transaction pool is full and node is forced to drop transactions. If transactions are dropped a DAG block might arrive with transactions missing. In this case there is a need to sync DAG again. At the same time allowing nodes to dag sync multiple times which is an expensive operations leaves us vulnerable to attacks. A limit is introduced that a peer can only ask for dag syncing once every kDagSyncingLimit(300) seconds.

Once transactions are dropped when transactions queue is full this is marked with transaction_overflow_time_ where node saves last time transactions were dropped. If in the next kTransactionOverflowTimeLimit (300 seconds) node receives a dag block with a missing transaction it is safe to assume that dag syncing needs to be done to receive this transaction.

This change aims to allow dag resyncing when transaction queue is full while at the same time limiting how often dag sync can be requested.

### DIFF
--- a/libraries/core_libs/consensus/include/transaction/transaction_queue.hpp
+++ b/libraries/core_libs/consensus/include/transaction/transaction_queue.hpp
@@ -109,6 +109,15 @@ class TransactionQueue {
    */
   bool isTransactionKnown(const trx_hash_t& trx_hash) const;
 
+  /**
+   * @brief Have transactions been recently dropped due to queue reaching max size
+   * This call is thread-safe
+   * @return Returns true if txs were dropped
+   */
+  bool transactionsDropped() const {
+    return std::chrono::system_clock::now() - transaction_overflow_time_ < kTransactionOverflowTimeLimit;
+  }
+
  private:
   // Transactions in the queue per account ordered by nonce
   std::unordered_map<addr_t, std::map<val_t, std::shared_ptr<Transaction>>> account_nonce_transactions_;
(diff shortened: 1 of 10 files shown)


libraries/core_libs/consensus/include/transaction/transaction_queue.hpp
```diff
@@ -122,6 +133,13 @@ class TransactionQueue {
 
   ExpirationCache<trx_hash_t> known_txs_;
 
+  // Last time transactions were dropped due to queue reaching max size
+  uint64_t transaction_overflow_time_ = 0;
```
JakubFornadel, line 137: use std::chrono::system_clock::time_point so it is clear also by looking at type that it is time point. Then also this function get much cleaner:

```
bool transactionsDropped() const {
    return (std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch())
                .count() -
            transaction_overflow_time_) < kTransactionOverflowTimeLimit;
  }
```

as you can simply use now() - transaction_overflow_time_

libraries/core_libs/consensus/include/transaction/transaction_queue.hpp
```diff
@@ -122,6 +133,13 @@ class TransactionQueue {
...
+
+  // If transactions are dropped within last kTransactionOverflowTimeLimit seconds, dag blocks with missing transactions
+  // will not be treated as malicious
+  const uint64_t kTransactionOverflowTimeLimit = 300;
```
JakubFornadel, line 141: use const std::chrono::milliseconds kTransactionOverflowTimeLimit{300};

mfrankovi: Done

libraries/core_libs/network/include/network/tarcap/taraxa_peer.hpp
```diff
@@ -91,8 +115,10 @@ class TaraxaPeer : public boost::noncopyable {
   std::atomic<size_t> pbft_previous_round_next_votes_size_ = 0;
   std::atomic<PbftPeriod> last_status_pbft_chain_size_ = 0;
   std::atomic_bool peer_dag_synced_ = false;
+  std::atomic_uint64_t peer_dag_synced_time_ = 0;
```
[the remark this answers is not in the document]
mfrankovi: the only thing is that here I use atomic types, if I am to use time_point I would have to add methods and protect with additional mutex, not sure if it is worth it than?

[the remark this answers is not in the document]
JakubFornadel: yeah probably not

JakubFornadel (approved):

PR #2187, opened 2022-11-09
chore: optimize block proposal and pbft leader selection

mfrankovi:
On block proposal only give up proposal if new dag block of same level was added to the DAG. Previous implementation would give up if an old DAG block was added to the DAG as well. This would unnecessary delay block proposal for both stale and non-stale blocks.

Pbft block leader selection now favors non-empty pbft blocks over empty ones.

### DIFF
--- a/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
+++ b/libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
@@ -1500,6 +1500,7 @@ std::shared_ptr<PbftBlock> PbftManager::identifyLeaderBlock_(PbftRound round, Pb
     leader_candidates[getProposal(v)] = v;
   }
 
+  std::shared_ptr<PbftBlock> empty_leader_block;
   for (auto const &leader_vote : leader_candidates) {
     const auto proposed_block_hash = leader_vote.second->getBlockHash();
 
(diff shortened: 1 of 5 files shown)


libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1430,23 +1431,45 @@ std::shared_ptr<PbftBlock> PbftManager::identifyLeaderBlock_(uint64_t round, uin
...
+    if (block->getPivotDagBlockHash() == NULL_BLOCK_HASH) {
+      leader_candidates_empty_blocks[getProposal(v)] = block;
+    } else {
+      leader_candidates[getProposal(v)] = block;
+    }
   }
 
-  for (auto const &leader_vote : leader_candidates) {
-    const auto proposed_block_hash = leader_vote.second->getBlockHash();
+  for (auto const &leader : leader_candidates) {
```
JakubFornadel, line 1451: if we wanted to be fancy and not repeat code, this could be in lambda function:

```
  // Empty blocks are only selected as leader if there is no block proposed with dag blocks
  for (auto const &leader : leader_candidates_empty_blocks) {
    auto leader_block = leader.second;

    if (!compareBlocksAndRewardVotes_(leader_block)) {
      LOG(log_er_) << "Incomplete or invalid proposed block " << leader_block->getBlockHash() << ", period " << period
                   << ", round " << round;
      continue;
    }
    if (!pbft_chain_->checkPbftBlockValidation(leader_block)) {
      LOG(log_er_) << "Proposed block " << leader_block->getBlockHash() << " failed validation, period " << period
                   << ", round " << round;
      continue;
    }
    return leader_block;
  }

```


and we just call it with leader_candidates and leader_candidates_empty_blocks

kstdl (approved):

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: Done

JakubFornadel (approved):

PR #2190, opened 2022-11-09
bump explorer version

rjonczy:
## Purpose

Dependency of mongodb with broken probes 

Bump version of explorer. Which is using latest mongodb chart.

## Changes made

Changed version of explorer helm chart


rattrap (approved):
LGTM

VargaElod23 (approved):
LGTM

PR #2193, opened 2022-11-10
Fix: do not start build-docker-image when using prnet commands

rjonczy:
## Purpose

`build-docker-image` was triggered when using command for prnets:
- \deploy
- \cleanup
- \redeploy

Added conditions to job.

## Changes made

Added conditions to job.


JakubFornadel (approved):

rattrap (approved):
Perfect. This should save us some $$$
LGTM

rattrap (commented):
LGTM

PR #2210, opened 2022-11-16
[helm-chart] - v0.2.4: allow to spec persistance per node type

rjonczy:
## Purpose

We want to allow to spec persistance (PVC) per node type.

Adds persistance settings (PVC) for each node type

## Changes made

Added persistance to manifests

### DIFF
--- a/charts/taraxa-node/templates/boot-node.yaml
+++ b/charts/taraxa-node/templates/boot-node.yaml
@@ -50,7 +50,7 @@ spec:
               mountPath: /bin/entrypoint.sh
               readOnly: true
               subPath: entrypoint.sh
-            - name: {{ .Values.bootnode.persistence.name }}
+            - name: data
               mountPath: /root/.taraxa
       containers:
         - name: boot-node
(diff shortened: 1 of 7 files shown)


charts/taraxa-node/templates/boot-node.yaml
```diff
@@ -118,26 +118,30 @@ spec:
...
-        - {{ .Values.persistence.accessMode | quote }}
-    {{- if .Values.persistence.storageClass }}
-    {{- if (eq "-" .Values.persistence.storageClass) }}
+        - {{ .Values.bootnode.persistence.accessMode | quote }}
+    {{- if .Values.bootnode.persistence.storageClass }}
+    {{- if (eq "-" .Values.bootnode.persistence.storageClass) }}
       storageClassName: ""
     {{- else }}
-      storageClassName: "{{ .Values.persistence.storageClass }}"
+      storageClassName: "{{ .Values.bootnodepersistence.storageClass }}"
```
rattrap, line 140: Typo here - `.Values.bootnodepersistence`

rattrap (changes requested):

rattrap (approved):

PR #2214, opened 2022-11-18
chore: low nonce and insufficient balance trxs handling improvement

mfrankovi:
Any transaction that is part of the dag syncing packet and part of the dag block within it will be forced pushed to queue since it is about to be included in the dag block. It will be pushed into nonproposable transactions so that we do not propose it in a block until the dag block is processed.

Transaction queue will have a limit on non proposable transactions up to 20% of its total size.

Suspicious packets exception will be raised only if both peer suspicious packets are over the limit and non proposable transactions in the queue are over the limit.


libraries/core_libs/network/src/tarcap/packets_handlers/dag_sync_packet_handler.cpp
```diff
@@ -129,7 +143,7 @@ void DagSyncPacketHandler::process(const PacketData& packet_data, const std::sha
       std::chrono::duration_cast<std::chrono::seconds>(std::chrono::system_clock::now().time_since_epoch()).count();
   peer->peer_dag_syncing_ = false;
 
-  LOG(log_dg_) << "Received DagSyncPacket with blocks: " << received_dag_blocks_str
+  LOG(log_er_) << "Received DagSyncPacket with blocks: " << dag_blocks_to_log
```
JakubFornadel, line 146: why error ?

mfrankovi: Reverted to debug. It was changed to error for some debugging that I did and I forget to change it back.

JakubFornadel (approved):

libraries/core_libs/network/src/tarcap/packets_handlers/transaction_packet_handler.cpp
```diff
@@ -76,12 +76,14 @@ inline void TransactionPacketHandler::process(const PacketData &packet_data, con
         }
         case TransactionStatus::InsufficentBalance:
         case TransactionStatus::LowNonce: {
-          if (peer->reportSuspiciousPacket()) {
+          // Raise exception in trx pool is over the limit and this peer already has too many suspicious packets
+          if (peer->reportSuspiciousPacket() && trx_mgr_->nonProposableTransactionsOverTheLimit()) {
```
kstdl, line 80: So we won't disconnect from peer that is sending a lot of suspicious packets(over peer limit) until we will reach this global limit. Why is this better then logic we had before?

kstdl (approved):

Issue #1781, opened 2022-06-03
Cleanup predefined/test configs


JakubFornadel:
Configs for tests & predefined configs is just 1 huge mess:
- configs.json in test folder dontcontaint most of the params
- something is set in predefined()
- something is set in make_node_cfgs
- using lambda function to create predefined config is for what ? It does not make sense...

PR #2221, opened 2022-11-29
chore: simplify testing configuration. restructurizing test utils

kstdl:
#1781


libraries/config/src/config.cpp
```diff
@@ -80,20 +73,37 @@ FullNodeConfig::FullNodeConfig(const Json::Value &string_or_object, const Json::
...
-
   // blocks_per_year config param is calculated from lambda_ms
   uint64_t year_ms = 365 * 24 * 60 * 60;
   year_ms *= 1000;
   // we have fixed 2*lambda time for proposing step and adding some expecting value for filter and certify steps
   const uint32_t expected_block_time = 2 * genesis.pbft.lambda_ms + 700;
   genesis.state.dpos.blocks_per_year = year_ms / expected_block_time;
 
-  is_light_node = getConfigDataAsBoolean(root, {"is_light_node"}, true, false);
+  is_light_node = getConfigDataAsBoolean(root, {"is_light_node"}, true, is_light_node);
```
JakubFornadel, line 106: hmm last argument == default value, why is it changed from false to is_light_node ?

libraries/config/src/genesis.cpp
```diff
@@ -63,73 +63,57 @@ void dec_json(Json::Value const& json, Genesis& obj) {
   dec_json(json["dag"], obj.dag);
 }
 
-const Genesis& Genesis::predefined(std::string const& name) {
-  if (auto i = predefined_->find(name); i != predefined_->end()) {
-    return i->second;
-  }
-  throw std::runtime_error("unknown chain config: " + name);
+// const Genesis& Genesis::predefined(std::string const& name) {
```
JakubFornadel, line 66: do we want to keep this commented code for some reason ?

JakubFornadel (approved):

libraries/config/src/config.cpp
kstdl: It basically changing nothing, just seems to be more logical for me to keep previous value if don't have value to set

libraries/config/src/genesis.cpp
kstdl: No, I will remove it. Thought that I removed it already, but I guess it was somehow messed during rebase

PR #2245, opened 2022-12-15
Small improvements

MatusKysel:

### DIFF
--- a/libraries/cli/include/cli/config_jsons/mainnet/mainnet_config.json
+++ b/libraries/cli/include/cli/config_jsons/mainnet/mainnet_config.json
@@ -86,14 +86,6 @@
           {
             "name": "SUMMARY",
             "verbosity": "INFO"
-          },
-          {
-            "name": "PBFT_MGR",
-            "verbosity": "INFO"
-          },
-          {
-            "name": "DAG_PROPOSER",
-            "verbosity": "INFO"
           }
         ],
         "outputs": [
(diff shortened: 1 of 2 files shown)


kstdl (approved):

libraries/cli/include/cli/config_jsons/testnet/testnet_config.json
```diff
@@ -71,14 +71,6 @@
...
             "name": "SUMMARY",
             "verbosity": "INFO"
-          },
-          {
-            "name": "PBFT_MGR",
-            "verbosity": "INFO"
-          },
-          {
-            "name": "DAG_PROPOSER",
-            "verbosity": "INFO"
```
JakubFornadel, line 81: why did you remove those 2 channels ?

JakubFornadel (changes requested):
.

libraries/cli/include/cli/config_jsons/testnet/testnet_config.json
JakubFornadel: If there are some spamming messages, create issue with those so we can fix it, because it is supposed to log only info when user creates vote, dag block or pbft block which are all rewarded, not that frequent and definitely should not be spamming. This is not right solution

JakubFornadel: so what are useful logs for community users in your opinion then ? If not such most basic thing as logging when he created pbft or dag block then I dont know what ? If user monitors his logs, he probably wants to see at least some basic logs about his activity + errors

JakubFornadel: yeah I think your approach is wrong too :) If it was some problem that takes long to solve then yes, this is probably 1 or 2 logs where we simply set verbosity to debug instead of info but you rather turn it off completely ?

JakubFornadel: I will do it no problem just give me the messages as you already investigated it ?

JakubFornadel: > and if you want new channels enable for community user please join discord and explain then why you enabled it

do you know that you did approve it ? https://github.com/Taraxa-project/taraxa-node/pull/2196. Did you read the description ? So you enabled it too in a fact :)

JakubFornadel (approved):

Issue #2247, opened 2022-12-19
Prometheus integration


kstdl:
Add prometheus library https://github.com/jupp0r/prometheus-cpp

PR #2248, opened 2022-12-19
feat: add metrics service. add some basic metrics updaters

kstdl:
#2247

### DIFF
--- a/libraries/cli/include/cli/config_jsons/default/default_config.json
+++ b/libraries/cli/include/cli/config_jsons/default/default_config.json
@@ -11,6 +11,10 @@
       "http_port": 9777,
       "ws_port": 6777
     },
+    "prometheus": {
+      "listen_port": 8888,
+      "polling_interval_ms": 1000
+    },
     "listen_ip": "0.0.0.0",
     "listen_port": 10002,
     "transaction_interval_ms": 100,
--- a/libraries/config/include/config/network.hpp
+++ b/libraries/config/include/config/network.hpp
@@ -8,6 +8,12 @@
 
 namespace taraxa {
 
+struct PrometheusConfig {
+  std::string address;
+  uint16_t listen_port = 0;
+  uint16_t polling_interval_ms = 1000;
+};
+
 struct ConnectionConfig {
   std::optional<uint16_t> http_port;
   std::optional<uint16_t> ws_port;
@@ -51,6 +57,7 @@ struct NetworkConfig {
 
   std::optional<ConnectionConfig> rpc;
   std::optional<ConnectionConfig> graphql;
+  std::optional<PrometheusConfig> prometheus;
 
   void validate() const;
 };
--- a/libraries/config/src/network.cpp
+++ b/libraries/config/src/network.cpp
@@ -5,13 +5,8 @@
 namespace taraxa {
 
 void dec_json(const Json::Value &json, PrometheusConfig &config) {
-  if (auto listen_port = getConfigData(json, {"listen_port"}, true); !listen_port.isNull()) {
-    config.listen_port = listen_port.asUInt();
-  }
-
-  if (auto polling_interval_ms = getConfigData(json, {"polling_interval_ms"}, true); !polling_interval_ms.isNull()) {
-    config.polling_interval_ms = polling_interval_ms.asUInt();
-  }
+  config.listen_port = getConfigData(json, {"listen_port"}).asUInt();
+  config.polling_interval_ms = getConfigData(json, {"polling_interval_ms"}).asUInt();
 }
 
 void ConnectionConfig::validate() const {
--- a/libraries/config/src/network.cpp
+++ b/libraries/config/src/network.cpp
@@ -4,6 +4,16 @@
 
 namespace taraxa {
 
+void dec_json(const Json::Value &json, PrometheusConfig &config) {
+  if (auto listen_port = getConfigData(json, {"listen_port"}, true); !listen_port.isNull()) {
+    config.listen_port = listen_port.asUInt();
+  }
+
+  if (auto polling_interval_ms = getConfigData(json, {"polling_interval_ms"}, true); !polling_interval_ms.isNull()) {
+    config.polling_interval_ms = polling_interval_ms.asUInt();
+  }
+}
+
 void ConnectionConfig::validate() const {
   if (!http_port && !ws_port) {
     throw ConfigException("Either http_port or ws_port post must be specified for connection config");
@@ -115,6 +125,14 @@ void dec_json(const Json::Value &json, NetworkConfig &network) {
     network.graphql->address = listen_ip;
 
     dec_json(graphql_json, *network.graphql);
+  }
+
+  if (auto prometheus_json = getConfigData(json, {"prometheus"}, true); !prometheus_json.isNull()) {
+    network.prometheus.emplace();
+    // ip address
+    network.prometheus->address = network.listen_ip;
+
+    dec_json(prometheus_json, *network.prometheus);
   }
 }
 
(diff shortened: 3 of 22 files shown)


libraries/config/include/config/network.hpp
```diff
@@ -8,6 +8,12 @@
 
 namespace taraxa {
 
+struct PrometheusConfig {
+  std::string address;
+  uint16_t listen_port = 0;
+  uint16_t polling_interval_ms = 1000;
```
JakubFornadel, line 14: btw shouldnt we use const for all config params ?

libraries/config/src/network.cpp
```diff
@@ -4,6 +4,16 @@
 
 namespace taraxa {
 
+void dec_json(const Json::Value &json, PrometheusConfig &config) {
+  if (auto listen_port = getConfigData(json, {"listen_port"}, true); !listen_port.isNull()) {
```
JakubFornadel, line 8: if whole section "prometheus" is optional, then the values in it dont should not be optional I guess - if you specify "primetheus", you must specify also things in it

libraries/core_libs/node/include/node/node.hpp
```diff
@@ -28,6 +28,9 @@
 #include "transaction/transaction.hpp"
 #include "vote_manager/vote_manager.hpp"
 
+namespace metrics {
```
JakubFornadel, line 31: I guess everything what is our code should be also in taraxa namespace

libraries/core_libs/node/src/node.cpp
```diff
@@ -118,6 +132,29 @@ void FullNode::init() {
                                        next_votes_mgr_, dag_mgr_, trx_mgr_);
 }
 
+void FullNode::setupMetricsUpdaters() {
+  auto network_metrics = metrics_->getMetrics<metrics::NetworkMetrics>();
+  network_metrics->setPeersCountUpdater([&network = network_]() { return network->getPeerCount(); });
```
JakubFornadel, line 137: I think in general it is better practice to use copy here -> network = network_ due to potential dangling reference and 1 time copy of pointer is cheap anyway

libraries/metrics/include/metrics/metrics_group.hpp
```diff
@@ -0,0 +1,71 @@
+#pragma once
+
+#include <prometheus/gauge.h>
+#include <prometheus/registry.h>
+
+#include <iostream>
+
+namespace metrics {
```
JakubFornadel, line 8: should be taraxa::metrics

libraries/metrics/include/metrics/metrics_group.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+using MetricGetter = std::function<double()>;
+using MetricUpdater = std::function<void()>;
+
+/**
+ * @brief add method that is setting specific gauge metric.
+ */
+#define ADD_GAUGE_METRIC(method, name, description)                                                  \
```
JakubFornadel, line 16: any other way how to solve this ? Using defines in modern C++ is pretty much avoided if possible

libraries/metrics/include/metrics/metrics_group.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+#include <prometheus/gauge.h>
+#include <prometheus/registry.h>
+
+#include <iostream>
+
+namespace metrics {
+
+using MetricGetter = std::function<double()>;
+using MetricUpdater = std::function<void()>;
```
JakubFornadel, line 11: These might be inside MetricsGroup class as they are related to that class ?

libraries/metrics/include/metrics/network_metrics.hpp
```diff
@@ -0,0 +1,14 @@
...
+#include "metrics/metrics_group.hpp"
+
+namespace metrics {
+class NetworkMetrics : public MetricsGroup {
+ public:
+  inline static const std::string group_name = "network";
+  NetworkMetrics(std::shared_ptr<prometheus::Registry> registry) : MetricsGroup(std::move(registry)) {}
+  ADD_GAUGE_METRIC_WITH_UPDATER(setPeersCount, "peers_count", "Count of peers that node is connected to")
+  ADD_GAUGE_METRIC_WITH_UPDATER(setDiscoveredPeersCount, "discovered_peers_count", "Count of discovered peers")
+  ADD_GAUGE_METRIC_WITH_UPDATER(setSyncingDuration, "syncing_duration_sec", "Time node is currently in sync state")
```
JakubFornadel, line 12: you could for example have some init function that would call normal "register" function to add these instead of defines for example ?

libraries/metrics/include/metrics/metrics_group.hpp
kstdl: Yes it isn't perfect. But it seems to me that this is better than copy-pasting a lot of code for every metric. Another option is to store some string->label mapping, so we can call some set method with the string. But metrics won't be strictly defined in this case. Small example. This code 
```cpp
  ADD_GAUGE_METRIC_WITH_UPDATER(setPeersCount, "peers_count", "Count of peers that node is connected to")
```
will become this for almost every metric
```cpp
void setPeersCount(double v) { 
  static auto& label = addMetric<prometheus::Gauge>(group_name + "_" +"peers_count",  "Count of peers that node is connected to").Add({});
  label.Set(v);
}
void setPeersCountUpdater(MetricGetter getter) {
  updaters_.push_back([this, getter]() { setPeersCount(getter()); });
}
```

libraries/core_libs/node/include/node/node.hpp
JakubFornadel: @MatusKysel so what is the purpose of taraxa namespace then ? when would you use it vs when not ?

libraries/config/include/config/network.hpp
kstdl: To have const fields here we should change the way we are parsing jsons(use struct constructor for it, not some custom function). Better to be a separate task

libraries/cli/include/cli/config_jsons/default/default_config.json
```diff
@@ -11,6 +11,10 @@
       "http_port": 9777,
       "ws_port": 6777
     },
+    "prometheus": {
```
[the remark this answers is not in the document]
kstdl: Makes sense. But if we will just remove it from config jsons there will be no option to check what could be enabled. And this requires changes in deployment scripts. So it should be changed somehow there. Could be a separate task

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "dag/dag.hpp"
 #include "logger/logger.hpp"
+#include "metrics/transaction_queue_metrics.hpp"
```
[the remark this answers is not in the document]
kstdl: I tried to use it in the classes before and just forgot to remove. Fixed

libraries/metrics/src/metrics_service.cpp
```diff
@@ -0,0 +1,35 @@
...
+void MetricsService::start() {
+  if (!exposer_) {
+    return;
+  }
+  thread_ = std::make_unique<std::thread>([&]() {
+    for (;;) {
+      for (auto& r : metrics_) {
+        r.second->updateData();
+      }
+      std::this_thread::sleep_for(std::chrono::milliseconds(polling_interval_ms_));
```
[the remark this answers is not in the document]
kstdl: It is running in dedicated thread already. So I don't think something more complicated will work better here

JakubFornadel (approved):

libraries/cli/include/cli/config_jsons/default/default_config.json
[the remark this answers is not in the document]
kstdl: https://github.com/Taraxa-project/taraxa-node/issues/2261

PR #2272, opened 2023-01-13
Fix proposed blocks develop

JakubFornadel:


JakubFornadel:
> LGTM. The change in storage means this will require a network reset correct?
yes we must reset devnet after this change

PR #2289, opened 2023-01-24
fix: prune unit test

mfrankovi:
When pruning state db, if block over proposal period were pruned, transaction estimation was not done correctly.
Changed the test to prune only block below proposal period

### DIFF
--- a/tests/full_node_test.cpp
+++ b/tests/full_node_test.cpp
@@ -714,12 +714,18 @@ TEST_F(FullNodeTest, sync_five_nodes) {
   context.assert_all_transactions_known();
   context.assert_balances_synced();
 
-  // Prune state_db of one node to remove all but last 10 block state and verify all ballances still match
+  // Prune state_db of one node
   auto prune_node = nodes[nodes.size() - 1];
-  auto last_block = prune_node->getPbftChain()->getPbftChainSize() - 10;
-  prune_node->getFinalChain()->prune(last_block);
-  context.assert_all_transactions_success();
-  context.assert_all_transactions_known();
+  const uint32_t min_blocks_to_prune = 50;
+  // This ensures that we never prune blocks that are over proposal period
+  ASSERT_HAPPENS({20s, 100ms}, [&](auto &ctx) {
+    const auto max_level = prune_node->getDagManager()->getMaxLevel();
+    const auto proposal_period = prune_node->getDB()->getProposalPeriodForDagLevel(max_level);
+    ASSERT_TRUE(proposal_period.has_value());
+    context.dummy_transaction();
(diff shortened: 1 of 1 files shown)


JakubFornadel (approved):

tests/full_node_test.cpp
```diff
@@ -714,12 +714,18 @@ TEST_F(FullNodeTest, sync_five_nodes) {
...
-  auto last_block = prune_node->getPbftChain()->getPbftChainSize() - 10;
-  prune_node->getFinalChain()->prune(last_block);
-  context.assert_all_transactions_success();
-  context.assert_all_transactions_known();
+  const uint32_t min_blocks_to_prune = 50;
+  // This ensures that we never prune blocks that are over proposal period
+  ASSERT_HAPPENS({100s, 100ms}, [&](auto &ctx) {
+    const auto max_level = prune_node->getDagManager()->getMaxLevel();
+    const auto proposal_period = prune_node->getDB()->getProposalPeriodForDagLevel(max_level);
+    EXPECT_TRUE(proposal_period.has_value());
```
kstdl, line 724: Shouldn't this be WAIT_EXPECT_TRUE?

tests/full_node_test.cpp
```diff
@@ -714,12 +714,18 @@ TEST_F(FullNodeTest, sync_five_nodes) {
...
-  // Prune state_db of one node to remove all but last 10 block state and verify all ballances still match
+  // Prune state_db of one node
   auto prune_node = nodes[nodes.size() - 1];
-  auto last_block = prune_node->getPbftChain()->getPbftChainSize() - 10;
-  prune_node->getFinalChain()->prune(last_block);
-  context.assert_all_transactions_success();
-  context.assert_all_transactions_known();
+  const uint32_t min_blocks_to_prune = 50;
+  // This ensures that we never prune blocks that are over proposal period
+  ASSERT_HAPPENS({100s, 100ms}, [&](auto &ctx) {
```
kstdl, line 721: Is it normal that this test could possibly run for 100s? That is very big value and we should set as low value as possible for this waits imo

tests/full_node_test.cpp
mfrankovi: No, this should never be false.

tests/full_node_test.cpp
mfrankovi: Yes, I put that while I was testing it. It should not be more than 10s. I will replace it with 20s

tests/full_node_test.cpp
kstdl: So maybe better to use ASSERT_TRUE here? Maybe EXPECT is also ok here, but it would be clear with assert that with instantly fail the test

mfrankovi: Replaced with ASSERT_TRUE

PR #2300, opened 2023-01-31
Issue 2299/ddos protection

JakubFornadel:

### DIFF
--- a/libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
+++ b/libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
@@ -331,18 +331,21 @@ void TaraxaCapability::interpretCapabilityPacket(std::weak_ptr<dev::p2p::Session
     return;
   }
 
+  const auto [hp_queue_size, mp_queue_size, lp_queue_size] = thread_pool_->getQueueSize();
+  const size_t tp_queue_size = hp_queue_size + mp_queue_size + lp_queue_size;
+
   // Check max allowed packets queue size
   if (kConf.network.ddos_protection.max_packets_queue_size) {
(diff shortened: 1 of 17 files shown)


libraries/core_libs/network/src/tarcap/taraxa_capability.cpp
```diff
@@ -331,18 +331,21 @@ void TaraxaCapability::interpretCapabilityPacket(std::weak_ptr<dev::p2p::Session
...
+  const size_t tp_queue_size = hp_queue_size + mp_queue_size + lp_queue_size;
+
   // Check max allowed packets queue size
   if (kConf.network.ddos_protection.max_packets_queue_size) {
-    const auto [hp_queue_size, mp_queue_size, lp_queue_size] = thread_pool_->getQueueSize();
-    if (hp_queue_size + mp_queue_size + lp_queue_size > kConf.network.ddos_protection.max_packets_queue_size) {
+    if (tp_queue_size > kConf.network.ddos_protection.max_packets_queue_size) {
       LOG(log_wr_) << "Ignored " << convertPacketTypeToString(packet_type) << ". Max allowed packets queue size "
-                   << kConf.network.ddos_protection.max_packets_queue_size << " exceeded";
+                   << kConf.network.ddos_protection.max_packets_queue_size << " exceeded: " << tp_queue_size;
```
mfrankovi, line 341: I don't think we should ever ignore packets. This can have other consequences later on and throw different errors. If our queue size is too big I think we should start disconnecting nodes with most traffic and not ignore packets.

JakubFornadel: @mfrankovi fixed

mfrankovi (approved):

PR #2315, opened 2023-02-13
chore: dag production dependent on stake

mfrankovi:
VRF threshold is created for each stake vote. Minimal vote threshold is chosen for the node. This makes dag blocks production dependent on stake without increasing the occurrence of stale blocks.

Default threshold_upper value is reduced because the initial high value causes too many blocks to be produces when new network is starting.

Efficiency of 50% is chosen for target. With efficiency of 70% and the new algorithm stale block although rare would appear more often.

For minimum difficulty always propose a dag block even if someone has just produced the block of the same level. Since minimum difficulty calculates quickly there is still a big chance for a reward.

Sending transactions and dag block always in same order would give advantage to the node that receives it first. This is specially true if faucet is used for most of the transactions. Add random order when gossiping transactions and dag blocks.


libraries/common/include/common/vrf_wrapper.hpp
```diff
@@ -54,7 +54,22 @@ class VrfSortitionBase {
...
  private:
-  void thresholdFromOutput() const { threshold_ = (((uint16_t)output_[1] << 8) | output_[0]); }
+  void thresholdFromOutput(uint16_t stake) const {
+    threshold_ = (((uint16_t)output_[1] << 8) | output_[0]);
+    if (stake > 1) {
+      uint16_t min_threshold = threshold_;
+      uint16_t threshold_candidate = threshold_;
+      // Generate sequence of thresholds using simple generator with original threshold as seed
+      for (uint16_t stake_counter = 1; stake_counter < stake; stake_counter++) {
+        const uint32_t a = 48271;  // Multiplier term used by C++11's minstd_rand
```
kstdl, line 64: I think it shouldn't be inside of this for cycle. And why is it uint32_t result of its multiplication assigns to uint16_t? I think it should be also uint16_t

libraries/vdf/include/vdf/sortition.hpp
```diff
@@ -62,11 +62,14 @@ class VdfSortition : public vrf_wrapper::VrfSortitionBase {
...
       "9f9330e44715b3f2066f903081836c1c83ca29126f8fdc5f5922bf3f9ddb4540171691ac"
       "cc1ef6a34b2a804a18159c89c39b16edee2ede35");
-  bool verifyVrf(const vrf_pk_t& pk, const bytes& vrf_input) const;
+  bool verifyVrf(const vrf_pk_t& pk, const bytes& vrf_input, uint16_t stake) const;
 
   std::pair<bytes, bytes> vdf_sol_;
   unsigned long vdf_computation_time_ = 0;
   uint16_t difficulty_ = 0;
+  // Stake is divided by 1/1000 or 0.1%
+  static const uint32_t kStakeDivider = 1000;
```
kstdl, line 71: Isn't `kVotesMultiplier` name fits better here?

libraries/vdf/include/vdf/sortition.hpp
```diff
@@ -62,11 +62,14 @@ class VdfSortition : public vrf_wrapper::VrfSortitionBase {
       "9aee2a207e5173a7ee8f90ee9ab9b6a745d27c6e850e7ca7332388dfef7e5bbe6267d1f7"
       "9f9330e44715b3f2066f903081836c1c83ca29126f8fdc5f5922bf3f9ddb4540171691ac"
       "cc1ef6a34b2a804a18159c89c39b16edee2ede35");
-  bool verifyVrf(const vrf_pk_t& pk, const bytes& vrf_input) const;
+  bool verifyVrf(const vrf_pk_t& pk, const bytes& vrf_input, uint16_t stake) const;
```
kstdl, line 65: stake isn't really stake here, but some value proportional to votes/total_votes. I was a little bit confused with it first time because thought that we are passing real stake and should be iterating too much because of it

mfrankovi: It is stake in promil (part per thousand). I did it this way because I wanted it to not depend on total number of votes and act similar if total votes count is small or large.

kstdl: But you are passing `vote_count * kStakeDivider / total_vote_count` here. So it is not related to stake, but to votes count. Or am I missing something?

libraries/vdf/include/vdf/sortition.hpp
mfrankovi: It defines that I am using stakes in "part per thousand". The total stake, all votes counts are divided per thousand so I think divider is a better name. This is why I called it that way. Maybe simply kPartPerThousandDivider?

libraries/common/include/common/vrf_wrapper.hpp
mfrankovi: yes, I will move it out and change to uint16_t

kstdl (approved):

JakubFornadel (approved):

PR #2322, opened 2023-02-15
fix: incorrect binary for rpc_test in ctest

kstdl:
Fix incorrect binary in ctest for `rpc_test`. So we didn't executed this at all during ctest runs


JakubFornadel (approved):

kstdl:
Also included fix for gas values in `RPCTest.eth_estimateGas`

PR #2331, opened 2023-02-17
chore: fix crash on getting account from future blocks

MatusKysel:


kstdl (approved):

JakubFornadel (approved):

justinsnapp:
@MatusKysel  we want to also update mainnet and testnet with this fix right?

PR #2352, opened 2023-02-28
chore: optimize rewards votes handling

mfrankovi:
The goal of this optimization was to avoid saving cert votes twice in the db, once as a cert vote when cert voting a block and second time when treating the same votes as reward votes.

- 2t+1 cert votes are no longer saved once node receives 2t+1 votes but only a short time later when we actually push a pbft block to the chain. They are saved in a db batch as part of an atomic db operation when moving to the next period.
- Additional reward votes that come after pushing the pbft block into chain is saved in reward_votes column
- Total reward votes will consists of votes bundle saved as 2t+1 cert votes and additional reward votes saved individually

### DIFF
--- a/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
+++ b/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
@@ -96,13 +96,16 @@ class VoteManager {
   std::optional<PbftRound> determineNewRound(PbftPeriod current_pbft_period, PbftRound current_pbft_round);
 
   /**
-   * @brief Replace current reward votes info with new period, round & block hash based on vote
+   * @brief Replace current reward votes with new period, round & block hash based on vote
    *
    * @param period
    * @param round
+   * @param step
    * @param block_hash
+   * @param batch
    */
-  void resetRewardVotesInfo(PbftPeriod period, PbftRound round, const blk_hash_t& block_hash);
+  void resetRewardVotes(PbftPeriod period, PbftRound round, PbftStep step, const blk_hash_t& block_hash,
+                        DbStorage::Batch& batch);
 
   /**
    * @brief Check reward votes for specified pbft block
@@ -115,11 +118,11 @@ class VoteManager {
                                                                        bool copy_votes);
 
   /**
-   * @brief Get reward votes from reward_votes_ with the round during which was the previous block pushed
+   * @brief Get reward votes with the round during which was the previous block pushed
    *
    * @return vector of reward votes
    */
-  std::vector<std::shared_ptr<Vote>> getProposeRewardVotes();
+  std::vector<std::shared_ptr<Vote>> getRewardVotes();
 
   /**
    * @brief Get current reward votes pbft block period
@@ -280,6 +283,7 @@ class VoteManager {
   blk_hash_t reward_votes_block_hash_;
   PbftRound reward_votes_period_;
   PbftRound reward_votes_round_;
+  std::vector<vote_hash_t> extra_reward_votes_;
   mutable std::shared_mutex reward_votes_info_mutex_;
 
   // Cache for current 2T+1 - <Vote type, <period, two_t_plus_one for period>>
--- a/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
+++ b/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
@@ -32,10 +32,18 @@ VoteManager::VoteManager(const addr_t& node_addr, const PbftConfig& pbft_config,
   auto db_votes = db_->getAllTwoTPlusOneVotes();
 
   auto loadVotesFromDb = [this](const std::vector<std::shared_ptr<Vote>>& votes) {
+    bool reward_votes_info_set = false;
     for (const auto& vote : votes) {
       // Check if votes are unique per round, step & voter
       if (!isUniqueVote(vote).first) {
         continue;
+      }
+
+      if (!reward_votes_info_set && vote->getType() == PbftVoteTypes::cert_vote) {
+        reward_votes_info_set = true;
+        reward_votes_block_hash_ = vote->getBlockHash();
+        reward_votes_period_ = vote->getPeriod();
+        reward_votes_round_ = vote->getRound();
       }
 
       addVerifiedVote(vote);
@@ -45,11 +53,9 @@ VoteManager::VoteManager(const addr_t& node_addr, const PbftConfig& pbft_config,
 
   loadVotesFromDb(db_->getAllTwoTPlusOneVotes());
   loadVotesFromDb(db_->getOwnVerifiedVotes());
-
-  if (const auto reward_votes = db_->getRewardVotes(); !reward_votes.empty()) {
-    loadVotesFromDb(reward_votes);
-    resetRewardVotesInfo(reward_votes[0]->getPeriod(), reward_votes[0]->getRound(), reward_votes[0]->getBlockHash());
-  }
+  auto reward_votes = db_->getRewardVotes();
+  for (const auto& vote : reward_votes) extra_reward_votes_.emplace_back(vote->getHash());
+  loadVotesFromDb(reward_votes);
 }
 
 void VoteManager::setNetwork(std::weak_ptr<Network> network) { network_ = std::move(network); }
@@ -227,9 +233,9 @@ bool VoteManager::addVerifiedVote(const std::shared_ptr<Vote>& vote) {
     LOG(log_nf_) << "Added verified vote: " << hash;
     LOG(log_dg_) << "Added verified vote: " << *vote;
 
-    // Save in db only those reward votes that have the same round as round during which we pushed the block into chain
-    if (is_valid_potential_reward_vote && reward_votes_round_ == vote->getRound()) {
-      db_->saveRewardVote(vote);
+    if (is_valid_potential_reward_vote) {
+      extra_reward_votes_.emplace_back(vote->getHash());
+      db_->saveExtraRewardVote(vote);
     }
 
     const auto total_weight = (found_voted_value_it->second.first += weight);
@@ -279,7 +285,9 @@ bool VoteManager::addVerifiedVote(const std::shared_ptr<Vote>& vote) {
           {two_plus_one_voted_block_type, std::make_pair(vote->getBlockHash(), vote->getStep())});
 
       // Save only current pbft period & round 2t+1 votes bundles into db
-      if (vote->getPeriod() == current_pbft_period_ && vote->getRound() == current_pbft_round_) {
+      // Cert votes are saved once the pbft block is pushed in the chain
+      if (vote->getType() != PbftVoteTypes::cert_vote && vote->getPeriod() == current_pbft_period_ &&
+          vote->getRound() == current_pbft_round_) {
         std::vector<std::shared_ptr<Vote>> votes;
         votes.reserve(found_voted_value_it->second.second.size());
         for (const auto& tmp_vote : found_voted_value_it->second.second) {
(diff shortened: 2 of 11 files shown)


libraries/core_libs/storage/include/storage/storage.hpp
```diff
@@ -143,6 +143,7 @@ class DbStorage : public std::enable_shared_from_this<DbStorage> {
   std::atomic<bool> snapshots_enabled_ = true;
   const uint32_t kDbSnapshotsMaxCount = 0;
   std::set<PbftPeriod> snapshots_;
+  std::vector<vote_hash_t> reward_votes_keys_;
```
JakubFornadel, line 146: if you put this cache into VoteManager, we would not need to call getRewardVotes() from db, which is still using db iterator but it could use this in memory vector. getRewardVotes() would be called only during node startup to init this cache

mfrankovi: I have rewritten this PR, please review all of it again

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -38,6 +38,12 @@ VoteManager::VoteManager(const addr_t& node_addr, const PbftConfig& pbft_config,
         continue;
       }
 
+      if (vote->getType() == PbftVoteTypes::cert_vote) {
+        reward_votes_block_hash_ = vote->getBlockHash();
+        reward_votes_period_ = vote->getPeriod();
+        reward_votes_round_ = vote->getRound();
+      }
```
JakubFornadel, line 45: this will happen 2t+1 times...

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -227,9 +230,9 @@ bool VoteManager::addVerifiedVote(const std::shared_ptr<Vote>& vote) {
     LOG(log_nf_) << "Added verified vote: " << hash;
     LOG(log_dg_) << "Added verified vote: " << *vote;
 
-    // Save in db only those reward votes that have the same round as round during which we pushed the block into chain
-    if (is_valid_potential_reward_vote && reward_votes_round_ == vote->getRound()) {
```
JakubFornadel, line 231: we dont need to check round anymore ?

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -279,7 +282,9 @@ bool VoteManager::addVerifiedVote(const std::shared_ptr<Vote>& vote) {
           {two_plus_one_voted_block_type, std::make_pair(vote->getBlockHash(), vote->getStep())});
 
       // Save only current pbft period & round 2t+1 votes bundles into db
-      if (vote->getPeriod() == current_pbft_period_ && vote->getRound() == current_pbft_round_) {
+      // Cert votes are saved once the pbft block is pushed in the chain
```
JakubFornadel, line 285: 1 potential problem I see here is that if we are not able to push the block due to missing tx/dag block and we want to restart the node to reset caches etc... we will lose those votes

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -537,26 +542,55 @@ std::optional<PbftRound> VoteManager::determineNewRound(PbftPeriod current_pbft_
...
+
+        // 2t+1 votes block already set
+        if (found_two_t_plus_one_voted_block != found_round_it->second.two_t_plus_one_voted_blocks_.end()) {
+          assert(found_two_t_plus_one_voted_block->second.first == block_hash);
+          auto found_voted_value_it = found_step_it->second.votes.find(block_hash);
+          // Add voted value
+          if (found_voted_value_it != found_step_it->second.votes.end()) {
+            std::vector<std::shared_ptr<Vote>> votes;
+            votes.reserve(found_voted_value_it->second.second.size());
+            for (const auto& tmp_vote : found_voted_value_it->second.second) {
```
JakubFornadel, line 577: why do we need this if you can just send vector of cert votes from the place where we call resetRewardVotes ?

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -537,26 +542,55 @@ std::optional<PbftRound> VoteManager::determineNewRound(PbftPeriod current_pbft_
...
+      }
+    }
   }
 
   LOG(log_dg_) << "Reward votes info reset to: block_hash: " << block_hash << ", period: " << period
                << ", round: " << round;
 }
 
 bool VoteManager::isValidRewardVote(const std::shared_ptr<Vote>& vote) const {
-  std::shared_lock lock(reward_votes_info_mutex_);
```
JakubFornadel, line 558: so what is protecting reward_votes_block_hash_ and reward_votes_period_ now ?

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -686,10 +719,9 @@ std::pair<bool, std::vector<std::shared_ptr<Vote>>> VoteManager::checkRewardVote
   return {false, {}};
 }
 
-std::vector<std::shared_ptr<Vote>> VoteManager::getProposeRewardVotes() {
-  std::shared_lock lock(reward_votes_info_mutex_);
-  const auto reward_votes = getTwoTPlusOneVotedBlockVotes(reward_votes_period_, reward_votes_round_,
-                                                          TwoTPlusOneVotedBlockType::CertVotedBlock);
+std::vector<std::shared_ptr<Vote>> VoteManager::getRewardVotes() {
+  auto reward_votes = getTwoTPlusOneVotedBlockVotes(reward_votes_period_, reward_votes_round_,
```
JakubFornadel, line 723: no need for mutex protection ?

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
```diff
@@ -280,7 +283,8 @@ class VoteManager {
   blk_hash_t reward_votes_block_hash_;
   PbftRound reward_votes_period_;
   PbftRound reward_votes_round_;
-  mutable std::shared_mutex reward_votes_info_mutex_;
+  // reward_votes_ will only keep additional reward votes which are not part of the original 2t+1 cert votes
```
JakubFornadel, line 286: lets call it extra_reward_votes or something similar

tests/network_test.cpp
```diff
@@ -629,7 +629,10 @@ TEST_F(NetworkTest, node_pbft_sync) {
   period_data2.transactions.push_back(g_signed_trx_samples[3]);
 
   db1->savePeriodData(period_data2, batch);
-  db1->replaceRewardVotes(votes_for_pbft_blk2, batch);
+  std::cout << "Before VOTE cc" << std::endl;
```
JakubFornadel, line 632: remove

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: yes, basically all of the 2t+1 cert votes should have this value the same and this is done on startup. I will change it with some flag to do only once.

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: There is no harm to store reward votes from different rounds. Before we used it to enforce that the same round is used for all reward votes, but now this information is kept in 2t+1 votes so we use whatever round was in 2t+1 votes

kstdl (approved):

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: if we came to the point of having 2t+1 cert votes without having a pbft block or dag block there is something else seriously wrong on this specific node. Since every node saves its own votes anyhow I considered this scenario and from the two options:
1. As implemented in PR accepting the loss of votes in the scenario
2. Saving 2t+1 votes before pushing a new pbft block to the chain and then saving it in another storage after pushing block to the chain, decrease performance since we need two db saves and a copy

I would choose 1. but if everyone think 2. is a safer option I am not strongly set against it. But if we go to option 2. one could argue option 3. where we save all cert votes all the time not to lose any even before we have 2t+1. Because currently we only save to db when we reach 2t+1. If we restart before reaching 2t+1 we do not save and we lose these votes

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: There is a race condition. Some new votes might have arrived in the meantime and be stored in verified_votes and then we would lose those.

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: I have put everything under the same mutex verified_votes_access_. Maybe the name of the mutex should be changed. This method isValidRewardVote is private and it is only called within this class from the code already protected by the mutex.

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: getTwoTPlusOneVotedBlockVotes has a mutex protection inside

tests/network_test.cpp
mfrankovi: leftover debugging code. I will remove it

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -537,26 +542,55 @@ std::optional<PbftRound> VoteManager::determineNewRound(PbftPeriod current_pbft_
...
+
+  auto found_period_it = verified_votes_.find(period);
+  // Add period
+  if (found_period_it != verified_votes_.end()) {
+    auto found_round_it = found_period_it->second.find(round);
+    // Add round
+    if (found_round_it != found_period_it->second.end()) {
+      auto found_step_it = found_round_it->second.step_votes.find(step);
+      // Add step
+      if (found_step_it != found_round_it->second.step_votes.end()) {
```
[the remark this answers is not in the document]
mfrankovi: Changed to using asserts because votes always need to be there

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: Fixed

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
mfrankovi: Fixed

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
JakubFornadel: the harm is in performance that we save each such vote in db, even though we dont need it ?

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
JakubFornadel: you have to put the lock before calling getTwoTPlusOneVotedBlockVotes, not after... and then you will again lock the same lock inside getTwoTPlusOneVotedBlockVotes. Based on usage seems like the orig. lock that protected just this info regarding period, round and hash of reward votes was better approach as these are separate info unrelated to verified_votes and its lock

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -537,26 +544,44 @@ std::optional<PbftRound> VoteManager::determineNewRound(PbftPeriod current_pbft_
...
+                                   DbStorage::Batch& batch) {
+  // Save 2t+1 cert votes to database, remove old reward votes
+  std::scoped_lock lock(verified_votes_access_);
+  reward_votes_block_hash_ = block_hash;
+  reward_votes_period_ = period;
+  reward_votes_round_ = round;
+
+  auto found_period_it = verified_votes_.find(period);
+  assert(found_period_it != verified_votes_.end());
+  auto found_round_it = found_period_it->second.find(round);
```
JakubFornadel, line 561: segfault/or whatever would happen right here as found_period_it might be == end()....
Those asserts are ignored in release mode so you can use them but you have to use it like this:

```
auto found_period_it = verified_votes_.find(period);
if (found_period_it == verified_votes_.end()) {
  asset
  return
}
```

mfrankovi: Fixed

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: Fixed with the single mutex approach now. The problem with having two mutexes as before is that now with my change there are a lot of places in the code where we would need to lock both together or one after another and I fear that increases chance for a deadlock.

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -699,6 +748,11 @@ std::vector<std::shared_ptr<Vote>> VoteManager::getProposeRewardVotes() {
     return {};
   }
 
+  // Add additional reward votes that were not part of 2t+1
```
JakubFornadel, line 751: not part of 2t+1 ? All you need here is to call original method `getTwoTPlusOneVotedBlockVotes` - it does not return 2t+1 votes, it returns all founf votes for 2t+1 voted block

having to create getTwoTPlusOneVotedBlockVotesInternal because it uses the same lock is again a sign of not ideal architecture

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
JakubFornadel: no need to worry about deadlock - this new mutex is locked only in single place when we reset reward votes so it is very easy to check for any possibility of deadlock... based on some of the changes in this PR, which were made just because of this single lock usage, it sort of shows the bad design

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
```diff
@@ -280,7 +295,8 @@ class VoteManager {
   blk_hash_t reward_votes_block_hash_;
   PbftRound reward_votes_period_;
   PbftRound reward_votes_round_;
-  mutable std::shared_mutex reward_votes_info_mutex_;
+  // extra_reward_votes_ will only keep additional reward votes which are not part of the original 2t+1 cert votes
+  std::vector<std::shared_ptr<Vote>> extra_reward_votes_;
```
JakubFornadel, line 299: based on this comment https://github.com/Taraxa-project/taraxa-node/pull/2352/files#r1136811150, I dont think you actually need this memory vector at all - you need to store those reward votes only in db and then load it during node init

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -227,9 +232,9 @@ bool VoteManager::addVerifiedVote(const std::shared_ptr<Vote>& vote) {
     LOG(log_nf_) << "Added verified vote: " << hash;
     LOG(log_dg_) << "Added verified vote: " << *vote;
 
-    // Save in db only those reward votes that have the same round as round during which we pushed the block into chain
-    if (is_valid_potential_reward_vote && reward_votes_round_ == vote->getRound()) {
+    if (is_valid_potential_reward_vote) {
       db_->saveRewardVote(vote);
```
JakubFornadel, line 236: we could also call this saveExtraRewardVote as well as db column to show that not all reward votes are store here as previously

libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
mfrankovi: I removed it. The only downside of removing it is when we delete column in reward votes we will now delete with all the votes hashes. But from what I checked this is fast and it should not be a problem. Alternative would be to keep the vote hashes or extra reward votes only for deleting the db

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: Done

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: one of the things I did not like in previous code was that we are locking both locks at the same time. This always introduces risk of deadlock. I have returned it back to using two mutexes but with some changes to never lock it together

libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
mfrankovi: Renamed everything related to extra rewards

PR #2363, opened 2023-03-02
chore: refactor reward votes saving

mfrankovi:
Reward votes are no longer saved separately from cert_votes. All the need for reward votes is handled from verified_votes cache and db. This saves both space and makes processing votes and syncing simple and faster.

### DIFF
--- a/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
+++ b/libraries/core_libs/consensus/include/vote_manager/vote_manager.hpp
@@ -203,6 +203,16 @@ class VoteManager {
                                                                    TwoTPlusOneVotedBlockType type) const;
 
   /**
+   * Get 2t+1 voted block votes for first round with specific period and type, e.g. soft/cert/next voted block
+   *
+   * @param period
+   * @param type
+   * @return vector of votes if 2t+1 voted block votes found, otherwise empty vector
+   */
+  std::vector<std::shared_ptr<Vote>> getTwoTPlusOneVotedBlockVotes(PbftPeriod period,
+                                                                   TwoTPlusOneVotedBlockType type) const;
+
+  /**
    * Get all 2t+1 voted block next votes(both for null block as well as specific block) for specific period and round
    *
    * @param period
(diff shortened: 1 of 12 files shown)


JakubFornadel (commented):
reward votes stored separately because in case of node restart we need all of them, not just 2t+1. What happens if the node restarts and looses reward votes from memory but received block that contains some of the votes that were just in memory (not in db) and because of restart we lost them ?

mfrankovi:
> reward votes were stored separately because in case of node restart we need all of them, not just 2t+1. What happens if the node restarts and looses reward votes from memory but received block that contains some of the votes that were just in memory (not in db) and because of restart we lost them ?
> 
> In general we need just 2t+1 soft/next votes as we check if 2t+1 votes are present during voting steps... For reward votes we simply need all of them as we are validating each one, not just 2t+1 and thats why they were stored in db in different way than the rest
> 
> I think this PR #2352 is a way to go

But even though we only need 2t+1 during voting steps after we do have 2t+1 I think it is better to simply keep the votes in the same column. Any new vote that comes after that which is not needed for 2t+1 but as a "reward vote" will still be saved in the same column. Before we would save the same vote two times first as a cert vote in latest_round_two_t_plus_one_votes and than later again in reward votes column. Now it is only saved once in one column.

If a restart happens all 2t+1 and additional reward votes are stored in Columns::latest_round_two_t_plus_one_votes. So I do not see us losing any reward vote on restart unless I am missing something?

JakubFornadel:
> > reward votes were stored separately because in case of node restart we need all of them, not just 2t+1. What happens if the node restarts and looses reward votes from memory but received block that contains some of the votes that were just in memory (not in db) and because of restart we lost them ?
> > In general we need just 2t+1 soft/next votes as we check if 2t+1 votes are present during voting steps... For reward votes we simply need all of them as we are validating each one, not just 2t+1 and thats why they were stored in db in different way than the rest
> > I think this PR #2352 is a way to go
> 
> But even though we only need 2t+1 during voting steps after we do have 2t+1 I think it is better to simply keep the votes in the same column. Any new vote that comes after that which is not needed for 2t+1 but as a "reward vote" will still be saved in the same column. Before we would save the same vote two times first as a cert vote in latest_round_two_t_plus_one_votes and than later again in reward votes column. Now it is only saved once in one column.
> 
> If a restart happens all 2t+1 and additional reward votes are stored in Columns::latest_round_two_t_plus_one_votes. So I do not see us losing any reward vote on restart unless I am missing something?

Scenario:
- there are 4 nodes in network, each of wich produces 1 vote so there are 4 cert votes in total:  A, B, C, D for each block
- you receive cert votes A,B,C (2t+1) for block 1, you save them into db, push block into chain and move to period 2  
- now you receive delayed cert vote D for block 1, but it is not saved into db in your solution (just in memory)
- someone proposes block 2 with reward votes A,B,C,D
- your node gets restarted
- you just lost the reward vote D because it was not saved in db before restart

PR #2388, opened 2023-03-16
chore: adds transaction generating script

rattrap:


charts/taraxa-node/templates/secrets.yaml
```diff
@@ -36,5 +36,6 @@ data:
   SLACK_TOKEN: {{ .Values.slack.token | b64enc | quote }}
   EXPLORER_DELEGATION_PRIVATE_KEY: {{ .Values.config.consensusnode.explorerDelegationPrivateKey | b64enc | quote }}
   EXPLORER_FAUCET_PRIVATE_KEY: {{ .Values.explorer.faucet.privKey | b64enc | quote }}
+  TRANSACTION_GENERATION_PRIVATE_KEY: {{ .Values.transactionGeneration.privKey | b64enc | quote }}
```
rjonczy, line 39: keep in mind to add this into `SealedSecret` in gitops-ledger repo later

rjonczy: Can't find modification of chart version in `Chart.yaml`

charts/taraxa-node/templates/transaction-generation-script.yaml
```diff
@@ -0,0 +1,126 @@
...
+  transactions.py: |-
+    import logging
+    import coloredlogs
+    import time
+    from dotenv import load_dotenv
+    from os import getenv
+    from web3 import Web3
+
+    logger = logging.getLogger()
+    coloredlogs.install(level=logging.DEBUG, logger=logger)
```
rjonczy, line 66: could that be pulled from ENV? So we can have an impact on logging level

charts/taraxa-node/templates/transaction-generation-service.yaml
```diff
@@ -0,0 +1,73 @@
+{{ if .Values.transactionGeneration.enabled }}
```
rjonczy, line 1: lets rename file to: `transaction.generation.statefulset.yaml`. I have another PR, which cleans and renames files.

charts/taraxa-node/templates/transaction-generation-service.yaml
```diff
@@ -0,0 +1,73 @@
...
+      containers:
+        - name: transaction-generation
+          image: "python:3.8"
+          imagePullPolicy: IfNotPresent
+          env:
+          - name: PRIVATE_KEY
+            valueFrom:
+              secretKeyRef:
+                name: {{ .Release.Name }}
+                key: TRANSACTION_GENERATION_PRIVATE_KEY
```
rjonczy, line 42: can't we use the same name var name `PRIVATE_KEY` and key in secret ?

rjonczy (changes requested):

charts/taraxa-node/templates/transaction-generation-service.yaml
rattrap: The secret is for the whole taraxa-node deployment. There are a lot of private keys there.

PR #2397, opened 2023-03-20
Fix peer chainsize testnet

JakubFornadel:
## Purpose

Fix from develp that we need also in testnet:
https://github.com/Taraxa-project/taraxa-node/pull/2386

### DIFF
--- a/libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
+++ b/libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
@@ -63,7 +63,7 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
     peer->markPbftBlockAsKnown(pbft_block->getBlockHash());
   }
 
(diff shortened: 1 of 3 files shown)


libraries/core_libs/network/src/tarcap/packets_handlers/vote_packet_handler.cpp
```diff
@@ -63,7 +63,9 @@ void VotePacketHandler::process(const PacketData &packet_data, const std::shared
     peer->markPbftBlockAsKnown(pbft_block->getBlockHash());
   }
 
-  processVote(vote, pbft_block, peer, true);
+  if (processVote(vote, pbft_block, peer, true)) {
+    return;
```
mfrankovi, line 67: Why return if vote was processed successfully?

JakubFornadel: yeah thats bug - I wonder how is it possible that it works on devnet

kstdl (approved):

PR #2411, opened 2023-03-23
[helm-chart]: labels for kube-monkey

rjonczy:
## Purpose

We want to be able to opt-in for chaos. Used [chaos-monkey](https://github.com/asobti/kube-monkey)

Opt-ins for chaos by adding labels to `StatefulSet`s.

## Changes made

Added labels to pod templates (and metamata of `StatefullSep` app as well) in `StatefulSet` of boot, rpc and consensus nodes.


rattrap (approved):
LGTM

PR #2431, opened 2023-04-04
Migrations

kstdl:
1. Add migration functionality including base class for migrations and migration manager that will apply registered migrations. 
2. Add migration for transactions hashes

### DIFF
--- a/libraries/core_libs/storage/include/storage/migration/transaction_hashes.hpp
+++ b/libraries/core_libs/storage/include/storage/migration/transaction_hashes.hpp
@@ -8,8 +8,9 @@
 namespace taraxa::storage::migration {
 class TransactionHashes : public migration::Base {
  public:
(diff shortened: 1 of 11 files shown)


libraries/core_libs/storage/include/storage/migration/transaction_hashes.hpp
```diff
@@ -0,0 +1,15 @@
...
+
+#include "final_chain/final_chain.hpp"
+#include "storage/migration/migration_base.hpp"
+#include "transaction/transaction.hpp"
+
+namespace taraxa::storage::migration {
+class TransactionHashes : public migration::Base {
+ public:
+  TransactionHashes(std::shared_ptr<DbStorage> db) : migration::Base(db) {}
+  std::string id() override { return "TransactionHashes"; }
```
JakubFornadel, line 12: as we have also cpp file, lets move all implementations there ?

JakubFornadel (approved):

PR #2435, opened 2023-04-05
hotfix: gas estimations for transactions with internal calls

kstdl:
Hotfix for transactions estimation. Before this fix some transactions with internal calls to other contracts(and maybe in some other cases) was running out of gas. So estimations wasn't correct

### DIFF
--- a/libraries/core_libs/network/rpc/eth/Eth.cpp
+++ b/libraries/core_libs/network/rpc/eth/Eth.cpp
@@ -56,14 +56,50 @@ class EthImpl : public Eth, EthParams {
     auto t = toTransactionSkeleton(_json);
     auto blk_n = parse_blk_num(_blockNumber);
     prepare_transaction_for_call(t, blk_n);
-    return toJS(call(blk_n, t).code_retval);
+    auto ret = call(blk_n, t);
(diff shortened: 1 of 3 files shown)


libraries/core_libs/network/rpc/eth/Eth.cpp
```diff
@@ -56,14 +56,52 @@ class EthImpl : public Eth, EthParams {
...
+
+    auto is_enough_gas = [&](gas_t gas) -> bool {
+      t.gas = gas;
+      auto res = call(blk_n, t);
+      if (!res.consensus_err.empty()) {
+        std::cout << "res.consensus_err " << res.consensus_err << std::endl;
+        throw std::runtime_error(res.consensus_err);
+      }
+      if (!res.code_err.empty()) {
+        std::cout << "res.code_err " << res.code_err << std::endl;
```
JakubFornadel, line 79: do we want to have couts here ?

JakubFornadel (approved):

libraries/core_libs/network/rpc/eth/Eth.cpp
kstdl: definitely not

kstdl:
We will include it into next release, but lets have branch for now to merge it if we will need it

PR #2459, opened 2023-04-24
Reward votes initialization

mfrankovi:
Set reward votes period and round from 2t+1 votes

### DIFF
--- a/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
+++ b/libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
@@ -31,19 +31,25 @@ VoteManager::VoteManager(const addr_t& node_addr, const PbftConfig& pbft_config,
 
   auto db_votes = db_->getAllTwoTPlusOneVotes();
 
-  auto addVerifiedVotes = [this](const std::vector<std::shared_ptr<Vote>>& votes) {
-    bool reward_votes_info_set = false;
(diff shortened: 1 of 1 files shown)


libraries/core_libs/consensus/src/vote_manager/vote_manager.cpp
```diff
@@ -31,8 +31,8 @@ VoteManager::VoteManager(const addr_t& node_addr, const PbftConfig& pbft_config,
 
   auto db_votes = db_->getAllTwoTPlusOneVotes();
 
-  auto addVerifiedVotes = [this](const std::vector<std::shared_ptr<Vote>>& votes) {
-    bool reward_votes_info_set = false;
+  bool reward_votes_info_set = false;
+  auto addVerifiedVotes = [this, &reward_votes_info_set](const std::vector<std::shared_ptr<Vote>>& votes) {
```
JakubFornadel, line 35: If someone in the future change the order and calls addVerifiedVotes(own_votes); before addVerifiedVotes(db_->getAllTwoTPlusOneVotes()); it will not work...

It woulb be more error prone if the lambd function look like this:
```
auto addVerifiedVotes = [this, &](const std::vector<std::shared_ptr<Vote>>& votes, bool reward_votes_info_set = false) {
  ...
}

addVerifiedVotes(db_->getAllTwoTPlusOneVotes(), true);

```

mfrankovi: Done

PR #2488, opened 2023-05-23
feat: accept decimal and hex as int value inputs

kstdl:
Accept decimal values, not only hexes for ints


mfrankovi:
The change seems good. One thing that I am worried is do we have some scripts/code/rpc expecting this to be treated as hex and this change might brake it. Because there is a lot of places in the code where jsToInt is used.

mfrankovi (approved):

PR #2501, opened 2023-06-06
bugfix: do not ignor pbft size check on syncing

MatusKysel:
## Purpose
We had this feature https://github.com/Taraxa-project/taraxa-node/commit/fc1e61bb1c07f52c31726e0b80cd895618a29e1d in our code, but since we are marking peer as malicious when is requesting syncing to higher period as we have available we have to remove this


mfrankovi (approved):

JakubFornadel:
why are there empty changes to taraxa-evm on this PR ?

JakubFornadel (approved):

PR #2516, opened 2023-06-30
implement getTotalDelegation function

JakubFornadel:
## Purpose

evm pr https://github.com/Taraxa-project/taraxa-evm/pull/176


rattrap:
Can't have a batch param. We need it to return the total number of staked tokens for that account and we can only call it once.

PR #2520, opened 2023-07-10
feat: disable dpos contract call depth check in magnolia hardfork

kstdl:
Evm PR should be checked and merged first: https://github.com/Taraxa-project/taraxa-evm/pull/177


kstdl:
We don't need this PR as it only updates taraxa-evm submodule which will be updated in target branch

PR #2551, opened 2023-08-16
chore: async clear light node history

mfrankovi:
Even with some latest optimizations getting transactions from period data to be able to delete old data from Columns::final_chain_receipt_by_trx_hash can sometimes take up to 20 seconds. With DagManager::clearLightNodeHistory running synchronously blocking the node from pushing new blocks to the chain this causes node to get out of synced state.

Change is made that clearing light node history is now done on a separate thread. Since this is done once an hour, a new thread is always created since there is no need to have an active thread in thread pool which is only utilized once an hour for a light node.

### DIFF
--- a/libraries/core_libs/storage/src/storage.cpp
+++ b/libraries/core_libs/storage/src/storage.cpp
@@ -374,7 +374,7 @@ std::optional<h256> DbStorage::getGenesisHash() {
 }
 
 DbStorage::~DbStorage() {
-  if (clear_history_thread_) clear_history_thread_->join();
+  if (clear_history_future_) clear_history_future_->wait();
   for (auto cf : handles_) {
     checkStatus(db_->DestroyColumnFamilyHandle(cf));
   }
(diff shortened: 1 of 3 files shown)


libraries/core_libs/storage/src/storage.cpp
```diff
@@ -564,75 +565,85 @@ std::optional<SortitionParamsChange> DbStorage::getParamsChangeForPeriod(PbftPer
...
-      db_->DeleteRange(write_options_, handle(Columns::dag_blocks_level), start_slice, end_slice);
-
+    it = std::unique_ptr<rocksdb::Iterator>(db_->NewIterator(read_options_, handle(Columns::dag_blocks_level)));
+    it->SeekToFirst();
+    if (it->Valid()) {
+      uint64_t start_level;
+      memcpy(&start_level, it->key().data(), sizeof(uint64_t));
+      if (clear_history_thread_) clear_history_thread_->join();
+      clear_history_thread_ =
+          std::make_unique<std::thread>([this, start_period, start_level, initial, dag_level_to_keep, end_period]() {
```
kstdl, line 582: Maybe its better to use [std::async](https://en.cppreference.com/w/cpp/thread/async) here?

kstdl (approved):

PR #2566, opened 2023-08-31
chore: light node size optimization

mfrankovi:
- Delete data from trx_period, pbft_block_period and final_chain_log_blooms_index columns older than light node history
- dag_block_period column was not correctly deleted, fixed

With these changes size of the db snapshot columns with 7 days history:
period_data 3.2GB
final_chain_receipt_by_trx_hash 413MB
trx_period 371MB

All other columns smaller than 100MB

### DIFF
--- a/libraries/core_libs/storage/src/storage.cpp
+++ b/libraries/core_libs/storage/src/storage.cpp
@@ -8,6 +8,7 @@
 
 #include "config/version.hpp"
 #include "dag/sortition_params_manager.hpp"
+#include "final_chain/final_chain.hpp"
 #include "rocksdb/utilities/checkpoint.h"
 #include "storage/uint_comparator.hpp"
 #include "vote/vote.hpp"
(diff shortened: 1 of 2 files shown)


libraries/core_libs/storage/src/storage.cpp
```diff
@@ -1122,6 +1136,22 @@ std::vector<std::shared_ptr<DagBlock>> DbStorage::getFinalizedDagBlockByPeriod(P
   return ret;
 }
 
+std::pair<blk_hash_t, std::vector<std::shared_ptr<DagBlock>>>
+DbStorage::getLastPbftblockHashAndFinalizedDagBlockByPeriod(PbftPeriod period) {
+  std::vector<std::shared_ptr<DagBlock>> ret;
+  blk_hash_t last_pbft_block_hash;
+  if (auto period_data = getPeriodDataRaw(period); period_data.size() > 0) {
+    auto dag_blocks_data = dev::RLP(period_data)[DAG_BLOCKS_POS_IN_PERIOD_DATA];
```
kstdl, line 1144: I think its better to move RLP to a separate variable, so you can reuse it below to get pbft block hash

mfrankovi: Fixed

kstdl (approved):

PR #2602, opened 2023-11-09
chore: improve syncing speed

mfrankovi:
Rocksdb options sync which was set to false only in final_chain db operations makes every write very slow because it waits for the data to be written to disk. This is specially noticeable on slower SSD and HDD. Removing it improves syncing speed. I do not see any reason why we ever used this.

When syncing faster than processing the blocks we sync which can happen often, a condition variable is used to notify the waiting blocks once a new block is finalized. This also reduces waiting time and improves syncing speed.

### DIFF
--- a/libraries/common/src/vrf_wrapper.cpp
+++ b/libraries/common/src/vrf_wrapper.cpp
@@ -29,12 +29,17 @@ std::optional<vrf_proof_t> getVrfProof(vrf_sk_t const &sk, bytes const &msg) {
   return {};
 }
 
-std::optional<vrf_output_t> getVrfOutput(vrf_pk_t const &pk, vrf_proof_t const &proof, bytes const &msg) {
+std::optional<vrf_output_t> getVrfOutput(vrf_pk_t const &pk, vrf_proof_t const &proof, bytes const &msg, bool strict) {
   vrf_output_t output;
-  // crypto_vrf_verify return 0 on success!
-  if (!crypto_vrf_verify(output.data(), const_cast<unsigned char *>(pk.data()),
-                         const_cast<unsigned char *>(proof.data()), const_cast<unsigned char *>(msg.data()),
-                         msg.size())) {
+  if (strict) {
+    // crypto_vrf_verify return 0 on success!
+    if (!crypto_vrf_verify(output.data(), const_cast<unsigned char *>(pk.data()),
+                           const_cast<unsigned char *>(proof.data()), const_cast<unsigned char *>(msg.data()),
+                           msg.size())) {
+      return output;
+    }
+  } else {
+    crypto_vrf_proof_to_hash(output.data(), const_cast<unsigned char *>(proof.data()));
     return output;
   }
   return {};
@@ -47,11 +52,8 @@ dev::bytes VrfSortitionBase::makeVrfInput(taraxa::level_t level, const dev::h256
   return s.invalidate();
 }
 
-bool VrfSortitionBase::verify(const vrf_pk_t &pk, const bytes &msg, uint16_t vote_count) const {
-  if (!isValidVrfPublicKey(pk)) {
-    return false;
-  }
-  auto res = vrf_wrapper::getVrfOutput(pk, proof_, msg);
+bool VrfSortitionBase::verify(const vrf_pk_t &pk, const bytes &msg, uint16_t vote_count, bool strict) const {
+  auto res = vrf_wrapper::getVrfOutput(pk, proof_, msg, strict);
   if (res != std::nullopt) {
     output_ = res.value();
     thresholdFromOutput(vote_count);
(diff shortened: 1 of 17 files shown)


libraries/core_libs/node/src/node.cpp
```diff
@@ -381,24 +397,31 @@ void FullNode::rebuildDb() {
...
 
     if (period - 1 == conf_.db_config.rebuild_db_period) {
       break;
     }
-    while (final_chain_->last_block_number() != period - 1) {
-      thisThreadSleepForMilliSeconds(5);
-      LOG(log_nf_) << "Waiting on PBFT blocks to be processed. PBFT chain size " << pbft_mgr_->pbftSyncingPeriod()
-                   << ", final chain size: " << final_chain_->last_block_number();
+
+    if (period % 10000) {
```
kstdl, line 418:
```suggestion
    if (period % 10000 == 0) {
```

mfrankovi: Fixed

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
```diff
@@ -1702,21 +1702,27 @@ std::optional<std::pair<PeriodData, std::vector<std::shared_ptr<Vote>>>> PbftMan
...
-    return std::nullopt;
-  }
-  for (uint32_t i = 0; i < period_data.transactions.size(); i++) {
-    if (!non_finalized_transactions.contains(period_data.transactions[i]->getHash())) {
-      LOG(log_er_) << "Synced PBFT block " << pbft_block_hash << " has incorrect transaction "
-                   << period_data.transactions[i]->getHash();
+  const uint32_t sanity_check_interval = 100;
+  // Next section should not be possible for a valid cert voted block, perform check only each sanity_check_interval not
+  // to hurt performance of syncing/rebuilding
+  if (period_data.pbft_blk->getPeriod() % sanity_check_interval == 0) {
```
kstdl, line 1708: Is it taking a lot of time? 
I think it's better to have it every block or don't have at all. Because if it is fine on 100 block it doesn't mean that it is ok for previous blocks

libraries/common/src/vrf_wrapper.cpp
```diff
@@ -29,12 +29,17 @@ std::optional<vrf_proof_t> getVrfProof(vrf_sk_t const &sk, bytes const &msg) {
...
-                         msg.size())) {
+  if (strict) {
+    // crypto_vrf_verify return 0 on success!
+    if (!crypto_vrf_verify(output.data(), const_cast<unsigned char *>(pk.data()),
+                           const_cast<unsigned char *>(proof.data()), const_cast<unsigned char *>(msg.data()),
+                           msg.size())) {
+      return output;
+    }
+  } else {
+    crypto_vrf_proof_to_hash(output.data(), const_cast<unsigned char *>(proof.data()));
```
kstdl, line 42: What is this function doing?

mfrankovi: I did the measurement and verify vrf together with getting the output takes about 300microseconds, just getting the output without full validation takes only 20 microseconds. If there is 50 votes per block for 5m blocks this is about 20 hours of processing to verify all the votes for each block. This is by far the biggest CPU bottleneck of current syncing/rebuilding.

libraries/core_libs/consensus/src/pbft/period_data_queue.cpp
```diff
@@ -38,7 +38,7 @@ bool PeriodDataQueue::push(PeriodData &&period_data, const dev::p2p::NodeID &nod
   const auto period = period_data.pbft_blk->getPeriod();
   std::unique_lock lock(queue_access_);
 
-  if (period != std::max(period_, max_pbft_size) + 1) {
+  if (period != std::max(period_, max_pbft_size) + 1 && (queue_.empty() && period != max_pbft_size + 2)) {
```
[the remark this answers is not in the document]
mfrankovi: Maybe there is a better way to fix it. So I often got an error when we are just about to get sync. We are pushing a block to the queue at the moment where previous block was retrieved from the queue but not yet pushed in the chain. It is a valid scenario that the block we are pushing could be either 1 or two blocks ahead if one block is just about to get pushed in the chain.
There is no danger in this change because in PbftManager::processPeriodData we do another check that the proper block is in the queue. 
I added a comment about this.

libraries/core_libs/consensus/src/pbft/pbft_manager.cpp
mfrankovi: It is taking about 1.8millisecond per block. So it is 1.8s per 1000 block or if we look at mainnet with 5m blocks it is about 3 hours of processing when processing full mainnet blocks.

mfrankovi: I removed it completely

kstdl (approved):

PR #2636, opened 2023-12-19
rpc: add block number to gas estimation

MatusKysel:

### DIFF
--- a/CMakeModules/ProjectJSONRPCCPP.cmake
+++ b/CMakeModules/ProjectJSONRPCCPP.cmake
@@ -4,10 +4,8 @@ set(include_path "${prefix}/include")
 
 ExternalProject_Add(jsonrpccpp
     PREFIX "${prefix}"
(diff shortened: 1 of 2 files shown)


VargaElod23 (approved):

libraries/core_libs/network/rpc/eth/Eth.cpp
```diff
@@ -159,9 +160,14 @@ class EthImpl : public Eth, EthParams {
     return toJS(ret.code_retval);
   }
 
-  string eth_estimateGas(const Json::Value& _json) override {
+  string eth_estimateGas(const Json::Value& _json, const string& blockNumber) override {
     auto t = toTransactionSkeleton(_json);
-    auto blk_n = final_chain->last_block_number();
+    EthBlockNumber blk_n;
+    if (!blockNumber.empty()) {
```
kstdl, line 166: If it was done because of the tests I think it is better to specify "latest" there and don't have this if

kstdl (approved):

PR #2662, opened 2024-01-19
chore: fix commision rewards

MatusKysel:

### DIFF
--- a/libraries/config/include/config/hardfork.hpp
+++ b/libraries/config/include/config/hardfork.hpp
@@ -27,7 +27,7 @@ struct HardforksConfig {
   // disable it by default (set to max uint64)
   uint64_t fix_redelegate_block_num = -1;
   // disable it by default (set to max uint64)
-  uint64_t fix_commission_block_num = -1;
+  uint64_t cora_hf_block_num = -1;
   std::vector<Redelegation> redelegations;
   /*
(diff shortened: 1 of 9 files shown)


mfrankovi (approved):

libraries/config/include/config/hardfork.hpp
```diff
@@ -26,6 +26,8 @@ void dec_json(const Json::Value& json, MagnoliaHardfork& obj);
 struct HardforksConfig {
   // disable it by default (set to max uint64)
   uint64_t fix_redelegate_block_num = -1;
+  // disable it by default (set to max uint64)
+  uint64_t phalaenopsis_hf_block_num = -1;
```
JakubFornadel, line 30: I think it should be enabled by default `= 0`

JakubFornadel (approved):

libraries/config/include/config/hardfork.hpp
kstdl: This change will affect only some default networks that won't have this value in predefined genesis. So maybe we can discuss this on Monday meet?

JakubFornadel: yeah we can, but there is no reason for any network to have this off... Most of the hardforks should be enabled by default

PR #2689, opened 2024-02-07
chore: adding version to pbft block

mfrankovi:
Pbft block after aspen hf part two contains an optional extra field similar to extra_data field in ethereum block. Final chain ethereum block representation contains the field. After aspen hf, this field will contain a version info in format:
{'T', TARAXA_MAJOR_VERSION, TARAXA_MINOR_VERSION, TARAXA_PATCH_VERSION, TARAXA_NET_VERSION})

### DIFF
--- a/libraries/core_libs/consensus/src/final_chain/final_chain.cpp
+++ b/libraries/core_libs/consensus/src/final_chain/final_chain.cpp
@@ -197,8 +197,9 @@ class FinalChainImpl final : public FinalChain {
     auto rewards_stats = rewards_.processStats(new_blk, transactions_gas_used, batch);
     const auto& [state_root, total_reward] = state_api_.distribute_rewards(rewards_stats);
 
-    auto blk_header = append_block(batch, new_blk.pbft_blk->getBeneficiary(), new_blk.pbft_blk->getTimestamp(),
-                                   kBlockGasLimit, state_root, total_reward, new_blk.transactions, receipts);
+    auto blk_header =
+        append_block(batch, new_blk.pbft_blk->getBeneficiary(), new_blk.pbft_blk->getTimestamp(), kBlockGasLimit,
+                     state_root, total_reward, new_blk.transactions, receipts, new_blk.pbft_blk->getExtraDataRlp());
     // Update number of executed DAG blocks and transactions
     auto num_executed_dag_blk = num_executed_dag_blk_ + finalized_dag_blk_hashes.size();
     auto num_executed_trx = num_executed_trx_ + new_blk.transactions.size();
@@ -274,7 +275,7 @@ class FinalChainImpl final : public FinalChain {
   std::shared_ptr<BlockHeader> append_block(DB::Batch& batch, const addr_t& author, uint64_t timestamp,
                                             uint64_t gas_limit, const h256& state_root, u256 total_reward,
                                             const SharedTransactions& transactions = {},
-                                            const TransactionReceipts& receipts = {}) {
+                                            const TransactionReceipts& receipts = {}, const bytes& extra_data = {}) {
     auto blk_header_ptr = std::make_shared<BlockHeader>();
     auto& blk_header = *blk_header_ptr;
     auto last_block = block_header();
@@ -286,6 +287,7 @@ class FinalChainImpl final : public FinalChain {
     blk_header.gas_used = receipts.empty() ? 0 : receipts.back().cumulative_gas_used;
     blk_header.gas_limit = gas_limit;
     blk_header.total_reward = total_reward;
+    blk_header.extra_data = extra_data;
     dev::BytesMap trxs_trie, receipts_trie;
     dev::RLPStream rlp_strm;
     for (size_t i(0); i < transactions.size(); ++i) {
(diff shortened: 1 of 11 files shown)


libraries/types/pbft_block/src/pbft_block.cpp
```diff
@@ -11,22 +11,37 @@ namespace taraxa {
 PbftBlock::PbftBlock(bytes const& b) : PbftBlock(dev::RLP(b)) {}
 
 PbftBlock::PbftBlock(dev::RLP const& rlp) {
-  util::rlp_tuple(util::RLPDecoderRef(rlp, true), prev_block_hash_, dag_block_hash_as_pivot_, order_hash_,
-                  prev_state_root_hash_, period_, timestamp_, reward_votes_, signature_);
+  if (rlp.itemCount() == 9) {
```
JakubFornadel, line 14: this will be problematic because I am adding another optional parameter into pbft block so possible itemCount will be 8,9 or 10. If size == 9, we won't be able to say which optional param is present.

if you declare extra_data as std::optional<bytes> and leave it empty, it will put 1 zero byre into the rlp, otherwise the whole extra data and we don't need to check itemCount. You can then use util::rlp_tuple for all params...

mfrankovi: This will make hard fork much more complex since we will need to have different format of pbft block before and after the hardfork and there will be a lot more places in the code where we have to check if we are dealing with an old format block or a new one. Because adding a zero byte in rlp changes the hash, so all the pre-fork blocks need to have the old format. I mean it should be the same case for your param as well. We will still need to check the item count because it could always be an old format block. I will think some more if there is some better way to handle this.

JakubFornadel: ok give it a though. I don't see any other way how to do it at the moment...

mfrankovi: What do you think about putting the pillar block hash within the extra_data field so this way we only have one optional field? Ethereum standard for extra data is 32 bytes but we could expand it to more to fit both version and pillar block hash.

JakubFornadel: it would work probably. Not sure if it is good or bad solution to put it there...

mfrankovi: I made some changes that the extra_data part as a separate optional structure within the pbft block. So you could add your change within there. So with your change pbft block rlp count would no longer change, only the rlp within the PbftBlockExtraData would change

libraries/types/pbft_block/include/pbft/pbft_block.hpp
```diff
@@ -15,6 +15,57 @@ namespace taraxa {
  * @{
  */
 
+class PbftBlockExtraData {
```
kstdl, line 18: I think it's better to move it to separate files and have this version related things in a separate structure, so it will be easier to extend it later

JakubFornadel: yes and pls also separate implementations into the cpp file

libraries/types/pbft_block/include/pbft/pbft_block.hpp
```diff
@@ -15,6 +15,57 @@ namespace taraxa {
...
+  PbftBlockExtraData(const bytes& data) {
+    if (data.size() > kExtraDataMaxSize) {
+      throw std::runtime_error("Pbft block invalid, extra data size over the limit");
+    }
+    dev::RLP rlp(data);
+    util::rlp_tuple(util::RLPDecoderRef(rlp, true), major_version_, minor_version_, patch_version_, net_version_,
+                    node_implementation_);
+  }
+
+  bytes rlp() const {
```
JakubFornadel, line 38: the implementations into the cpp file

libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -197,8 +197,10 @@ class FinalChainImpl final : public FinalChain {
     auto rewards_stats = rewards_.processStats(new_blk, transactions_gas_used, batch);
     const auto& [state_root, total_reward] = state_api_.distribute_rewards(rewards_stats);
 
-    auto blk_header = append_block(batch, new_blk.pbft_blk->getBeneficiary(), new_blk.pbft_blk->getTimestamp(),
-                                   kBlockGasLimit, state_root, total_reward, new_blk.transactions, receipts);
+    auto blk_header =
+        append_block(batch, new_blk.pbft_blk->getBeneficiary(), new_blk.pbft_blk->getTimestamp(), kBlockGasLimit,
+                     state_root, total_reward, new_blk.transactions, receipts,
+                     new_blk.pbft_blk->getExtraData().has_value() ? new_blk.pbft_blk->getExtraData()->rlp() : bytes());
```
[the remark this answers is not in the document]
mfrankovi: Done

libraries/types/pbft_block/include/pbft/pbft_block.hpp
mfrankovi: Done

libraries/types/pbft_block/include/pbft/pbft_block.hpp
mfrankovi: Done

JakubFornadel (approved):

PR #2727, opened 2024-03-20
bugfix: fix rewards stats cleaning on crash

MatusKysel:
## Purpose
This PR is fixing issue, when crash happened after `processStats` but before `transition_state_commit`. In that scenario, we would clean the DB and on restart we would add only rewards from one block. With this approach we do cleanup as last after both commits (db & state) and in case of crash before `clear` function we have recovery function, that will deal with what was left in DB

### DIFF
--- a/libraries/core_libs/consensus/src/rewards/rewards_stats.cpp
+++ b/libraries/core_libs/consensus/src/rewards/rewards_stats.cpp
@@ -6,8 +6,7 @@
 
 namespace taraxa::rewards {
 Stats::Stats(uint32_t committee_size, const HardforksConfig& hardforks, std::shared_ptr<DB> db,
-             std::function<uint64_t(EthBlockNumber)>&& dpos_eligible_total_vote_count,
-             std::optional<EthBlockNumber> last_blk_num)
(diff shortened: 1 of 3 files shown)


libraries/core_libs/consensus/src/rewards/rewards_stats.cpp
```diff
@@ -1,22 +1,30 @@
...
 }
 
-void Stats::loadFromDb() {
+void Stats::recoverFromDb(std::optional<EthBlockNumber> last_blk_num) {
   auto i = db_->getColumnIterator(DB::Columns::block_rewards_stats);
   for (i->SeekToFirst(); i->Valid(); i->Next()) {
     PbftPeriod period;
     memcpy(&period, i->key().data(), sizeof(PbftPeriod));
+    if (last_blk_num && *last_blk_num >= period) {
+      std::cout << "Skipping block rewards stats recovery for period " << period << " as it is already processed"
```
kstdl, line 24: It's this log needed?

kstdl (approved):

JakubFornadel (approved):

PR #2739, opened 2024-04-11
chore: do not reward dag blocks with incorrect level

mfrankovi:
After bamboo HF, dag block with the same level or with level smaller than the last awarded level will not rewarded to prevent any reward exploit using DAG blocks level manipulation


libraries/core_libs/consensus/include/rewards/block_stats.hpp
```diff
@@ -22,10 +22,13 @@ class BlockStats {
...
    * @param dpos_vote_count - votes count for previous block
    * @param committee_size
+   * @param dag_blocks_author_level
    * @param aspen_dag_rewards - aspen dag rewards
+   * @param bamboo_dag_rewards - bamboo dag rewards
    */
   BlockStats(const PeriodData& block, const std::vector<gas_t>& trxs_gas_used, uint64_t dpos_vote_count,
-             uint32_t committee_size, const bool aspen_dag_rewards = false);
+             uint32_t committee_size, std::unordered_map<addr_t, uint64_t>& dag_blocks_author_level,
+             const bool aspen_dag_rewards = false, const bool bamboo_dag_rewards = false);
```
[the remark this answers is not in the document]
mfrankovi: Done

libraries/core_libs/consensus/include/rewards/block_stats.hpp
```diff
@@ -35,8 +38,11 @@ class BlockStats {
    *
    * @param block
    * @param aspen_dag_rewards
+   * @param bamboo_dag_rewards
+   * @param dag_blocks_author_level
    */
-  void processStats(const PeriodData& block, const bool aspen_dag_rewards);
+  void processStats(const PeriodData& block, const bool aspen_dag_rewards, const bool bamboo_dag_rewards,
+                    std::unordered_map<addr_t, uint64_t>& dag_blocks_author_level);
```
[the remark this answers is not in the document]
mfrankovi: Done

libraries/core_libs/consensus/include/rewards/block_stats.hpp
```diff
@@ -49,8 +55,11 @@ class BlockStats {
    * @brief Process Dag blocks and save stats in class for future serialization with aspen HF changes. returns
    *
    * @param block
+   * @param dag_blocks_author_level
+   * @param bamboo_dag_rewards
    */
-  void processDagBlocksAspen(const PeriodData& block);
+  void processDagBlocksAspen(const PeriodData& block, std::unordered_map<addr_t, uint64_t>& dag_blocks_author_level,
+                             const bool bamboo_dag_rewards);
```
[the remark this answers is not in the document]
mfrankovi: Done

libraries/core_libs/consensus/src/rewards/block_stats.cpp
```diff
@@ -117,8 +121,13 @@ void BlockStats::processDagBlocksAspen(const PeriodData& block) {
...
     if (dag_block.getDifficulty() == min_difficulty) {
-      validators_stats_[dag_block_author].dag_blocks_count_ += 1;
-      total_dag_blocks_count_ += 1;
+      const auto& dag_block_level = dag_block.getLevel();
+      // This check prevents rewarding dag block author which produces multiple dag blocks on the same level
+      if (!bamboo_dag_rewards || dag_block_level > dag_blocks_author_level[dag_block.getSender()]) {
+        dag_blocks_author_level[dag_block.getSender()] = dag_block_level;
+        validators_stats_[dag_block_author].dag_blocks_count_ += 1;
+        total_dag_blocks_count_ += 1;
+      }
```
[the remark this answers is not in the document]
mfrankovi: I believe they are. Specially for the blocks from the same block producer unless producer deliberately produces parallel blocks.

libraries/core_libs/consensus/src/rewards/rewards_stats.cpp
```diff
@@ -112,4 +114,20 @@ std::vector<BlockStats> Stats::processStats(const PeriodData& current_blk, const
...
+  for (uint64_t period = start_period; period <= last_block_num; period++) {
+    auto data = db_->getPeriodDataRaw(period);
+    if (data.size() == 0) throw std::runtime_error("populateDagBlockAuthorLevels missing period data in db");
+    PeriodData period_data(data);
+    for (auto dag_block : period_data.dag_blocks) {
+      dag_blocks_author_level_[dag_block.getSender()] = dag_block.getLevel();
+    }
+  }
+}
+
```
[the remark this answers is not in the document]
mfrankovi: On my machine this runs under 1 second so I think it should be fine. This is only done on restarts.

libraries/core_libs/consensus/src/rewards/block_stats.cpp
kstdl, line 130: I think it will be cleaner like this
```
// do't reward block with the same a lower level 
if (bamboo_dag_rewards && dag_block_level <= dag_blocks_author_level[dag_block_author]) {
   continue;
}
```

libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -116,6 +116,7 @@ class FinalChainImpl final : public FinalChain {
       for (uint64_t num = start; num <= *last_blk_num; ++num) {
         block_headers_cache_.get(num);
       }
+      rewards_.populateDagBlockAuthorLevels(*last_blk_num);
```
kstdl, line 119: Why is this called from final_chain?

mfrankovi: In final chain finalize method is where this is actually used and it requires last block number as a parameter so final chain constructor where we set the last block number seems to be the best place.

libraries/core_libs/consensus/src/rewards/block_stats.cpp
mfrankovi: I will change it to this

mfrankovi: Actually I cannot do it this way because there is a section under if which need to run. This part:
    }

kstdl: I think you can move it above, right before the if

mfrankovi: I moved the trx iteration above the if and rearranged it this way. Please check again

kstdl (approved):

PR #2761, opened 2024-05-28
Update deps

MatusKysel:
## Purpose
Everything that was easily updated, is updated. Code was just formatted with new clang-format. Some issue:
- conan setup is little bit more trickier with clang-18
- py_test are disabled as they had super old dependencies
- we can not update rocksdb to the latest (crash) 
- cppcheck require some work too


JakubFornadel (approved):

JakubFornadel:
Let's merge this asap as builds are not working anymore...

PR #2768, opened 2024-06-03
System transactions

kstdl:


libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -316,15 +339,30 @@ class FinalChainImpl final : public FinalChain {
...
+        auto i_rlp = util::rlp_enc(rlp_strm, receipt_idx);
+        trxs_trie[i_rlp] = sys_trx->rlp();
+
+        const auto& receipt = receipts[receipt_idx];
+        receipts_trie[i_rlp] = util::rlp_enc(rlp_strm, receipt);
+        db_->insert(batch, DB::Columns::final_chain_receipt_by_trx_hash, sys_trx->getHash(), rlp_strm.out());
+
+        blk_header.log_bloom |= receipt.bloom();
+        receipt_idx++;
+      }
```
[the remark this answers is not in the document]
kstdl: 1. This if here is to check that we have enough receipts, so it won't just crash if doesn't have a receipt at the index
2. Then I will need to implement getHash for the EVMTransaction, but I don't think it is easy as we need a signature for this, that we doesn't really have in the EVMTransaction

libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -240,6 +246,21 @@ class FinalChainImpl final : public FinalChain {
     }
     ////
 
+    //// Commit system transactions
+    if (!system_transactions.empty()) {
+      db_->addPeriodSystemTransactions(batch, system_transactions, new_blk.pbft_blk->getPeriod());
+      auto position = new_blk.transactions.size() + 1;
+      std::cout << "system trx size: " << system_transactions.size() << std::endl;
```
JakubFornadel, line 253: std::cout ?

libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -240,6 +246,21 @@ class FinalChainImpl final : public FinalChain {
...
+      for (const auto& trx : system_transactions) {
+        std::cout << "saving " << trx->getHash() << std::endl;
```
JakubFornadel, line 255: same here

libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -316,15 +338,30 @@ class FinalChainImpl final : public FinalChain {
     blk_header.extra_data = extra_data;
     dev::BytesMap trxs_trie, receipts_trie;
     dev::RLPStream rlp_strm;
-    for (size_t i(0); i < transactions.size(); ++i) {
-      auto const& trx = transactions[i];
-      auto i_rlp = util::rlp_enc(rlp_strm, i);
+    auto receipt_idx = 0;
```
JakubFornadel, line 341: optional: transation_id probably makes more sense as you are itereting over trasactions and system_transaction - receipts then depend on transaction and its id...

libraries/core_libs/consensus/src/final_chain/final_chain.cpp
```diff
@@ -316,15 +338,30 @@ class FinalChainImpl final : public FinalChain {
...
+
+      const auto& receipt = receipts[receipt_idx];
       receipts_trie[i_rlp] = util::rlp_enc(rlp_strm, receipt);
       db_->insert(batch, DB::Columns::final_chain_receipt_by_trx_hash, trx->getHash(), rlp_strm.out());
-      auto bloom = receipt.bloom();
-      blk_header.log_bloom |= bloom;
+
+      blk_header.log_bloom |= receipt.bloom();
+    }
+    if (system_transactions.size() && receipts.size() == (transactions.size() + system_transactions.size())) {
```
JakubFornadel, line 353: when can it happen that receipts.size() != (transactions.size() + system_transactions.size()) ?

JakubFornadel (approved):

PR #2830, opened 2024-08-12
chore: add rpc for retrieving nodes versions

mfrankovi:
RPC getNodeVersions will retrieve version for any node that produced a pbft block within last 6000 blocks(6 hours) and provide a summary of how many nodes are running each individual version.

### DIFF
--- a/libraries/core_libs/network/rpc/Taraxa.cpp
+++ b/libraries/core_libs/network/rpc/Taraxa.cpp
@@ -110,6 +110,58 @@ Json::Value Taraxa::taraxa_getScheduleBlockByPeriod(const std::string& _period) 
   }
 }
 
+Json::Value Taraxa::taraxa_getNodeVersions() {
+  try {
+    Json::Value res;
+    auto node = tryGetNode();
+    auto db = node->getDB();
+    auto period = node->getFinalChain()->last_block_number();
+    const uint64_t max_blocks_to_process = 6000;
+    std::map<addr_t, std::string> node_version_map;
+    std::multimap<std::string, std::pair<addr_t, uint64_t>> version_node_map;
+    std::map<std::string, std::pair<uint32_t, uint32_t>> version_count;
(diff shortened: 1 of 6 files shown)


libraries/core_libs/network/rpc/Taraxa.cpp
```diff
@@ -110,6 +110,52 @@ Json::Value Taraxa::taraxa_getScheduleBlockByPeriod(const std::string& _period)
   }
 }
 
+Json::Value Taraxa::taraxa_getNodeVersions() {
+  try {
+    Json::Value res;
+    auto node = tryGetNode();
+    auto db = node->getDB();
+    auto period = node->getFinalChain()->last_block_number();
+    const uint64_t max_blocks_to_process = 6000;
```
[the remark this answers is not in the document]
mfrankovi: On my machine it is run quickly. The reason for 6000 or 6 hours is that some nodes will produce a pbft block rarely. For instance node with minimal delegation on mainnet produces only one pbft block each 4 hours so to catch most of the online nodes we do need at least a couple of hours of pbft blocks production.

libraries/core_libs/network/rpc/Taraxa.cpp
```diff
@@ -110,6 +110,52 @@ Json::Value Taraxa::taraxa_getScheduleBlockByPeriod(const std::string& _period)
...
+    }
+
+    res["nodes"] = Json::Value(Json::arrayValue);
+    for (auto vn : version_node_map) {
+      Json::Value node_json;
+      node_json["node"] = vn.second.toString();
+      node_json["version"] = vn.first;
+      res["nodes"].append(node_json);
+    }
+    res["versions"] = Json::Value(Json::arrayValue);
```
kstdl, line 147: Aren't we more interested in the votes count and threshold? Maybe it should be added as separate mapping, so which version have how much votes

mfrankovi: I added vote count for each node and version as well

kstdl (approved):

PR #2845, opened 2024-09-03
feat: reject dag block with invalid sender

MatusKysel:


mfrankovi (commented):
This change means that once account was created and received some balance it would be locked for about 100 periods before it could use the balance? Do you think that this is ok?

PR #2852, opened 2024-09-18
fix: dag block bundle

mfrankovi:
Using uint16 as transaction index was unsafe for high transactions load where one pbft block could contain over 65k transactions.


kstdl (approved):

mfrankovi:
> I think we need to remeasure if with this change it will save any space.

I measured migration and the savings I got with uint32_t were exactly the same as with uint16_t. It should be the same since RLP saves both in it own dynamic number format

PR #2943, opened 2025-01-30
Revert rate limit

kstdl:


mfrankovi (commented):
I am fine if we for instance change the limit from 100 to 1000. I just would not remove it completely because this would kill the node when we are too slow to process requests or there are too many of them

kstdl:
@mfrankovi this isn't making anything better. As we saw yesterday RPC nodes were running out of memory with this limit. But it is messing with apps where we were seeing alot of random disconnects because there was like 1 long request some quick ones. Rate limiting could be implemented separately on the load balancer layer, but I don't think it should be part of the node. 
If you want to keep this you can just create a separate PR with config changes and get approves from @MatusKysel @JakubFornadel and @rattrap

mfrankovi (approved):

PR #2992, opened 2025-03-18
chore: large gas limit transactions handling

mfrankovi:
Any transaction over the propose dag gas limit should not be proposable in pool.

Include smaller gas limit transactions in packtrxs if large one does not fit.

### DIFF
--- a/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
@@ -154,6 +154,13 @@ TransactionStatus TransactionManager::insertValidatedTransaction(std::shared_ptr
     proposable = false;
   }
 
+  if (kConf.propose_dag_gas_limit < tx->getGas()) {
+    if (!insert_non_proposable) {
+      return TransactionStatus::Known;
+    }
+    proposable = false;
+  }
+
   const auto last_block_number = final_chain_->lastBlockNumber();
   LOG(log_dg_) << "Transaction " << trx_hash << " inserted in trx pool";
   return transactions_pool_.insert(std::move(tx), proposable, last_block_number);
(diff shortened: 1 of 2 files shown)


libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -354,6 +362,10 @@ std::pair<SharedTransactions, std::vector<uint64_t>> TransactionManager::packTrx
     total_weight += weight;
     if (total_weight > weight_limit) {
       trxs.resize(i);
+      total_weight -= weight;
+      if(weight_limit - total_weight >= 21000) {
```
JakubFornadel, line 366: kMinTxGas should be used instead of 21000

JakubFornadel (approved):

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -154,6 +155,13 @@ TransactionStatus TransactionManager::insertValidatedTransaction(std::shared_ptr
     proposable = false;
   }
 
+  if (kProposeDagBlockGasLimit < tx->getGas()) {
```
JakubFornadel, line 158: you can use kConf.propose_dag_gas_limit here, no need to define this constant kProposeDagBlockGasLimit ?

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
mfrankovi: Done

libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
mfrankovi: Done

PR #2995, opened 2025-03-20
send proposed blocks after last sync packet

JakubFornadel:
## Purpose

Fix transition from syncing to gossiping:

- added new packet type PbftBlocksBundlePacket
- after last sync packet is sent, PbftBlocksBundlePacket packets with bundles of 10 proposed blocks are sent
- PbftBlocksBundlePacket processing is blocked until sync_queue is empty

This PR should be merged first https://github.com/Taraxa-project/taraxa-node/pull/2921


JakubFornadel:
> I don't think it is ok, if you remember we need to have many if to pass e.g. status packet to both capabilities [871609d](https://github.com/Taraxa-project/taraxa-node/commit/871609da61cc8e6f7fe57eccc269e5c23c500500)

nope not in this case. It is ok

mfrankovi (approved):

PR #2996, opened 2025-03-21
feat: concurrent trx exstimation during dag verification

kstdl:

### DIFF
--- a/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
+++ b/libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
@@ -4,6 +4,7 @@
 #include <unordered_set>
 #include <utility>
 
+#include "common/thread_pool.hpp"
 #include "config/config.hpp"
 #include "logger/logger.hpp"
 #include "transaction/transaction.hpp"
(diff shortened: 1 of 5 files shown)


libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -23,6 +24,22 @@ TransactionManager::TransactionManager(const FullNodeConfig &conf, std::shared_p
   }
 }
 
+uint64_t TransactionManager::estimateTransactions(const SharedTransactions &trxs,
+                                                  std::optional<PbftPeriod> proposal_period) const {
+  util::ThreadPool estimation_thread_pool_(std::thread::hardware_concurrency() / 2);
```
JakubFornadel, line 29: this is declaring new threadpool everytime it enters estimateTransactions function ?

kstdl: yeah, because there is a possibility of multiple dag blocks verifications(and this estimations) at the same time

JakubFornadel: hmm and why is it also DagManager class member ?

kstdl: I will rework this using this https://github.com/Taraxa-project/taraxa-node/pull/2998

mfrankovi (approved):
When checking how estimation works internally I see that

PR #3004, opened 2025-03-28
fix of dag validation

MatusKysel:

### DIFF
--- a/tests/full_node_test.cpp
+++ b/tests/full_node_test.cpp
@@ -1625,12 +1625,15 @@ TEST_F(FullNodeTest, SoleiroliaHardfork) {
     EXPECT_EQ(node0->getTransactionManager()->estimateTransactionGas(trx2, node0->getFinalChain()->lastBlockNumber()),
               0);
   }
+  CleanupDirs();
   // After HF
   {
     auto node_cfgs = make_node_cfgs(1, 1, 5);
     for (auto &cfg : node_cfgs) {
       cfg.genesis.state.hardforks.soleirolia_hf.block_num = 0;
       cfg.genesis.dag.gas_limit = 31500000;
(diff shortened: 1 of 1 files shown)


libraries/core_libs/consensus/src/transaction/transaction_manager.cpp
```diff
@@ -92,8 +94,12 @@ std::pair<bool, std::string> TransactionManager::verifyTransaction(const std::sh
             "chain_id mismatch " + std::to_string(trx->getChainID()) + " " + std::to_string(kConf.genesis.chain_id)};
   }
 
+  const auto block_num = final_chain_->lastBlockNumber();
+  const auto isOnSoleiroliaHF = kConf.genesis.state.hardforks.isOnSoleiroliaHardfork(
+    std::max(block_num - (from_dag ? kDagExpiryLevelLimit : 0), uint64_t(0)));
```
kstdl, line 99: this is confusing and could probably overflow?

libraries/core_libs/network/src/tarcap/packets_handlers/latest/dag_sync_packet_handler.cpp
```diff
@@ -55,7 +55,7 @@ void DagSyncPacketHandler::process(DagSyncPacket&& packet, const std::shared_ptr
       continue;
     }
 
-    auto [verified, reason] = trx_mgr_->verifyTransaction(trx);
+    auto [verified, reason] = trx_mgr_->verifyTransaction(trx, true);
```
kstdl, line 58: do we have proposal_period there, so we could pass it, but not just true\false?

JakubFornadel (approved):

kstdl (approved):

PR #3021, opened 2025-04-16
support multiple wallets

JakubFornadel:

### DIFF
--- a/libraries/core_libs/consensus/src/dag/dag_block_proposer.cpp
+++ b/libraries/core_libs/consensus/src/dag/dag_block_proposer.cpp
@@ -15,17 +15,13 @@ DagBlockProposer::DagBlockProposer(const FullNodeConfig& config, std::shared_ptr
                                    std::shared_ptr<TransactionManager> trx_mgr,
                                    std::shared_ptr<final_chain::FinalChain> final_chain, std::shared_ptr<DbStorage> db,
                                    std::shared_ptr<KeyManager> key_manager)
-    : bp_config_(config.genesis.dag.block_proposer),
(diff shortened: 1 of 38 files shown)


libraries/core_libs/consensus/src/dag/dag_block_proposer.cpp
```diff
@@ -72,125 +64,137 @@ bool DagBlockProposer::proposeDagBlock() {
...
-  if (*proposal_period < kHardforks.magnolia_hf.block_num) {
-    max_vote_count = final_chain_->dposEligibleTotalVoteCount(*proposal_period);
-  } else {
-    max_vote_count = kValidatorMaxVote;
-  }
+  // Try to propose dag block for each wallet
+  // TODO: optimize
+  for (auto& node_dag_proposer_data : nodes_dag_proposers_data_) {
+    if (!isValidDposProposer(*proposal_period, node_dag_proposer_data.wallet.node_addr)) {
+      return false;
```
mfrankovi, line 71: I think this should be "continue" and not "return" so that if one of the validator is not valid proposer other still propose blocks. There is a couple of more places where "return" should be replaced with continue

JakubFornadel: pls check this fix https://github.com/Taraxa-project/taraxa-node/pull/3021/commits/3dcb0e8ceca79209bbe79d6f19dad2ae06824655

mfrankovi (approved):

kstdl (approved):

JakubFornadel:
force merged as build & tests passed. CircleCI failed on pushing image to repository

PR #3108, opened 2026-05-05
Fix proposal to unblock mainnet 

bender-cryptobauer:
Note, I have done some changes so I will re-run a build and run a node with it to make sure the fix is still good. I wanted to get it out though to see if anyone is still around to have a look.

## Purpose

The mainnet is currently stuck. I'm not an expert in this code base so I relied heavily on AI. But the [investigation](https://gist.github.com/bender-cryptobauer/d70b19a3767d5bcebf33be1d344da881) I ran as part of this sounds reasonable to me.

It seems like we ended up with a collision that prevents block finalization. This PR introduces a graceful fallback if a collision has happened. Note this is a heuristic and not the proper long-term fix.

## Changes made

Graceful fallback to the next period which in case of a collision should be the actual period we need to validate against.


bender-cryptobauer:
@JakubFornadel @rattrap

bender-cryptobauer:
ended up merging this into my own fork and producing a container image that I'm pushing to github

